PR #145, opened 2022-10-05
Units for pressure in entropy calculation

ViktoriiaBaib:
ASE.IdealGasThermo.get_entropy() requires pressure in Pa; 1 bar = 1e5 Pa


arosen93:
Thank you very much for your contribution and attention to detail, @ViktoriiaBaib! I will merge this PR and then fix the unit test that is associated with the thermo job. This is a good lesson that one's code is only as good as their unit tests, and in this case, the unit test was wrong.

Issue #215, opened 2023-04-07
Thermochemistry calcs fail with imaginary modes


arosen93:
As noted by @ViktoriiaBaib, thermochemistry jobs using thermochemistry from ASE currently fail if there are imaginary modes because ASE itself throws a ValueError [here](https://gitlab.com/ase/ase/-/blob/master/ase/thermochemistry.py#L453).

arosen93:
Addressing in #216

PR #216, opened 2023-04-08
Fix xTB thermo jobs from crashing if imag freqs appear

arosen93:


arosen93:
# Overview

This PR addresses #215, ensuring that QuAcc does not crash due to an ASE `ValueError` in `IdealGasThermo` if an imaginary vibrational mode is present. We'd like to still store the results in the database regardless and leave it to the user to decide what they want to do with it.

# Open Question

I have a question for @ViktoriiaBaib and @samblau. I'd like to add a [unit test](https://github.com/arosen93/quacc/blob/47299ab4325812df7176d05a2d9700e8f418db30/tests/recipes/xtb/test_xtb_recipes.py#L112) for a simple test case where an imaginary mode should be expected. Consider these toy examples and the question that results.

## A test case for a system with no vibrational modes

xTB optimization of H2O followed by a frequency calculation. The xTB frequencies as follows:

```
---------------------
  #    meV     cm^-1
---------------------
  0   65.4i    527.2i
  1   11.4i     92.3i
  2    5.7i     45.7i
  3    0.0i      0.0i
  4    0.1       0.6
  5   66.6     537.0
  6  190.7    1538.4
  7  451.7    3643.1
  8  452.7    3651.5
---------------------
```
Since water is non-linear, there are 3N-6 vibrational modes (because we project out the 6 translational and rotational modes). That leaves the following as the "true" vibrational frequencies: 1538.4, 3643.1, 3651.5 cm^-1.

## A test case for a system with at least one imaginary vibrational mode

Take a 90 degree H2O and run a frequency calculation without a prior optimization. The xTB frequencies are as follows:

Again, as before, the "true" 3N-6 vibrational frequencies are: 
```
---------------------
  #    meV     cm^-1
---------------------
  0  130.9i   1056.2i
  1   59.0i    475.6i
  2    0.0i      0.1i
  3    0.0i      0.0i
  4   16.7     134.9
  5   73.3     591.4
  6   88.8     716.2
  7  127.6    1029.0
  8  428.9    3459.5
---------------------
```

As you can see in the above example, in both cases the 3N-6 vibrational frequencies are all real. Is this expected? Should there be a better test system here? Any recommendations of how to catch the appropriate number of relevant imaginary modes, seeing as translations and rotations to be discarded could either be small real numbers of imaginary?

arosen93:
This is for my own notes:

## Setup
I compared using `xtb-python` with ASE's `Vibrations()` module and the `xtb` code using the executable directly without the Python interface.

In `xtb-python`, I ran
```python
atoms = Atoms(
    "OHH",
    positions=[
        (0.000000000, 0.000000000, 0.119262000),
        (0.000000000, 0.596309000, -0.643977000),
        (0.000000000, -0.763239000, -0.477047000),
    ],
)
atoms.calc = XTB(method='GFN2-xTB')
vib = Vibrations(atoms)
vib.run()
vib.summary()
```

In `xtb` without Python, I ran `xtb h2o.xyz --hess` where `h2o.xyz` is the `Atoms` object above.

## Output

With ASE, I get the following modes: 215.4i, 152.4i, 107.9i, 0.0i, 0.1, 299.6, 1681.2, 3466.5, 3599.3. However, due to the 3N-6 rule for vibrational frequencies, we only keep the 3 highest: 1681.2, 3466.5, 3599.3. The rest are formally translations or rotations and should be squashed to 0.

With `xtb` directly using its internal frequency calculation, I get the following: 0, 0, 0, 0, 0, 0, 1681.84, 3513.83, 3599.37.

As a result, we can see that both `xtb` and ASE are producing nearly identical results despite very different ways in which the frequencies are calculated. This is reassuring!

It also implies that the provided H2O molecule, despite being distorted artificially with a 90 degree bond angle, still has no imaginary modes. This is due to the translations and rotations that are projected out. 

## Summary

I did everything fine. It's just that the small number of atoms results in only 3 real modes and they're all positive regardless of the angle in water.

PR #393, opened 2023-06-09
Add support for a Q-Chem calculator and recipes (Part 1)

### LINKED ISSUE #204: Add `qchem` with optional Custodian support
This is getting kickstarted by @samblau.

This will be closed by #393.

samblau:
Overall this PR aims to accomplish the following:

1. A working Q-Chem ASE calculator built atop Pymatgen IO and Custodian's QCJob and error handlers
2. Enable Q-Chem + Sella optimization both for PES minima, transition states, and IRCs

ASR: Closes #204.


quacc/calculators/qchem.py
```diff
@@ -0,0 +1,100 @@
...
+import ase.units
+from ase.calculators.calculator import FileIOCalculator
+from pymatgen.io.ase import AseAtomsAdaptor
+from pymatgen.io.qchem.sets import ForceSet
+
+from quacc import SETTINGS
+from quacc.custodian import qchem as custodian_qchem
+
+
+class QChem(FileIOCalculator):
```
arosen93, line 21: The convention in Quacc is for all classes to be defined using a `@dataclass` wrapper so you don't have self assign all the variables. I will make that change for you later.

quacc/custodian/qchem.py
```diff
@@ -0,0 +1,76 @@
...
+    handlers = []
+    handlers_dict = {
+        "QChemErrorHandler": QChemErrorHandler(),
+    }
+
+    handlers = []
+    for handler_flag in qchem_custodian_handlers:
+        if handler_flag not in handlers_dict:
+            raise ValueError(f"Unknown Q-Chem error handler: {handler_flag}")
+        handlers.append(handlers_dict[handler_flag]
```
arosen93, line 56: missing a `)`

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def write_input(self, atoms, properties=None, system_changes=None):
+        FileIOCalculator.write_input(self, atoms, properties, system_changes)
+        mol = AseAtomsAdaptor.get_molecule(atoms)
+        mol.set_charge_and_spin(charge=self.charge, spin=self.spin_multiplicity)
+        qcin = ForceSet(mol, **self.qchem_input_params)
+        qcin.write("mol.qin")
+
+    def read_results(self):
+        data = QCOutput("mol.qout").data
```
arosen93, line 92: Not imported yet.

.github/workflows/tests.yaml
```diff
@@ -195,10 +195,54 @@ jobs:
...
+
+      - name: Set up conda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+          activate-environment: quacc-env
+
+      - name: Install pip packages
+        run: |
+          pip install .[strict,dev,qchem]
```
arosen93, line 228: I added a unique test suite for Q-Chem like I did for dftbplus, psi4, and tblite since they have their own unique dependencies.

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,132 @@
...
+    Atoms
+        The ASE Atoms object with attached Q-Chem calculator.
+    """
+
+    def __init__(
+        self,
+        input_atoms: Atoms,
+        charge: None | int = None,
+        spin_multiplicity: None | int = None,
+        qchem_input_params: dict = {},
```
arosen93, line 51: Initializing a kwarg with an empty dict isn't recommended. You should set `dict = None` and then in the code within the `__init__` do `self.qchem_input_params = qchem_input_params or {}`.

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,132 @@
...
+        self.spin_multiplicity = spin_multiplicity
+        self.qchem_input_params = qchem_input_params
+        self.use_custodian = use_custodian
+        self.kwargs = kwargs
+
+        # Instantiate the calculator
+        FileIOCalculator.__init__(
+            self,
+            restart=None,
+            ignore_bad_restart_file=FileIOCalculator._deprecated,
```
arosen93, line 67: I'm not familiar with `FileIOCalculator`. If this property is deprecated, do we need it?

arosen93:
Thanks, @samblau! Looks great so far. I left a few minor comments and manually fixed a few things. One additional comment that you should address:

We will want to wrap any functions requiring the `openbabel` dependency (which I added to `setup.py` under the `qchem` `extras_requires` category) with a `monty.dev.requires` decorator in the codebase. 

For instance, with `tblite`, it looks something like:

```python
from monty.dev import requires
try:
    from tblite.ase import TBLite
except ImportError:
    TBLite = None


@requires(
    TBLite,
    "tblite must be installed. Try pip install tblite[ase]",
)
def my_function_using_tblite
```

Additionally, in the test suite, we will want to mark all q-chem tests to be skipped if the necessary dependencies aren't installed, so they don't error out. See below for an example.

https://github.com/Quantum-Accelerators/quacc/blob/55b0383211c155ff83fdcca839daa0f1e09f080d/tests/recipes/tblite/test_tblite_recipes.py#L34-L37

I added a unique Q-chem unit test runner that installs `pip install quacc[qchem]` and will install `openbabel`.

.github/workflows/tests.yaml
```diff
@@ -195,10 +195,58 @@ jobs:
...
+          pip install chardet
+          pip install pytest-cov
+
+      - name: Install conda packages
+        run: |
+          conda install -c conda-forge openbabel
+
+      - name: Run tests with pytest
+        run: pytest tests/calculators/qchem tests/custodian_tests/test_qchem_custodian.py --cov=quacc --cov-report=xml
```
arosen93, line 237: Here I specify which tests to run so the qchem github actions runner isn't wasting time doing the whle unit test suite.

quacc/settings.py
```diff
@@ -128,6 +128,38 @@ class QuaccSettings(BaseSettings):
         description="After this many seconds, Custodian will stop running and ensure that VASP writes a STOPCAR",
     )
 
+    # ---------------------------
+    # Q-Chem Settings
+    # ---------------------------
+
+    # Q-Chem Settings: Main
+    QChem_CMD: str = Field(
```
arosen93, line 136: Let's use all-caps for the global settings names.

quacc/custodian/qchem.py
```diff
@@ -0,0 +1,76 @@
...
+            raise ValueError(f"Unknown Q-Chem error handler: {handler_flag}")
+        handlers.append(handlers_dict[handler_flag])
+
+    # Run Q-Chem
+
+    jobs = [
+        QCJob(
+            qchem_command=qchem_cmd,
+            max_cores=qchem_max_cores,
+            calc_loc=qchem_calc_loc,
```
arosen93, line 62: Should we do a check like `if not os.path.exists(qchem_calc_loc)` then `raise ValueError()` to make sure the path exists, or is that checked upstream in `QCJob`?

quacc/settings.py
```diff
@@ -128,6 +128,38 @@ class QuaccSettings(BaseSettings):
...
+        "qchem", description="Command to run the standard version of Q-Chem."
+    )
+
+    QChem_MAX_CORES: int = Field(
+        32, description="Maximum number of cores to use for the Q-Chem calculation."
```
arosen93, line 141: Could you explain what this does when you have a moment? It says "maximum," so is Custodian able to adjust the number of cores on-the-fly for some sort of error correction?

quacc/custodian/qchem.py
samblau: Sure, a check like this makes sense. The one thing to keep in mind is that the chem_calc_loc path (which defaults to /tmp) will most likely only exist on the compute node.

quacc/settings.py
samblau: Sounds good, thanks for making this change!

quacc/settings.py
samblau: That's a good question. When we named "qchem_max_cores" loooooong ago thinking that we may encounter cases where the number of cores would be dynamic. But in practice we never do, so this name would be clearer (at least in this code - I'm not changing it in Custodian) as QCHEM_CORES.

quacc/calculators/qchem.py
samblau: Thanks for taking care of this!

quacc/calculators/qchem.py
samblau: I was just copying from the ASE Q-Chem calculator here... not really sure what the deal is with the ignore_bad_restart_file property, but figured it was safest to conform to what they had done previously.

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,132 @@
...
+            else:
+                mol.set_charge_and_spin(charge=self.charge)
+        qcin = ForceSet(mol, **self.qchem_input_params)
+        qcin.write("mol.qin")
+
+    def read_results(self):
+        data = QCOutput("mol.qout").data
+        self.results["energy"] = data["final_energy"] * units.Hartree
+        tmp_grad_data = []
+        with zopen("131.0", mode="rb") as file:
```
samblau, line 112: FYI @arosen93, was just thinking about this more and realized this file will only be present if we are using Custodian. In fact, the entire approach that I'm taking only makes sense if we're using Custodian. Can we just hard-code that you must use Custodian rather than giving the user the choice not to?

quacc/custodian/qchem.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Handlers for Q-Chem
+    handlers_dict = {
+        "QChemErrorHandler": QChemErrorHandler(),
+    }
+
+    handlers = []
+    for handler_flag in qchem_custodian_handlers:
+        if handler_flag not in handlers_dict:
+            raise ValueError(f"Unknown Q-Chem error handler: {handler_flag}")
```
samblau, line 61: @arosen93 Since there's only one Q-Chem error handler (and I can't really see that ever changing), do we think that this approach is really necessary here?

quacc/calculators/qchem.py
arosen93: I was thinking about this as well. In an ideal world, I would really like to allow the calculator to operate like either a "normal" ASE calculator where it just calls the executable as well as with Custodian (similar to what is done with the VASP calculator). There are multiple philosophical reasons for this that I won't bore you with. However, if it really does not end up being practical or feasible with the setup you have going, I would entertain the idea of having a custom Q-chem custodian calculator, which would ultimately need to be reflected in the name of the class and such.

quacc/custodian/qchem.py
arosen93: If `qchem_custodian_handlers` is an input kwarg, then yeah we should check to make sure the user provided a valid one. However, if that's the only error handler, we can just drop the kwarg entirely and force that as the only handler in the function itself.

quacc/settings.py
arosen93: Ah, I see. We can revisit this later, but my preference would be to mimic the ASE calculator style of adding the number of cores and such to the input arguments of the calculator (e.g. see [here](https://databases.fysik.dtu.dk/ase/ase/calculators/qchem.html)) for consistency rather than having this be some environment variable. This makes it easier to swap settings on-the-fly or for multiple jobs in a workflow.

quacc/custodian/qchem.py
arosen93: Let's rename this to `qchem_scratch_loc` or something similar to be better aligned with the `QCLOCALSCR` environment variable that it inherits from in custodian if set to `None`.

quacc/custodian/qchem.py
```diff
@@ -0,0 +1,74 @@
...
+except ImportError:
+    ob = None
+
+
+@requires(
+    ob,
+    "Openbabel must be installed. Try conda install -c conda-forge openbabel",
+)
+def run_custodian(
+    qchem_cores: int,
```
arosen93, line 23: I think it might be better to have this be `qchem_cores: int = multiprocessing.cpu_count()` (with `import multiprocessing` at the top that way there is a reasonable default always. But we just have to double-check it's getting the CPU count from the remote worker and not the local machine. I think it should be.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,103 @@
...
+    summarize_opt_run,
+    summarize_run,
+    summarize_thermo_run,
+    summarize_vib_run,
+)
+from quacc.util.calc import run_ase_opt, run_ase_vib, run_calc
+from quacc.util.thermo import ideal_gas
+from quacc.calculators.qchem import QChem
+
+def static_job(
```
arosen93, line 22: The way Quacc is currently structured, I like to have some "default" values set that are reasonable. See https://github.com/Quantum-Accelerators/quacc/blob/ea1916e5929db032e0c5b0b3f92ed75c22b031c0/quacc/recipes/gaussian/core.py#L61-L77 for an example in the Gaussian calculator. I also like to have the user be forced to supply certain parameters as kwargs, such as the functional and basis set.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,103 @@
...
+        Dictionary of kwargs for the optimizer.
+
+    Returns
+    -------
+    dict
+        Dictionary of results from quacc.schemas.ase.summarize_opt_run
+    """
+
+    opt_kwargs = opt_kwargs or {}
+    if optimizer == "Sella":
```
arosen93, line 86: Since this doesn't seem specific to Q-Chem, I propose we ultimately move this to `quacc.util.calc` instead.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,103 @@
...
+    atoms.calc = calc
+    atoms = run_calc(atoms)
+    return summarize_run(
+        atoms,
+        input_atoms=input_atoms,
+        additional_fields={"name": "Q-Chem Static"},
+    )
+
+
+def opt_job(
```
arosen93, line 53: For better or for worse, the convention I've gone with is `relax_job` for structure/geometry relaxations, so let's stick with it for consistency's sake.

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,147 @@
...
+                    float(tmp_grad_data[ii * 3 + 2]),
+                ]
+            )
+        if data["pcm_gradients"] is not None:
+            gradient = data["pcm_gradients"][0]
+        else:
+            gradient = data["gradients"][0]
+        for ii, subgrad in enumerate(grad):
+            for jj, val in enumerate(subgrad):
+                assert abs(gradient[ii, jj] - val) < 1e-6
```
arosen93, line 145: `assert` should only be used in test suites. Use an `if` statement here and raise a `ValueError` if there's an issue.

quacc/recipes/qchem/core.py
arosen93, line 22: Before all jobs you'd consider to be a unique Slurm job, please add a `@ct.electron` dectorator above the function name (with `import covalent as ct` at the top). This will allow users to use Covalent as their manager if they wish while still also supporting Jobflow+FireWorks.

quacc/calculators/qchem.py
samblau: Fixed

quacc/recipes/qchem/core.py
samblau: fixed

quacc/custodian/qchem.py
samblau: I think that naming it `qchem_local_scratch_dir` in quacc is fine. I'm a bit hesitant to have that be set by `SCRATCH_DIR` because of the potential confusion between Q-Chem's two different scratch directories - `QCLOCALSCR` and `QCSCRATCH`.

quacc/custodian/qchem.py
samblau: Fixed. Will need to check and make sure it's working in practice once I start testing on LRC.

quacc/custodian/qchem.py
samblau: Dropped the kwarg and global setting entirely

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,514 @@
...
+
+try:
+    from sella import Sella, IRC
+except ImportError:
+    Sella = None
+
+
+def true_charge_and_spin(atoms, charge, mult):
+    mol = AseAtomsAdaptor.get_molecule(atoms)
+    if charge is not None:
```
arosen93, line 26: Not a bad idea. Chuck that into `quacc.util.atoms` so we can use it in other molecular DFT calculators in some way.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,514 @@
...
+    )
+    atoms.calc = calc
+    atoms = run_calc(atoms)
+    return summarize_run(
+        atoms,
+        input_atoms=input_atoms,
+        additional_fields={
+            "name": "Q-Chem Static",
+            "charge": true_charge,
+            "spin_multiplicity": true_spin
```
arosen93, line 130: Can remove these now. `summarize_run` and related functions now take a `charge_and_mult: Tuple[int.int]` keyword.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,627 @@
...
+        atoms,
+        input_atoms=input_atoms,
+        charge_and_multiplicity=(true_charge, true_spin),
+        additional_fields={"name": "Q-Chem Static"},
+    )
+
+
+@ct.electron
+@requires(
+    Sella,
```
arosen93, line 133: While sella is definitely the obvious choice here, it doesn't look like there is any specific need to wrap this or the following job in a `@requires`. Have the default optimizer be: `"optimizer": Sella if Sella else FIRE` and then you can remove the decorator.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,627 @@
...
+    "Sella must be installed. pip install sella",
+)
+def relax_job(
+    atoms: Atoms,
+    cores: int | None = None,
+    charge: int | None = None,
+    mult: int | None = None,
```
arosen93, line 140: FYI: I've changed it to `multiplicity` in all the molecular DFT calculators for consistency/clarity.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,627 @@
...
+    xc: str = "wb97mv",
+    basis: str = "def2-svpd",
+    scf_algorithm: str = "diis",
+    pcm_dielectric: str | None = None,
+    smd_solvent: str | None = None,
+    swaps: dict | None = None,
```
arosen93, line 146: FYI: I've  changed it to `calc_swaps` in all the calculators to distinguish from the newly introduced `opt_swaps` and such.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,627 @@
...
+    for key in swaps:
+        if key not in overwrite_inputs:
+            overwrite_inputs[key] = {}
+            for subkey in swaps[key]:
+                overwrite_inputs[key][subkey] = swaps[key][subkey]
+
+    qchem_input_params = {
+        "basis_set": basis,
+        "scf_algorithm": scf_algorithm,
+        "qchem_version": 6,
```
arosen93, line 226: Maybe best to have in the global settings? With a note that you have to use 6+. Just to make things future-proof/flexible.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,627 @@
...
+        charge=charge,
+        spin_multiplicity=mult,
+        qchem_input_params=qchem_input_params,
+    )
+    atoms.calc = calc
+    dyn = run_ase_opt(atoms, **opt_flags)
+
+    return summarize_opt_run(
+        dyn,
+        check_convergence=check_convergence,
```
arosen93, line 247: Perhaps it makes more sense for us to remove `check_convergence` as a kwarg from all the jobs and, in each summarize function, have `check_convergence: bool = SETTINGS.CHECK_CONVERGENCE` and link that to the global settings, independent of Q-Chem. Thoughts? As always, the user can override it temporarily with an environment variable or permanently by modifying the YAML.

quacc/recipes/qchem/core.py
samblau: So does that mean that everything which calls the Q-Chem calculator should be wrapped in a decorator for openbabel?

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,177 @@
...
+        The spin multiplicity of the molecular system.
+        Effectively defaults to the lowest spin state given the molecular structure and charge.
+    qchem_input_params
+        Dictionary of Q-Chem input parameters to be passed to pymatgen's ForceSet.
+    use_custodian
+        Whether to use Custodian to run Q-Chem.
+        Default is True in settings.
+    **kwargs
+        Additional arguments to be passed to the Q-Chem calculator. Takes all valid
+        ASE calculator arguments, in addition to those custom to Quacc.
```
arosen93, line 42: Not sure this line is relevant here since you aren't inheriting from the ASE Q-Chem calculator. Probably worth clarifying the code comment a bit, e.g. referring to the corresponding Pymatgen section.

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,177 @@
...
+    cores
+        Number of cores to use for the Q-Chem calculation.
+    charge
+        The total charge of the molecular system.
+        Effectively defaults to zero.
+    spin_multiplicity
+        The spin multiplicity of the molecular system.
+        Effectively defaults to the lowest spin state given the molecular structure and charge.
+    qchem_input_params
+        Dictionary of Q-Chem input parameters to be passed to pymatgen's ForceSet.
```
arosen93, line 36: When possible, would you mind referring to the specific module in the docstring where possible to help if someone is just inspecting the class with hover-over in VS Code? i.e. "passed to `pymatgen.io.qchem.sets.ForceSet`"

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,177 @@
...
+        self.charge = charge
+        self.spin_multiplicity = spin_multiplicity
+        self.qchem_input_params = qchem_input_params or {}
+        self.kwargs = kwargs
+        self.default_parameters = {
+            "cores": self.cores,
+            "charge": self.charge,
+            "spin_multiplicity": self.spin_multiplicity,
+        }
+        for key in self.qchem_input_params:
```
arosen93, line 73: Could you add a code comment above this monstrosity? lol

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,177 @@
...
+        self.command = self._manage_environment()
+
+        # Instantiate previous orbital coefficients
+        self.prev_orbital_coeffs = None
+
+        # Instantiate the calculator
+        FileIOCalculator.__init__(
+            self,
+            restart=None,
+            ignore_bad_restart_file=FileIOCalculator._deprecated,
```
arosen93, line 93: Let's remove the `ignore_bad_rest_file` kwarg, assuming nothing dies.

quacc/calculators/qchem.py
```diff
@@ -0,0 +1,177 @@
...
+        self.results["energy"] = data["final_energy"] * units.Hartree
+        tmp_grad_data = []
+        with zopen("131.0", mode="rb") as file:
+            binary = file.read()
+            for ii in range(int(len(binary) / 8)):
+                tmp_grad_data.append(
+                    struct.unpack("d", binary[ii * 8 : (ii + 1) * 8])[0]
+                )
+        grad = []
+        for ii in range(int(len(tmp_grad_data) / 3)):
```
arosen93, line 151: Some brief code comments in this general section would be helpful (for me).

quacc/custodian/qchem.py
```diff
@@ -0,0 +1,88 @@
...
+    Returns
+    -------
+    None
+    """
+    # Adapted from atomate.qchem.firetasks.run_calc
+
+    from custodian.qchem.handlers import QChemErrorHandler
+    from custodian.qchem.jobs import QCJob
+
+    qchem_cores = qchem_cores or multiprocessing.cpu_count()
```
arosen93, line 57: By the way, I've confirmed that if you included this earlier, e.g. in the calculator class, it should still work fine. So, include it where you think it is most natural.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+
+    opt_swaps = opt_swaps or {}
+    opt_defaults = {
+        "fmax": 0.01,
+        "max_steps": 1000,
+        "optimizer": Sella,
+        "optimizer_kwargs": {},
+    }
+    opt_flags = opt_defaults | opt_swaps
+    if "sella.optimize" not in opt_flags["optimizer"].__module__:
```
arosen93, line 322: Same comment here about a cleaner check.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+        raise RuntimeError("Only Sella should be used for TS optimization! Exiting...")
+
+    if pcm_dielectric is not None and smd_solvent is not None:
+        raise ValueError("PCM and SMD cannot be employed simultaneously! Exiting...")
+
+    overwrite_inputs = {"rem": {"method": xc}}
+
+    calc_swaps = calc_swaps or {}
+    for key in calc_swaps:
```
arosen93, line 331: Same comment here about the dict merging with the `|` operator, if possible

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+        "basis_set": basis,
+        "scf_algorithm": scf_algorithm,
+        "qchem_version": 6,
+        "pcm_dielectric": pcm_dielectric,
+        "smd_solvent": smd_solvent,
+        "overwrite_inputs": overwrite_inputs,
+    }
+
+    if scf_algorithm.lower() == "gdm":
+        qchem_input_params["max_scf_cycles"] = 200
```
arosen93, line 347: Same comment here about the max_scf_cycles check

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+
+
+@ct.electron
+@requires(
+    Sella,
+    "Sella must be installed. pip install sella",
+)
+def irc_job(
+    atoms: Atoms,
+    direction: str,
```
arosen93, line 374: type hint should be `Literal["forward", "reverse"]` (`from typing import Literal`)

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+    cores: int | None = None,
+    charge: int | None = None,
+    multiplicity: int | None = None,
```
arosen93, line 377: Let's use `spin_multiplicity`, which is the default elsewhere I think

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+    xc: str = "wb97mv",
```
arosen93, line 378: To nitpick: Do we want to use "method" instead of "xc" in general? Q-Chem can take methods beyond DFT, yes?

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+    basis: str = "def2-svpd",
```
arosen93, line 379: Following up the above comment, should a user be able to provide no basis? If so, we should add a `| None` typehint.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+    scf_algorithm: str = "diis",
+    pcm_dielectric: str | None = None,
+    smd_solvent: str | None = None,
+    calc_swaps: dict | None = None,
+    irc_swaps: dict | None = None,
+    relax_swaps: dict | None = None,
+    check_convergence: bool = True,
```
arosen93, line 517: This function can/should be greatly cleaned up. Refer to [this example](https://github.com/Quantum-Accelerators/quacc/blob/ccb4de931c5ee46ab7c5f3355c52d98c3a41025b/quacc/recipes/emt/slabs.py). Anytime you have a multi-step "job" like you have here, the better option is to pass the `irc_job` and `relax_job` as function kwargs. Then you can get rid of basically all of your kwargs here.

quacc/util/atoms.py
```diff
@@ -305,3 +305,32 @@ def copy_atoms(atoms: Atoms) -> Atoms:
         atoms.calc = calc
 
     return atoms
+
+
+def true_charge_and_spin(
```
arosen93, line 310: Can we rename this to `get_charge_and_spin`? It's not necessarily true :P

tests/util/test_atoms.py
```diff
@@ -141,3 +142,16 @@ def test_get_highest_block():
     assert get_highest_block(atoms) == "f"
     atoms = molecule("H2O")
     assert get_highest_block(atoms) == "p"
+
+
+def test_true_charge_and_spin():
+    atoms = molecule("CH3")
```
arosen93, line 148: ASE already initialized magnetic moments on the CH3, so the correct multiplicity for `molecule("CH3")` is likely set correctly even without your function.

Could you add one more test:

```
atoms = molecule("CH3")
atoms.set_initial_magnetic_moments([0.0]*len(atoms))
```

Copying and pasting your existing test with this additional `atoms` object.

quacc/recipes/qchem/core.py
```diff
@@ -0,0 +1,618 @@
...
+@ct.electron
+def relax_job(
+    atoms: Atoms,
+    cores: int | None = None,
+    charge: int | None = None,
+    multiplicity: int | None = None,
+    xc: str = "wb97mv",
+    basis: str = "def2-svpd",
+    scf_algorithm: str = "diis",
+    pcm_dielectric: str | None = None,
```
arosen93, line 134: Any chance we can chuck `scf_algorithm`, `pcm_dielectric`, and `smd_solvent` into the defaults and make them alterable via a `calc_swaps` swap?

arosen93 (changes requested):
Looks good! I added some more comments here. Sorry for making individual ones before. I forgot to hit "start review" so now there are individual ones and this set.

Please also make sure to update the `CHANGELOG.md`!

quacc/recipes/qchem/core.py
arosen93: Most likely, unless you can think of a cleaner alternative that will quickly tell the user things will go south if they launch the job, with a clear explanation as to how to fix it.

quacc/calculators/qchem.py
samblau: Changed to "Additional arguments to be passed to the FileIOCalculator."

arosen93 (changes requested):
I did a lot of refactoring and think it's in good shape now! I have to sort out this failing unit test because of the `CHECK_CONVERGENCE` argument, but other than that, I think we should be nearly at steady state.

I left a few minor comments for you now that I had time to do a thorough review.

quacc/settings.py
```diff
@@ -43,6 +43,10 @@ class QuaccSettings(BaseSettings):
     GZIP_FILES: bool = Field(
         True, description="Whether generated files should be gzip'd."
     )
+    CHECK_CONVERGENCE: bool = Field(
+        False,  # TODO: CHANGE THIS BACK TO TRUE!!!!
+        description="Whether to check for convergence in the `summarize_run`-type functions, if supported.",
+    )
```
arosen93, line 49: Note to self...

tests/recipes/qchem/test_qchem_recipes.py
arosen93, line 1: Is there any way to easily reduce the testing time for this suite, e.g. for setting the maximum optimization steps to a small, fixed number? Not a big deal if it's difficult, but it's currently the slowest test suite.

arosen93:
Ugh, codecov why do you not update with the most recent commit. Anger!!!

Anyway, I'm confident I can sort out codecov being a jerk. Here is the remaining review comment: https://github.com/Quantum-Accelerators/quacc/pull/393#pullrequestreview-1513184346

arosen93:
@samblau: As noted over Slack, I'm going to go ahead with the merge! Thank you for your amazing work on this. It was already in near-perfect shape, so I figured let's do it!

The changes I made since your last big push were all non-user facing things, so nothing to worry about. There is one exception though. I moved all TS-related recipes to `quacc.recipes.qchem.ts` since `qchem.core` was getting a bit unwieldy on its own. Happy to discuss later if you find that undesirable.

Issue #240, opened 2023-04-14
Add ML calculators (particularly `matgl`)


arosen93:
Ones based on ASE, in particular `matgl`.

arosen93:
Closing since it's unclear how long they'll be maintained, but we'll revisit this if people are interested.

arosen93:
Closed as duplicate.

PR #418, opened 2023-06-13
Add support for a NewtonNet calculator and recipes

kumaranu:
- [x] I have read the [Developer Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html).
- [x] I have added unit tests to cover my changes.
- [x] I have updated the documentation and `CHANGELOG.md` file accordingly.

EDIT by ASR: Partially addresses #240.


src/quacc/recipes/newtonnet/core.py
```diff
@@ -0,0 +1,399 @@
...
+from quacc.util.calc import run_ase_opt, run_calc
+from quacc.util.thermo import ideal_gas
+
+try:
+    from newtonnet.utils.ase_interface import MLAseCalculator as NewtonNet
+except ImportError:
+    NewtonNet = None
+
+
+def get_hessian(atoms):
```
arosen93, line 37: For all "internal" functions (ones that the user isn't expected to use), please rename them with `_` in the front, e.g. `_get_hessian` and `_add_stdev_and_hess`.

src/quacc/recipes/newtonnet/core.py
```diff
@@ -0,0 +1,399 @@
...
+    -------
+    dict
+        A summary of the run, including relevant information about the calculation results.
+    """
+    newtonnet_kwargs = newtonnet_kwargs or {}
+    opt_swaps = opt_swaps or {}
+    input_atoms = deepcopy(atoms)
+    # Define calculator
+    mlcalculator = NewtonNet(
+        model_path=SETTINGS.NEWTONNET_MODEL_PATH.split(":"),
```
arosen93, line 91: Remove the splitting based on ":". You should be able to simply specify a list of paths in the `settings.py` file, and it will be imported automatically as a `list[str]`.

src/quacc/recipes/newtonnet/core.py
```diff
@@ -0,0 +1,399 @@
...
+    return summarize_run(
+        atoms, input_atoms=input_atoms, additional_fields={"name": "NewtonNet Relax"}
+    )
+
+
+@ct.electron
+@requires(
+    NewtonNet,
+    "newtonnet must be installed. Checkout https://github.com/ericyuan00000/NewtonNet",
+)
```
arosen93, line 105: Can you use the same message as above? "NewtonNet must be installed. Try pip install quacc[newtonnet]"

src/quacc/recipes/newtonnet/core.py
```diff
@@ -0,0 +1,399 @@
...
+def relax_job(
+    atoms: Atoms,
+    fmax: float = 0.01,
+    max_steps: int = 1000,
+    optimizer: Optimizer = Sella,
```
arosen93, line 110: This line will cause a crash if `Sella` is not installed.

arosen93 (changes requested):
If you can also address these issues, then we should be good to go once your tests are done!

tests/recipes/newtonnet/config0.yml
arosen93, line 1: Do we need to update this? It's referring to some custom folders in Eric's directory. Not sure why that doesn't crash things.

src/quacc/recipes/newtonnet/core.py
kumaranu: Done with this one. Renamed the functions. Added docstrings and type hints.

tests/recipes/newtonnet/config0.yml
kumaranu: It is fine. Those paths are not used anywhere.

src/quacc/recipes/newtonnet/core.py
kumaranu: How will the input looks like when the ":" is removed in ".quacc.yaml"?
Will it contain the square brackets?

arosen93: I believe so. ["path1", "path2"]

src/quacc/recipes/newtonnet/core.py
kumaranu: Done.

src/quacc/recipes/newtonnet/core.py
kumaranu: Just a reminder that when custom_hessian is asked by the user, Sella is the only optimizer that can do TS optimization in our implementation.

arosen93: Maybe you can do something like Sam did here: https://github.com/Quantum-Accelerators/quacc/pull/393/files#diff-bd8364980ac1eda5810606c2a610911778b2e51df2db8ee501c4f0494cd8398dR194-R199

arosen93: You can do something like Sam did here: https://github.com/samblau/quacc/blob/b4899388f51ec20b42549748802b9f88bd4b64ba/src/quacc/recipes/qchem/core.py#L194-L199

arosen93: @kumaranu: Ah, I see your comment now. Then perhaps we should require the user to have Sella installed. If that's the case, you'll want to add that to the `monty.dev.requires` decorator.

src/quacc/recipes/newtonnet/core.py
kumaranu: Done.

src/quacc/recipes/newtonnet/core.py
```diff
@@ -0,0 +1,416 @@
...
+
+def _get_hessian(atoms: Atoms) -> np.ndarray:
+    """
+    Calculate and retrieve the Hessian matrix for the given molecular configuration.
+
+    This function takes an ASE Atoms object representing a molecular configuration and uses the
+    NewtonNet machine learning calculator to calculate the Hessian matrix. The calculated Hessian
+    matrix is then returned.
+
+    Args:
```
arosen93, line 45: Can you format the docstrings to be numpy format (like the others)? See the `-------` separator, "Parameters" instead of "args", etc.

This will ensure that the docstring renders correctly in the documentation. Thanks!

src/quacc/recipes/newtonnet/core.py
```diff
@@ -0,0 +1,416 @@
...
+    )
+    atoms.calc = mlcalculator
+
+    if use_custom_hessian:
+        if opt_flags["optimizer"].__name__ != "Sella":
+            raise ValueError("Custom hessian can only be used with Sella.")
+
+        opt_flags["optimizer_kwargs"]["hessian_function"] = _get_hessian
+
+    # Define calculator again TEST THIS WHILE RUNNING THE CALCULATIONS
```
arosen93, line 229: Just a reminder about this comment.

PR #434, opened 2023-06-16
Cleaner handling of ASE Optimizer selection

arosen93:


arosen93:
@samblau, @kumaranu: As shown in this PR, I am planning to modify the `optimizer` kwarg in `run_ase_opt()` to take the actual optimizer class as the input rather than a string representation, e.g. `run_ase_opt(optimizer=Sella)` instead of `run_ase_opt(optimizer="Sella")`, which is a much cleaner approach that allows the user to provide basically any optimizer they can import.

PR #442, opened 2023-06-18
Standardized mechanism for storing charge and spin in metadata

### LINKED ISSUE #421: The charge and spin multiplicity is not always passed to the output molecule metadata in the schemas
Take the Gaussian recipes for instance.

https://github.com/Quantum-Accelerators/quacc/blob/b53d4aab505b0f38c2b276a940b44e0ecb8c2194/quacc/recipes/gaussian/core.py#L19-L22

The user can optionally specify `charge` or `mult`, and if they do, it gets passed to the calculator. That means it will (likely) be in `atoms.calc.results`, but it _won't_ necessarily be in the `Molecule` object that is used to make the `MoleculeMetadata` in the `summarize_run` functions.

We need to come up with a way to ensure that the `Molecule` output schema always has the correct charge and multiplicity even if they are just passed as kwargs to the calculator.

Links to this Pymatgen PR: https://github.com/materialsproject/pymatgen/pull/3056

arosen93:
Closes #421.

I've patched in a `.charge` and `.spin_multiplicity` property for the `Atoms` object if it has no PBCs.

```python
import quacc
from ase.build import bulk, molecule

atoms = bulk("Cu")
print(atoms.charge) # --> returns None
print(atoms.spin_multiplicity) # --> returns None

atoms = molecule("CH4")
print(atoms.charge) # --> returns 0
print(atoms.spin_multiplicity) # --> returns 1

atoms = molecule("CH3")
print(atoms.charge) # --> returns 0
print(atoms.spin_multiplicity) # --> returns 2
```


arosen93:
Never mind... closing this and trying a different approach because this breaks (de)serialization and therefore FireWorks. Freaking `MSONable`, I swear........

PR #499, opened 2023-06-28
Use specific `warning` subclasses and add tool features to `pyproject.toml`

arosen93:


janosh:
I usually just apply a blanket warnings filter. don't remember ever encountering a case where seeing the warnings would have been useful.

```toml
[tool.pytest.ini_options]
addopts = "-p no:warnings"
```

arosen93:
Hm... that's a good point. I don't think it's ever helped me either.....

PR #586, opened 2023-07-12
Adding Defect Generation and Processing Capabilities

rwexler:
- [x] I have read the [Developer Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html).
- [ ] I have added unit tests to cover my changes.
- [ ] I have updated the documentation and `CHANGELOG.md` file accordingly.

This pull request introduces two additions to the quacc package, specifically to support defect generation and calculations:

In `quacc/recipes/emt/defects.py`, we introduce the `bulk_to_defects_flow` function. This function provides a workflow to create and relax defects. The function allows users to specify the defect generator, the charge state of the defects, the methods used for relaxation and static calculations, and additional arguments for these processes.

In `quacc/util/defects.py`, two utility functions are added: `get_defect_entry_from_defect` and `make_defects_from_bulk`. `get_defect_entry_from_defect` is a helper function that creates a `DefectEntry` object from a `Defect` object, capturing information about a defect in a specific supercell structure. On the other hand, `make_defects_from_bulk` generates a list of defects. The function prepares a supercell for each defect, applies distortions using ShakeNBreak, and returns the defects as ASE `Atoms` objects.


arosen93:
@rwexler --- thank you for opening this PR! How can I help get this to the finish line? Are you looking for guidance on the code itself, suggestions of how to write unit tests, or none of the above? Let me know!

rwexler:
@arosen93, thanks for writing this code; I'm excited to contribute.

I would say that my goals are as follows:

1. Get this code to work remotely on my server (can you confirm that this worked on yours?)
2. Get feedback on the code and suggestions for writing unit tests
3. Extend to calculations with VASP

.github/workflows/tests.yaml
```diff
@@ -214,11 +214,56 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
 
+  tests-defects:
```
arosen93, line 217: FYI: I added this to the GitHub Action workflows. You'll see that there is now a dedicated `tests-defects` workflow that will test all codes with the name "defects" in them. The reason I have this as a separate test suite is that there are additional dependencies your code requires, so I have it in its own little environment. You wouldn't have been expected to do this or know to do this, but I'm letting you know more just in case you're curious.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+defects = ["pymatgen-analysis-defects>=2023.7.12", "shakenbreak>=23.6.23"]
```
arosen93, line 39: FYI: I have added a set of "extras" to the installation process so that if you do `pip install -e .[defects]`, it would install the defect dependencies. This is helpful for users.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
+"""Defect recipes for EMT"""
```
arosen93, line 1: FYI: I formatted your code with `isort .` and then `black .` per the style quacc uses, in case you're curious. Most packages have a strict requirement to match their formatting, but I don't care --- it'll get reformatted on merge. Just letting you know though.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+def bulk_to_defects_flow(
+    atoms: Atoms | dict,
+    defectgen: (
+        AntiSiteGenerator
+        | ChargeInterstitialGenerator
+        | InterstitialGenerator
+        | SubstitutionGenerator
+        | VacancyGenerator
+        | VoronoiInterstitialGenerator
+    ) = VacancyGenerator,  # NOTE: I added a default. VacancyGenerator seemed like it might be the most popular.
```
arosen93, line 29: I added a default value for `defectgen` here because that's nice for the user. I chose `VacancyGenerator` because I thought that'd be the most popular option but you can change it if you'd like.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+    charge_state: int = 0,  # NOTE: I changed this from int | None = None because None does not work
```
arosen93, line 30: This originally was `charge_state: int | None = None`, which led to an error because `charge_state` had to be an integer. I have updated it.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+    defect_static: ct.electron | None = static_job,
+    defect_relax_kwargs: dict | None = None,
+    defect_static_kwargs: dict | None = None,
+) -> list[RunSchema | OptSchema]:
+    """
+    Workflow consisting of:
+
+    1. Defect generation
+
+    2. Defect relaxations
```
arosen93, line 42: I have made the relaxation step required because there are very few conceivable scenarios where someone would want to make the defect and do only a static calculation.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Parameters
+    ----------
+    atoms
+        Atoms object for the structure.
+    defectgen
+        Defect generator
+    charge_state
+        Charge state of the defect
+    make_defects_kwargs
```
arosen93, line 54: I have renamed this kwarg to better describe its purpose.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+        Keyword arguments to pass to the make_defects_from_bulk
+    defect_relax
```
arosen93, line 56: I have removed the `_electron` suffix here and in the following kwarg because I think it's cleaner. I did the same in `main` recently, so this is to just make it consistent.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+    @ct.lattice
+    def _relax_distributed(defects):
+        return [defect_relax(defect, **defect_relax_kwargs) for defect in defects]
+
+    @ct.electron
+    @ct.lattice
+    def _relax_and_static_distributed(defects):
+        return [
+            defect_static(
+                defect_relax(defect, **defect_relax_kwargs),
```
arosen93, line 85: I have passed in `defect` here instead of `defect["atoms"]` because quacc recipes can take in either the Atoms object or a dictionary containing the Atoms object. The latter is a little bit cleaner.

src/quacc/recipes/emt/defects.py
```diff
@@ -0,0 +1,98 @@
...
+                **defect_static_kwargs,
+            )
+            for defect in defects
+        ]
+
+    defects = ct.electron(make_defects_from_bulk)(
+        atoms, defectgen, charge_state=charge_state, **make_defects_kwargs
```
arosen93: The `defectgen` and `charge_state` are both keyword arguments, not positional arguments, so I made them with `=` signs.

src/quacc/util/defects.py
```diff
@@ -0,0 +1,174 @@
...
+# - Anytime an Atoms object is converted to a pmg structure, make sure
+# to reattach any .info flags to the Atoms object, e.g. via `new_atoms.info = atoms.info.copy()``.
+# Note that atoms.info is mutable, so copy it!
+# - All major functions should take in Atoms by default and return Atoms
+# by default. Pymatgen structures can be returned with an optional kwarg.
+# - If you modify the properties of an input Atoms object in any way, make sure to do so
+# on a copy because Atoms objects are mutable.
+
+
+def make_defects_from_bulk(
```
arosen93, line 34: I removed the `DummySpecies` as a user-supplied option. I think it's safe for us to assume that "X" should be used. Happy to revert if you strongly disagree (perhaps if there are multiple types of dummy species? Does that ever happen?).

src/quacc/util/defects.py
```diff
@@ -0,0 +1,174 @@
...
+                "distortions": distortions,
+                "bulk": atoms,
+                "defect": defect,
+            }
+            final_defect.info["defect_stats"] = defect_stats
+            final_defects.append(final_defect)
+    return final_defects
+
+
+def _get_defect_entry_from_defect(
```
arosen93, line 133: I renamed this with a `_` prefix to indicate that this isn't really a user-facing function.

tests/recipes/emt/test_emt_defect_recipes.py
```diff
@@ -0,0 +1,42 @@
...
+            os.remove(f)
+        if "quacc-tmp" in f or "job_" in f or f == "tmp_dir":
+            if os.path.islink(f):
+                os.unlink(f)
+            else:
+                rmtree(f)
+
+
+@pytest.mark.skipif(bulk_to_defects_flow is None, reason="requires quacc[defects]")
+def test_bulk_to_defects_flow():
```
arosen93, line 31: I made a test file for you. This is where you will need to write your unit tests. Note how it is in the `tests` directory, which is where `pytest` will look for tests to run.

Every unit test must be a function starting with the name `test_` so that pytest knows it is a test. All tests must have an `assert` statement that pytest will ensure must be True or an error will be raised. I made a very very simple assertion for you below, but you'll need to greatly expand upon this to test for various things in the output (see the code comment for details).

I wrapped the test function in a decorator that tells pytest not to run the test if you don't have all the dependencies installed. There is also a `cleanup_module` function that runs after the test suite is done, mainly cleaning up files and folders made at runtime. You don't have to touch that.

arosen93 (commented):
Hi @rwexler! This looks great. Thank you for your excellent work on this.

Honestly, it is in nearly perfect shape. The main thing that you'll need to do to get this merged is to write your unit tests. I left you a comment about that, but if you have any questions about it, I'm happy to chat more (over any platform, email/slack/github/zoom whatever). It should be somewhat self-explanatory with my example though. To run the tests locally, go into the `tests/recipes/emt` directory and run `pytest test_emt_defect_recipes.py` to see if they pass or fail. You'll need to `pip install pytest` first, I should note. The toy tests I made will pass, but you'll of course need to add more.

When I ran this on Slurm, the job failed not because of anything related to your workflow or quacc (I didn't get the error you received). I got the username error described in https://github.com/AgnostiqHQ/covalent-slurm-plugin/issues/70. Unfortunately, that is a bug in Covalent where dynamic workflows won't run properly on Slurm. However, I just spoke with Will at Covalent, and he said he has time to work on Slurm fixes in the next few days and this was the top priority item. So, hopefully we can rely on them to do a fix at which point doing a `pip install --upgrade covalent-slurm-plugin` will resolve things. Honestly, by the time you write tests and/or decide to write the VASP analogue (in a separate PR, I'd recommend), it'll probably/hopefully be solved.

As for next steps:

- Please go through my comments below. Resolve any of them that are no longer relevant or that you read and required no further response. For any that warrant a response, add a comment. 
- Pull the recent changes (`git pull`) to your local copy so you have everything up to date.
- Commit and push any further changes you'd like, which will update here automatically.
- Write some unit tests! You should do this for both the `quacc.util.defects` function and the recipe itself. The idea of unit tests is that they should be tests for as minimal pieces of code as possible (i.e. "units" of work, not functional tests). So, it's worth adding tests to both functions you added where you `assert` that certain outputs are as you expect. Also, since the unit tests run on GitHub eventually, we want them to be as short as possible while still being representative of real use cases. That's why I had the relaxation job be "converged" with an `fmax=5`. It ran quicker and gives the same info.
- Once the tests are written and high coverage is achieved, I will look it over once more and then we can merge this! The codecov comment above will update with each commit to tell you the total code coverage you have. It's been a bit buggy lately, but hopefully it'll work right for you.
- Then you can open a separate PR for the analogous VASP workflow

Note: It will make the process a lot easier in the future if you don't make commits in your `main` branch but rather have a custom branch for everything. You'll see what I mean if you haven't already :)

PR #664, opened 2023-08-12
Optimization schemas renamed to Dynamic schemas

exenGT:
- Renamed optimization schemas to dynamic schemas.
- Changed multiple names in summarize_opt_run() in src/quacc/schemas/ase.py.

### DIFF
--- a/src/quacc/schemas/ase.py
+++ b/src/quacc/schemas/ase.py
@@ -11,6 +11,7 @@ from ase.atoms import Atoms
 from ase.constraints import Filter
 from ase.io import Trajectory, read
 from ase.optimize.optimize import Optimizer
+from ase.md.md import MolecularDynamics
 from ase.thermochemistry import IdealGasThermo
 from ase.vibrations import Vibrations
 from atomate2.utils.path import get_uri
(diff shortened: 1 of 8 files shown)


src/quacc/schemas/ase.py
```diff
@@ -270,7 +271,10 @@ def summarize_opt_run(
     """
 
     additional_fields = additional_fields or {}
-    opt_parameters = dyn.todict() | {"fmax": dyn.fmax}
+    if isinstance(dyn, Optimizer):
+        dyn_parameters = dyn.todict() | {"fmax": dyn.fmax}
+    if isinstance(dyn, MolecularDynamics):
```
arosen93, line 276: Perhaps this line would be better if it is an `else` so that we are more general.

arosen93 (changes requested):
@exenGT: Thank you for highlighting this. I agree that we should refer to the base class (dynamics) so that things are more general.

I left one suggested change.

Note: This name change is a breaking change, so I will refrain from merging it for the time being.

arosen93:
@exenGT: Thanks again for your comment on this. I'm going to close this PR for now because it is a breaking change for users with the only gain being a clearer function name.

For now, I will simply update the docstring to emphasize that it supports the dynamics class.

PR #719, opened 2023-08-21
Enable direct parsing of Pymatgen-formatted VASP input sets

### LINKED ISSUE #718: Use Pymatgen YAML for VASP MP sets
The problem right now is that we have a custom [MPScanSet.yaml](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/presets/vasp/MPScanSet.yaml) that is effectively a mirror of what's in Pymatgen. This is not ideal because we don't want to duplicate work and there is also a very real chance that there are changes upstream that won't be accounted for here.

The main limiting factor is that the syntax used in the quacc preset YAMLs is based (mostly) on ASE input arguments, so the syntax for all the Pymatgen VASP set parameters is not necessarily 1:1 comparable. This can be solved, however, by making a very simple Pymatgen --> Quacc VASP YAML adapter (e.g. convert all INCAR flags to lowercase and any entries that are not INCAR flags can have a special dictionary mapping). Then we can remove the MP YAML here and simply point to the one shipped with Pymatgen since it's a dependency of quacc anyway.

Update: Solved in #719, pending a bug fix for Python 3.9.

arosen93:
Closes #718.


arosen93:
I will revisit this later. There's some subtleties to sort out to make this more compatible, although it works with `MPSCANRelaxSet`.

PR #859, opened 2023-09-02
Remove redundant `Path()` calls when using `.resolve()`

arosen93:
FYI @OmAximani0.


OmAximani0:
> FYI @OmAximani0.

Got it :+1:

PR #921, opened 2023-09-15
Add Q-Chem Freq

samblau:
Extends Q-Chem calculator and recipes to enable Frequency calculations


src/quacc/calculators/qchem.py
```diff
@@ -199,3 +214,27 @@ def read_results(self):
                 struct.unpack("d", binary[ii * 8 : (ii + 1) * 8])[0]
                 for ii in range(len(binary) // 8)
             )
+        if self.job_type == "freq":
+            tmp_hess_data = []
+            # Read Hessian scratch file in 8 byte chunks
+            with zopen("132.0", mode="rb") as file:
+                binary = file.read()
+                for ii in range(int(len(binary) / 8)):
```
Andrew-S-Rosen, line 222: Are you sure you want to use `int` here and not `round`? Note that `int(4.7)` goes to 4, not 5.

src/quacc/calculators/qchem.py
```diff
@@ -199,3 +214,27 @@ def read_results(self):
...
+                    tmp_hess_data.append(
+                        struct.unpack("d", binary[ii * 8 : (ii + 1) * 8])[0]
+                    )
+            hess_dim = len(data["species"])*3
+            self.results["hessian"] = np.zeros(shape=(hess_dim, hess_dim))
+            for ii, val in enumerate(tmp_hess_data):
+                self.results["hessian"][ii%hess_dim][int(ii/hess_dim)] = val
```
Andrew-S-Rosen, line 229: I don't really understand what's going on here. Can you explain this a bit? Are you just restructuring the data? If so, perhaps it makes more sense to use a numpy reshape function for that rather than iterating over values?

Andrew-S-Rosen (commented):
Looks great! Can you address these small comments? Then we're good to go.

PR #937, opened 2023-09-16
Clean up Q-Chem recipes

Andrew-S-Rosen:
@samblau: I refactored the Q-Chem recipes since the `static_job` and `freq_job` were nearly identical, with the exception of the `job_type`, `additional_fields`, and the basis set kwarg. They now both operate from a common base function to minimize code duplication.

I also added `qchem_internal_relax_job` that does a geom opt using Q-Chem's optimizers. Obviously we think this is going to be less efficient than Sella, but since it was so trivial to add, I included it.


Andrew-S-Rosen:
Never mind. Wasn't satisfied with the end result.

PR #945, opened 2023-09-16
Expose TRICs

samblau:
Make it easy for the user to turn on translation and rotation internal coordinates (TRICs) when using the Sella optimizer. TRICs are particularly important when optimizing complexes of multiple molecules held together by non-covalent interactions, for which traditional bond, angle, and dihedral internal coordinates are insufficiently expressive to allow fragments to translate and rotate with respect to each other.


src/quacc/utils/calc.py
```diff
@@ -161,6 +161,18 @@ def run_ase_opt(
     if relax_cell and atoms.pbc.any():
         atoms = ExpCellFilter(atoms)
 
+    if optimizer_kwargs["use_TRICs"] and optimizer.__name__ != "Sella":
```
Andrew-S-Rosen, line 164: Can probably condense this to:

```python
    if optimizer_kwargs["use_TRICs"]:
        if optimizer.__name__ != "Sella":
            msg = "Can only use translation rotation internal coordinates aka TRICs with Sella!"
            raise ValueError(msg)
        optimizer_kwargs.pop("use_TRICs")
        from sella import Internals
        ...
```

Andrew-S-Rosen, line 164: Note: You mean `optimizer_kwargs.get("use_TRICs")` so that it works when that key isn't present at all.

src/quacc/utils/calc.py
```diff
@@ -161,6 +161,18 @@ def run_ase_opt(
...
+        msg = "Can only use translation rotation internal coordinates aka TRICs with Sella!"
+        raise ValueError(msg)
+    elif optimizer_kwargs["use_TRICs"] and optimizer.__name__ == "Sella":
+        optimizer_kwargs.pop("use_TRICs")
+        from sella import Internals
+        ints = Internals(atoms,allow_fragments=True)
+        ints.find_all_bonds()
+        ints.find_all_angles()
+        ints.find_all_dihedrals()
+        optimizer_kwargs["internal"] = ints
```
Andrew-S-Rosen, line 174: Any situation where the user would be providing `internal` manually?

Andrew-S-Rosen:
Another thought: I'm not sure how many of these kinds of "optional" Sella kwargs there may be. If there happens to be a bunch, it could make sense to add a `sella_kwargs` kwarg to `run_ase_opt`. If there aren't though, the current proposal definitely makes more sense. Just some food for thought either way.

src/quacc/utils/calc.py
samblau: Yes. Added another conditional to check for this.

src/quacc/utils/calc.py
samblau: Yup, thanks

samblau: Makes sense - changed. Thanks

PR #954, opened 2023-09-18
Update schema in Q-Chem calculator

### LINKED ISSUE #942: Clarify Q-Chem output schema
Currently, the output schema for Q-Chem is unclear to the end-user. For instance, some units are ASE-compatible whereas others aren't, and there is no documentation on which is which.

This is in contrast with things like the VASP schema, where everything is relying on an emmet pydantic model that has clear documentation.

Andrew-S-Rosen:
Closes #942.


Andrew-S-Rosen:
Review requested from @samblau.

PR #1087, opened 2023-10-20
Add an `additional_fields` kwarg to all jobs

Andrew-S-Rosen:
- [X] I have read the [Developer Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉
- [X] I have added relevant unit tests. Note: Your PR will likely not be merged without proper tests.

The purpose of this PR is to add an `additional_fields` kwarg to all jobs. This is primarily meant to assist with provenance tracking. For instance, if you are running a workflow on a material "MP-10", you could set `additional_fields={"tag": "MP-10"}` for the relevant jobs. Several workflow engines support tagging, but we want to provide an option for this independent of the workflow engine in case the user doesn't run with one or the workflow engine doesn't support it.

### DIFF
--- a/src/quacc/recipes/dftb/core.py
+++ b/src/quacc/recipes/dftb/core.py
@@ -28,6 +28,7 @@ def static_job(
     method: Literal["GFN1-xTB", "GFN2-xTB", "DFTB"] = "GFN2-xTB",
     kpts: tuple | list[tuple] | dict | None = None,
     calc_swaps: dict | None = None,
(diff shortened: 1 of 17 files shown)


Andrew-S-Rosen:
@samblau: Any thoughts? The implementation here is trivial, but I'm curious if you would prefer a different approach (or maybe you don't care because you do the tagging via Jobflow/FireWorks).

PR #1189, opened 2023-11-12
Update to use the new NewtonNet 1.1 release

Andrew-S-Rosen:
Tagging @kumaranu.


Andrew-S-Rosen:
@kumaranu: The new version of NewtonNet (1.1) has a breaking change. Namely, before the calculator had `method="autograd"` as the default. Now, there is no `method` kwarg and in its place there is `hess_method` which is equal to `None` by default.

To make sure your recipes were backwards-compatible, I needed to add `hess_method = "autograd"` to all recipes (except when `use_custom_hessian=True`, in which case we set `hess_method=None`).

PR #1238, opened 2023-11-22
Add espresso recipes and custom calculator

### LINKED ISSUE #1224: First steps for espresso recipes

Very nice project,

I would like to add some recipes for Quantum Espresso, especially for other executable ph.x, q2r.x, matdyn.x, bands.x and neb.x. I quickly created a quaacc.recipes.espresso.core module (have a look on my fork on the espresso branch)

I tried with covalent and everything seems to work fine, on my local machine anyway. Before going further, I am not sure what is the correct way to deal with parameters, pseudopotentials etc... should it be passed into each function call? should the pp_path be set using the ESPRESSO_PSEUDO env variable, etc...

Some guidances for 'best practices' would be highly appreciated!

tomdemeyere:
- [x] I have read the [Developer Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉

A first PR for espresso additions, this is very much draft stage and does not aim to work, just to show the general idea. Right now I have other things to do but I believe this is a good first draft. There will be probably a thousand details that you don't like in this PR (variable names, etc...) so I will let you have a look. I will come back to it later on.

The aim is to make it simple for the future when adding other espresso binaries, for that I created mini-classes based on a more generic 'EspressoCalculator' class. The process to add a binary would be as follows:

1. Add a class in quacc.calculators.espresso.calculator (few lines)
2. Add a Fortran keylist in quacc.calculators.espresso.keys
3. add a write_BIN_specific function into quacc.calculators.espresso.io as well as read_bin_output 

And that's it, there is nothing else to change, everything is done under the hood. Stage 3 is by far the most time-consuming, 1 and 2 shouldn't take more than a few minutes of work, and GPT is able to provide 2 if needed based on the binary doc page.

Presets are a good idea, I added a quick one based on the cutoff table given by the SSSP library. I am thinking about adding more specific ones in the future (slabs, bulk...)

I think for the pp we shouldn't deal with env variables other than the QUACC ones since I am not aware of how different workflow managers deal with them. Covalent for example always uses env variables defined when the server starts and it does not seem intuitive to change them on the fly.

ASR edit: Closes #1224. Closes #1306. Closes #1307.


Andrew-S-Rosen:
@tomdemeyere: Fantastic! Thank you very much. I look forward to giving this a look shortly.

src/quacc/calculators/espresso/keys.py
```diff
@@ -0,0 +1,168 @@
+# We work ase.io.espresso style, but we need to add the namelist keys for each binary.
+# !!! For now this is GPT3.5 generated, must be checked !!!
```
Andrew-S-Rosen, line 2: Amazing. Love it.

Andrew-S-Rosen (commented):
Hi @tomdemeyere: Thank you for starting this PR! This already looks like it's on the way to a wonderful contribution.

From the recipe and settings side of things, everything looks good. I have no major comments there (just left a few minor comments).

My main question, which is based a bit in naivety because I'm not familiar with the ASE version, is what the purpose is for making the custom calculator as opposed to using or inheriting from ASE's `Espresso` one. Would you be able to walk me through your logic on this? I'm not necessarily opposed to a custom calculator, but it adds a fair bit of maintenance burden, so I want to make sure it is essential. I also left a few big-picture questions about the calculator, e.g. why not inherit from `GenericFileIOCalculator`. I haven't given the calculator a thorough review yet in light of these questions.

Happy to answer any other questions you may have!

src/quacc/recipes/espresso/core.py
```diff
@@ -0,0 +1,133 @@
...
+    Returns
+    -------
+    RunSchema
+        Dictionary of results from [quacc.schemas.ase.summarize_run][]
+    """
+
+    calc_defaults = {
+        "ecutwfc": 40,
+        "ecutrho": 160,
+        "nspin": 1,
```
Andrew-S-Rosen, line 66: I would recommend setting something like `"nspin": 2 if atoms.has("initial_magmoms") else 1`

src/quacc/recipes/espresso/core.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # This will be useful if we add presets...
+    input_data = calc_swaps.get("input_data", {})
+    input_data = construct_namelist(input_data)
+    input_data = {k: dict(v) for k, v in input_data.items()}
+    input_data['control']['pseudo_dir'] = ESPRESSO_PP_PATH
+
+    if preset:
+        config = load_yaml_calc(ESPRESSO_PRESET_PATH / f"{preset}")
+        preset_pp = parse_pp_and_cutoff
```
Andrew-S-Rosen, line 118: If we decide that a new calculator is indeed the best approach, I would recommend adding `preset` as a kwarg of the calculator and handling the parameter updating business there. That's what I did for the (slightly over-engineered) custom `Vasp` calculator [here](https://github.com/Quantum-Accelerators/quacc/blob/dca0cdec3be183877504d7849e073c81733d7ba1/src/quacc/calculators/vasp/vasp.py#L233-L243).

src/quacc/calculators/espresso/core.py
```diff
@@ -0,0 +1,31 @@
+from quacc.calculators.espresso.utils import construct_namelist
+from quacc.calculators.espresso.io import _write, _read
+
+# Very simple class that can run calculations ASE style
+# I don't think we even need atoms object here.
+class EspressoCalculator:
```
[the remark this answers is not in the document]
tomdemeyere: The class name is misleading here, this 'Calculator' is made to deal with other Quantum Espresso binaries (ph.x, neb.x, pp.x, ...) when ASE cannot help (at least not without major tweaking). These binaries do not need atoms objects, most of them are always done after a pw.x calculation. The plan here is to let quacc use ase.calculators.espresso without doing major changes to do pw.x jobs. But each time a user wants to integrate other binaries in a workflow quacc will have to use this special Calculator (which name should probably be changed)

I think there will be no real advantage to make this class inherit from genericfileio?

src/quacc/recipes/espresso/core.py
tomdemeyere: I was hoping that a new calculator would not be needed for pw.x, but I understand that it is annoying to have all that in _base_job... I will come back to it after a first successful run

Andrew-S-Rosen: I understand more now! This isn't a problem. If you end up making multiple "base" job functions, you can always abstract that part away into a separate function.

src/quacc/calculators/espresso/core.py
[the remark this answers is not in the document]
Andrew-S-Rosen: Ah, thank you for the explanation! I've never run QE calculations before, so this is very helpful. I will take another look with this mind after Thanksgiving :)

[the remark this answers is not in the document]
tomdemeyere: QE has 30+ binaries like that. Although it can be quite dense, it is somewhat perfect for workflows... (I am actually surprised that no one did that before, maybe I am not so well informed). Enjoy Thanksgiving!

src/quacc/recipes/espresso/core.py
```diff
@@ -0,0 +1,132 @@
...
+        config = load_yaml_calc(ESPRESSO_PRESET_PATH / f"{preset}")
+        preset_pp = parse_pp_and_cutoff(config, atoms)
+        input_data = input_data | preset_pp['input_data']
+        calc_swaps = calc_swaps | preset_pp['pseudopotentials']
+
+    calc_swaps["input_data"] = input_data
+
+    profile = EspressoProfile(argv=ESPRESSO_CMD.split())
+    atoms.calc = Espresso(profile = profile,
+                          **calc_swaps)
```
Andrew-S-Rosen, line 124: You probably want to do something like most of the other recipes where you have `calc_flags = merge_dict(calc_defaults, calc_swaps)` and then pass `**calc_flags` to the calculator. Doing so allows you to define some default parameters that can be easily overriden by the user via the **kwargs of the recipe.

src/quacc/recipes/espresso/core.py
```diff
@@ -0,0 +1,132 @@
...
+    # and stick with it, which will be the nested dict here.
+    input_data = calc_swaps['input_data']
+    input_data = construct_namelist(input_data)
+    input_data['control']['pseudo_dir'] = ESPRESSO_PP_PATH
+
+    if preset:
+        config = load_yaml_calc(ESPRESSO_PRESET_PATH / f"{preset}")
+        preset_pp = parse_pp_and_cutoff(config, atoms)
+        input_data = input_data | preset_pp['input_data']
+        calc_swaps = calc_swaps | preset_pp['pseudopotentials']
```
Andrew-S-Rosen, line 118: Is `input_data` a nested dict? If so, be careful, as the `|` operator will only merge based on primary key, which is often not what you want. The `quacc.utils.dicts.merge_dicts` function will do a nested dict merge if that's more appropriate. Feel free to resolve this comment if it's not relevant.

src/quacc/settings.py
```diff
@@ -121,6 +122,42 @@ class QuaccSettings(BaseSettings):
...
+    )
+    ESPRESSO_MATDYN_CMD: Path = Field(
+        Path("matdyn.x"),
+        description=("Path to the espresso matdyn.x executable."),
+    )
+    ESPRESSO_DYNMAT_CMD: Path = Field(
+        Path("dynmat.x"),
+        description=("Path to the espresso dynmat.x executable."),
+    )
+
```
Andrew-S-Rosen, line 160: Are these executables typically found within a single directory? If so, perhaps it would be easier to have a single variable (e.g. `ESPRESSO_EXE_DIR`). That would help later since if there are a lot more executables one might want to add, they won't have to enumerate each one as a setting. However, if that's not the typical structure of QE's executables, then this comment isn't relevant. (As you can tell, I have not used QE before).

src/quacc/recipes/espresso/neb.py
```diff
@@ -0,0 +1,86 @@
...
+        Dictionary of results from [quacc.schemas.ase.summarize_run][]
+    """
+    # List must be size >= 2
+    # Preset are assumed to be for all atoms in list
+
+    neb_calc = Neb(neb_input_data,
+                   pw_input_data)
+    # For some reason this function needs an Atoms object
+    # we do that for now
+    _, tmpdir, job_results_dir = calc_setup(Atoms(), copy_files=copy_files)
```
Andrew-S-Rosen, line 73: Ooops! Very good catch. Looks like that was leftover from a refactoring effort.

Patched in #1242.

src/quacc/calculators/espresso/calculators.py
```diff
@@ -0,0 +1,127 @@
...
+        self.input_data = input_data
+        super().__init__()
+
+class Neb(EspressoCalculator):
+
+    # I think it makes sense to have the defaults
+    # here. They are not supposed to change.
+    # Users can have a look at them by doing:
+    # from quacc.calculators.espresso.calculators import Neb
+    # Neb._calc_defaults without looking at the source code
```
Andrew-S-Rosen, line 26: Hmmm! This is one I'm not 100% confident about. Perhaps it would make more sense to have as a global variable in `recipes.espresso.neb` (i.e. outside the scope of any particular function). This will still allow importing of the defaults (`from quacc.recipes.espresso.neb import _BASE_NEB_PARAMS` or something) while still making it immediately visible to users.

I have purposefully tried to keep all defaults within the recipe modules because that is often where people will immediately look for details (e.g. "oh, a relax_job... let me open that up to see what it does"). If things are _never_ supposed to be changed by the user, I care a bit less though so could be swayed. 

Regardless, I would recommend making the `num_of_images` accessible via a kwarg in the recipe (e.g. `n_images`) since that's one where users typically want to adjust things.

src/quacc/calculators/espresso/calculators.py
```diff
@@ -0,0 +1,127 @@
...
+        super().__init__()
+        self.input_file = 'neb.dat'
+
+    # It would be nice to be able to run sanity_checks BEFORE
+    # the calculation is submitted to any kind of HPC
+    # This could also raise warnings to remind the user
+    # about known problems i.e. if first_last_opt is not True
+    # and the first and last images contains spin state. Running
+    # the first singlepoint with fresh wavefunctions might end up
+    # in a different spin state, screwing up the whole NEB calculation.
```
Andrew-S-Rosen, line 58: There are cases where one wants to run an NEB calculation where the initial and final spin states differ. I do this quite frequently, e.g. during spin-crossover reactions or other multi-state reactivity reaction coordinates. So, I'd be hesitant about enforcing the same spin manifold here (although a sanity check to avoid _accidental_ scenarios with the wavefunctions does seem useful).

src/quacc/calculators/espresso/core.py
```diff
@@ -0,0 +1,48 @@
+from quacc.calculators.espresso.utils import construct_namelist
+from quacc.calculators.espresso.io import _write, _read
+from quacc.utils.files import load_yaml_calc
+
+# Very simple class that can run calculations ASE style
+# I don't think we even need atoms object here.
+class EspressoCalculator:
```
Andrew-S-Rosen, line 7: You'll probably want to come up with a slightly clearer name for this so as to distinguish it from the main ASE calculator. Not quite sure yet.

src/quacc/calculators/espresso/calculators.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Probably the best way is to have a recipe that only does sanity checks
+    # with a different runner (local) and then the actual calculation with
+    # the HPC runner. This way we can also check if the calculation is
+    # going to fail.
```
Andrew-S-Rosen, line 63: If I'm understanding your comment correctly, I'll note that this is something I have not yet fully figured out how to do smoothly in a programmatic way.

In principle, if quacc _only_ supported Covalent (for instance), one could do a `@ct.electron(executor="local")` around the relevant function that is basically no-cost. However, because quacc supports several workflow engines out of the box and they all specify executors in slightly different ways, there is no clean mechanism for doing this at the moment (since the kwargs of the `@job` decorator will differ for each workflow engine).

There are probably some ideas I can pull from the [psiflow](https://github.com/molmod/psiflow) project, which uses a YAML file to have the user define multiple configurations (e.g. one for local, one for simulations, one for ML) that could be pulled from, but that's not something I've had time to try to sort out.

I should note that simply not including a decorator at all would make it run locally, but this isn't viable because the inputs to the function might be a `Future` from another function call, which would only be resolved if you were using a `@job`.

src/quacc/calculators/espresso/core.py
```diff
@@ -0,0 +1,48 @@
+from quacc.calculators.espresso.utils import construct_namelist
```
Andrew-S-Rosen, line 1: General comment: This pseudo-calculator seems very useful! If this is something you would like to host externally (e.g. to enable use in other projects without relying on `quacc` as a dependency), certainly feel free to abstract it away to its own package. We can add it as a dependency here in quacc. I don't think the "calculator" interface is using anything from quacc specifically aside from the YAML reader. Admittedly, this is me inquiring if there's a way to offload some of the maintenance burden on the calculator side (where my expertise is not as strong as the recipes), but I am not opposed to having it here if that's what you think is best.

src/quacc/calculators/espresso/keys.py
```diff
@@ -0,0 +1,168 @@
...
+
+from ase.io.espresso import Namelist
+
+PH_KEYS = Namelist([
```
Andrew-S-Rosen, line 6: Would you be able to briefly explain what the purpose is of this namelist stuff? Thanks. I trust it makes sense but haven't dug through the ASE docs for this yet. [Of course, if there are ways we can avoid direct enumeration of things, that'd be great, but I imagine you are well-aware of that].

src/quacc/calculators/espresso/core.py
```diff
@@ -0,0 +1,48 @@
...
+from quacc.calculators.espresso.io import _write, _read
+from quacc.utils.files import load_yaml_calc
+
+# Very simple class that can run calculations ASE style
+# I don't think we even need atoms object here.
```
Andrew-S-Rosen, line 6: Probably don't need an `Atoms`. Sidenote: I know that the `EspressoProfile` mechanism in ASE lets you define a custom executable (so it can be things other than pw.x) and you might even be able to execute it without an `Atoms` object by doing `calc.run()` or something like that. But I imagine it can't do all the things you're trying to do here. I haven't looked at the new profile interface to QE  much.

Andrew-S-Rosen (commented):
Thanks!! I dropped a few comments here since I had some time to give it another look. Overall, it's looking good.

src/quacc/calculators/espresso/calculators.py
tomdemeyere: It happens that I optimise the initial and final state to low forces and then when I launch the NEB calculation one/both of them suddenly have high forces due to a different spin state. What I means here is that if we detect that initial/final states have spin and that 'first_last_opt' is not true (meaning that NEB will consider initial and final state as optimised and freeze them) we print a warning, no action taken.

src/quacc/calculators/espresso/calculators.py
tomdemeyere: The burden of running it locally will be on the users, so it will be a possibility for them, but yes I don't think there is anyway to enforce that directly in quacc

src/quacc/calculators/espresso/core.py
tomdemeyere: I thought about tweaking the ase espresso calc to do the job. If you have an idea for passing custom write/read functions in a simple that might be possible without being too ugly

src/quacc/calculators/espresso/calculators.py
Andrew-S-Rosen: Ah, makes sense! Thanks for clarifying!

src/quacc/calculators/espresso/core.py
Andrew-S-Rosen: Hmm... maybe there's a way. I'll give this a look and let you know if I think of a clean route.

tomdemeyere: When you have time, have a look at the latest commit it should be possible to work around using custom EspressoTemplate and Espresso calc functions. What's left is to code the read/write functions for each binary, if this works then I will be able to clean quite a lot a code. Is this what you were thinking about?

The only binary I can think of that will need additional tweaking is neb.x

Andrew-S-Rosen: @tomdemeyere: Yes, something like that was what I had in mind! I wasn't sure the particulars, but indeed I was thinking of something just like what you proposed.

buildbot-princeton:

tomdemeyere:
I somewhat successfully ran a cell-relaxation + phonon workflow (yay!) but this required specific fiddling:

Quantum Espresso re-uses files from previous runs, ph.x uses charge densities and wave functions as well as data files (which are written from the last runs). The way things work for now is that after every step, clean_calc is called and everything is deleted from scratch.

One way to workaround that is to use the 'copy_files' list to copy back files to scratch, 3 problems with that:

- Files are potentially big, maybe we don't want to copy them around.
- Their name will change depending on how you compiled QE.
- Their number will vary depending on how you parallelize your calculation.

I think the best solution here is to bypass clean_calc somehow. Only calling it at the end? Surely QE will not be the only software with this problem anyway?

Andrew-S-Rosen:
@tomdemeyere: Great news. Let's get to the bottom of the file-passing issue.

>  The way things work for now is that after every step, `calc_cleanup` is called and everything is deleted from scratch.

You raise some good points, and I'm very open to suggestions on how to handle file transfer better.

First, let me share information you're probably already aware of (just to make sure we're on the same page). Apologies if you know all this.

The files are never deleted per se. The `calc_cleanup` function moves files out of the temporary scratch (sub)directory (defined by `SETTINGS.SCRATCH_DIR`) back to the results directory (defined by `SETTINGS.RESULTS_DIR`) where they are stored permanently. The current working directory also changes from the scratch directory to this results (sub)directory.

With that said, if you want to pass a wavefunction from job 1 to job 2, one can use `copy_files=["WAVECAR"]` (idk what it's called in QE) in the kwarg to job 2, and you're good to go. Alternatively, if you are developing a pre-made flow, you could always do `copy_files=[Path(results1["dir_name"], "WAVECAR")]` to be a bit more explicit. The first example is shown [here](https://github.com/Quantum-Accelerators/quacc/blob/081c3376ee44eae6da973db37871e9cdf471160e/src/quacc/recipes/vasp/core.py#L165-L172) for VASP. The second example is shown [here](https://quantum-accelerators.github.io/quacc/user/advanced/file_transfers.html#transfers-from-a-known-file-location) in the docs.

Now, from your comment above, it sounds like you're aware of this copying mechanism. Let me discuss a few of your points.

> Files are potentially big, maybe we don't want to copy them around.

This is indeed true. However, in a managed workflow, there is no guarantee that job 2 is run immediately after job 1. It could be run days or weeks later --- or potentially even on another machine altogether. For this reason, we don't want to keep the scratch directory alive after each `@job` ends, nor do we want to rely on its existence.

> Their name will change depending on how you compiled QE. Their number will vary depending on how you parallelize your calculation.

If you are defining two separate `@job`s, then this is in principle up to the user to decide when they call `copy_files`. If you are making a single pre-made `@flow`, then indeed hard-coding the name of the file(s) will be a challenge --- one that can probably only be resolved with some additional setting.

So, with all the above info out of the way: could you perhaps share an example script (even if it's just a sketch) for this proposed workflow? That might help me help you the best way I can.

src/quacc/recipes/espresso/core.py
```diff
@@ -0,0 +1,211 @@
...
+        },
+        'qpts': [(0, 0, 0)]
+    }
+
+    copy_files = copy_files or \
+        ['pwscf.xml',
+         'pwscf.save/data-file-schema.xml',
+         'pwscf.save/charge-density.dat.gz',
+         'paw.txt'
+         ]
```
Andrew-S-Rosen, line 138: This will work locally, but once you start deploying calculations remotely with a workflow manager, there is no guarantee about where your current working directory will be. As such, copying from a relative file path won't work anymore.

Andrew-S-Rosen: It looks like what you're actually trying to do is inject workflow logic into a job.

Perhaps what makes more sense is to make a custom `@flow` that consists of two `@job`s: `relax_job` and then `ph_job`. This then gives you more control over things. For instance, you'll know where the previous job's files are: it'd be `relax_job_output["dir_name"]`, assuming `relax_job_output` is the output of `relax_job`.

Andrew-S-Rosen:
With the above comments in mind, what may make the most sense is to make a custom `@flow` that consists of two `@job`s: `relax_job` and then `ph_job`. This then gives you more control over things. For instance, you'll know where the previous job's files are: it'd be `relax_job_output["dir_name"]`, assuming `relax_job_output` is the output of `relax_job`.

[Here](https://quantum-accelerators.github.io/quacc/dev/recipes/flows.html) is a description of `@flow`s. What you're doing is very similar to what I've done in the [`mp_relax_flow`](https://github.com/Quantum-Accelerators/quacc/blob/081c3376ee44eae6da973db37871e9cdf471160e/src/quacc/recipes/vasp/mp.py#L128-L173).

tomdemeyere:
Thanks for your replies, I understand better now. I was indeed using a flow, I see two ways for the second job of the flow to read in previous data:

1 - It is possible to specify which dir to look for data in QE input files, avoiding the need to copy files by directly reading them in the results["dir_name"], the only problem I see is that the files are gzipped? Is this a reliable solution?

2 - Use copy files, but this would need to copy a whole directory structure, as the name and number of file is unpredictable (two exact same jobs will be different on different machines)

Andrew-S-Rosen:
> 1 - It is possible to specify which dir to look for data in QE input files, avoiding the need to copy files by directly reading them in the results["dir_name"], the only problem I see is that the files are gzipped? Is this a reliable solution?

I know it was crossed out, but I'll answer anyway. In short, this would be plausible to implement. I am not too concerned about the results directory not being "high performance" because usually that has to do with writing rather than reading, and even if it is somewhat slow, the operation is done once (as opposed to a QE calculation that is streaming an output file to disk, which is crucial to do on a performant file system).  It's also okay if the files are gzip-d because Python has a `gzip` utility that can unzip a file in memory. However, this points to the potential problem with this approach --- for large systems, you will be using a lot of memory, and that could potentially cause issues. But not a huge problem in the scheme of things.

> 2 - Use copy files, but this would need to copy a whole directory structure, as the name and number of file is unpredictable (two exact same jobs will be different on different machines)

I have been thinking that this might be a more natural approach in most cases, so this might be a good opportunity to evaluate that hypothesis. I was thinking we could implement a `prev_results_dir` kwarg (rather than a `copy_files` kwarg). This would be a directory, and all files would be copied rather than any specific file. I feel like this is probably what most people would want anyway. Sure, you'll copy things like old input files, but they'll get overwritten and it won't matter.

Option 2 is probably the mechanism I would suggest using. Feel free to give it a go, and when things are getting close to a merge-able state for you, I'll clean things up. If it ends up making a lot of sense, I'll probably use `prev_results_dir` kwargs in most of the other recipes.

As a side-note, it's a bit awkward to have to explicitly implement the `prev_results_dir` (or `copy_files`) kwarg in each and every recipe. If you can think of a good way around that without refactoring everything into classes, I'm open to suggestions.

tomdemeyere:
Something else that I would like to sort out: It seems that I can't change SETTINGS on the fly within a script, everything will still use the default value? I thought this should be as simple as importing SETTINGS and changing them?

```python 
from quacc import SETTINGS
from quacc.recipes.espresso.core import static_job, ph_job

SETTINGS.ESPRESSO_PP_PATH = '...'

@flow
def workflow(atoms):
    result1 = static_job(atoms, 'sssp_1.3.0_pbe_efficiency', input_data = input_data, kspacing = 0.2)
    result2 = ph_job()
    return result2

dispatch_id = ct.dispatch(workflow)(atoms)
result = ct.get_result(dispatch_id, wait=True)

```

This would not work because the SETTINGS imported in static_job are not the same instance?

Andrew-S-Rosen:
@tomdemeyere: Modifying global variables is tricky, but I think in this case it might be because of how you're using Covalent. The `@flow` (`@lattice`) is a pseudo-compiler that doesn't know anything about variables defined outside its scope. What happens if you move the call to `SETTINGS` inside the `@flow`?

tomdemeyere:
Moving everything inside the @flow does not change anything unfortunately. In the documentation you have a section 'Modifying the Global Settings in a Script' which says that it is possible to make changes on the fly, I think it would be nice to have it for all engines?

Andrew-S-Rosen:
@tomdemeyere: In principle, it _should_ work (but clearly not). I'll look into it and reply with more details here later this afternoon (PST).

buildbot-princeton:

Andrew-S-Rosen:
@tomdemeyere: 

> I did some runs on hpcs, the only problem I have to report is that python dictionaries (and lists it seems) that are passed as parameters to the static_job `@job` are considered as 'Electron Dict' by covalent and have a whole node attached to them. As a results they will have a slurm job running as well, which makes little sense.

There are a few subtleties to highlight when it comes to object passing in any workflow engine, Covalent included. If passing an iterable, like a `Dict` or `List`, as the input to a `Job` (i.e. `Electron`) in Covalent, a separate node will be shown in the GUI. However, this node will be resolved server-side and should not queue up via Slurm. This is true for all pre- and post-processing nodes. Covalent needs to (un)pickle them so they can be transported between server and execution agent.

There is one caveat to this, and it has to do with passing unresolved objects between jobs. Consider a workflow where you have relax job --> static job. If you pass `results["atoms"]` obtained from `relax_job` as the input to `static_job`, Covalent has to do two things: 1) Unpack/resolve the `"atoms"` key of `results` (which is not known until runtime); 2) Pass this to the second job. This will cause an additional node for the unpacking step, which I imagine is what you saw. This scenario is described in much more detail in https://github.com/AgnostiqHQ/covalent/discussions/1589. In short, you can use a `workflow_executor` kwarg to specify where these processing `Electron`s are carried out (e.g. server-side). But Covalent is also in the process of ensuring this process happens automatically and more seamlessly in the background of other running tasks. See the aforementioned discussion for more details.

This is a complexity for most workflow engines. For a more technical breakdown, you can refer to [a blog post](https://parsl-project.org/2023/11/30/passing-complex-data-structures.html) I wrote for Parsl on this topic as well.

src/quacc/settings.py
```diff
@@ -121,6 +127,42 @@ class QuaccSettings(BaseSettings):
...
+        description=("Path to the espresso q2r.x executable."),
+    )
+    ESPRESSO_MATDYN_CMD: Path = Field(
+        Path("matdyn.x"),
+        description=("Path to the espresso matdyn.x executable."),
+    )
+    ESPRESSO_DYNMAT_CMD: Path = Field(
+        Path("dynmat.x"),
+        description=("Path to the espresso dynmat.x executable."),
+    )
```
Andrew-S-Rosen, line 164: Message from before:

Are these executables typically found within a single directory? If so, perhaps it would be easier to have a single variable (e.g. ESPRESSO_EXE_DIR). That would help later since if there are a lot more executables one might want to add, they won't have to enumerate each one as a setting. However, if that's not the typical structure of QE's executables, then this comment isn't relevant. (As you can tell, I have not used QE before).

An alternative might be to define a single setting that is a dictionary of paths, e.g. 

```python
{"ph": Path("ph.x"), "neb": "neb.x", ...}
```

This has the benefit that it's only one setting, and it's easy to augment as needed.

src/quacc/settings.py
```diff
@@ -90,6 +88,14 @@ class QuaccSettings(BaseSettings):
...
         ),
     )
+    COMMON_TMPDIR: str | None = Field(
+        None,
+        description=(
+            "Whether to use a common temporary directory for all jobs."
+            "This is useful for parallel workflows where the same files are needed "
+            "by multiple jobs. Should be a string."
+        ),
+    )
```
Andrew-S-Rosen, line 98: I am a bit hesitant to add a generic setting like this for several reasons. Let's try not to rely on this if possible.

src/quacc/runners/prep.py
```diff
@@ -57,10 +59,16 @@ def calc_setup(
...
     # Create a tmpdir for the calculation within the scratch_dir
-    time_now = datetime.now(timezone.utc).strftime("%Y-%m-%d-%H-%M-%S-%f")
-    tmpdir = Path(
-        mkdtemp(prefix=f"quacc-tmp-{time_now}-", dir=SETTINGS.SCRATCH_DIR)
-    ).resolve()
+    # Just a test to see if we can avoid heavy io
+    if tmpdir:
+        prefix = f"quacc-tmp-common-{tmpdir}"
+        tmpdir = Path(SETTINGS.SCRATCH_DIR / prefix)
+        tmpdir.mkdir(parents=True, exist_ok=True)
```
Andrew-S-Rosen, line 66: I remain hesitant about changes to the path handling for the reasons previously described. You don't want files to live in scratch after a calculation has been completed because the scratch directory may often be on a filesystem that is purged, and there is no guarantee for when the follow-up job in the workflow will be executed. The scratch directory is not meant to persistent. I'd rather have file transfers than scenarios subject to data loss.

If the problem is that the copy operations are too expensive, then the user should define the `RESULTS_DIR` to be on the same filesystem as `SCRATCH_DIR`, since then it should be a fairly trivial operation.

src/quacc/runners/ase.py
```diff
@@ -36,7 +36,11 @@ class VibKwargs(TypedDict, total=False):
 
 
 def run_calc(
-    atoms: Atoms, geom_file: str | None = None, copy_files: list[str] | None = None
+    atoms: Atoms,
+    geom_file: str | None = None,
+    copy_files: list[str] | None = None,
+    tmpdir: str | None = None,
+    calc_clean: bool = True,
```
Andrew-S-Rosen, line 43: See comments below about the file I/O stuff. I'm not yet convinced that this is ideal.

src/quacc/recipes/espresso/neb.py
Andrew-S-Rosen, line 1: To make it easier to merge, let's save any functionality beyond core recipes to a follow-up PR. It is easier to review and approve small chunks at a time.

src/quacc/recipes/espresso/core.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@job
+def static_job(
+    atoms: Atoms,
+    preset: str | None = None,
+    copy_files: list[str] | None = None,
+    tmpdir: str | None = None,
+    parallel_info: dict[str] | None = None,
+    calc_clean: bool = True,
```
Andrew-S-Rosen, line 27: Here and in several places, docstrings will need to be added.

After our discussion, I'm not sure we'll need `tmpdir` or `calc_clean`. Let's circle back to those.

src/quacc/calculators/espresso/utils.py
```diff
@@ -0,0 +1,154 @@
...
+        'pseudopotentials': pseudopotentials,
+        'input_data': tmp_input_data}
+
+
+def namelist_to_string(parameters):
+    # This function aim to write basic input keywords to a file
+    # For additional cards (binary specifics) this is down in the child class
+    # Stolen from ase.io.espresso,
+    # maybe this could be made to a function "write_namelist"
+    # and imported here? See with the ASE gods.
```
Andrew-S-Rosen, line 141: I am an ASE maintainer. I can review and approve an MR for this.

src/quacc/calculators/espresso/utils.py
```diff
@@ -0,0 +1,154 @@
+import warnings
+
+from ase.io.espresso import Namelist
+
+
+# Stolen from ase.io.espresso and modified to add a parameter 'keys',
+#  Maybe there is a better way of doing this?
```
Andrew-S-Rosen, line 7: I'm not an expert on QE, but it would be nice to not rely on copying things so that upstream changes can be reflected here. I understand if that's not possible, however.

Andrew-S-Rosen (commented):
I have left a few more comments. Most of this should be pretty acceptable and self-explanatory. The only point of potential contention is about the tmpdir handling stuff. Let's chat about that over Zoom.

Since this PR is already quite large in scope, I would recommend limiting any future feature additions until tests are added and saving those features for a separate PR to make it easier to review and merge.

Chat with you soon.

tomdemeyere:
A large amount of things I kept was purely to simplify testing (the tmpdir, etc), I moved that to another local branch and cleaned up that one to simplify the whole process.

To come back on what you said ealier:

> @tomdemeyere:
> 
> > I did some runs on hpcs, the only problem I have to report is that python dictionaries (and lists it seems) that are passed as parameters to the static_job `@job` are considered as 'Electron Dict' by covalent and have a whole node attached to them. As a results they will have a slurm job running as well, which makes little sense.
> 
> There are a few subtleties to highlight when it comes to object passing in any workflow engine, Covalent included. If passing an iterable, like a `Dict` or `List`, as the input to a `Job` (i.e. `Electron`) in Covalent, a separate node will be shown in the GUI. However, this node will be resolved server-side and should not queue up via Slurm. This is true for all pre- and post-processing nodes. Covalent needs to (un)pickle them so they can be transported between server and execution agent. This should take virtually no time at all.
> 
> There is one caveat to this, and it has to do with passing unresolved objects between jobs. Consider a workflow where you have relax job --> static job. If you pass `results["atoms"]` obtained from `relax_job` as the input to `static_job`, Covalent has to do two things: 1) Unpack/resolve the `"atoms"` key of `results` (which is not known until runtime); 2) Pass this to the second job. This will cause an additional node for the unpacking step, which I imagine is what you saw. By default, the executor for this `Electron Dict` is inherited from the parent workflow (e.g. Slurm).
> 
> This second scenario is described in much more detail in [AgnostiqHQ/covalent#1589](https://github.com/AgnostiqHQ/covalent/discussions/1589), which I encourage you to check out. In short, you can use a `workflow_executor` kwarg to modify where these processing `Electron`s are carried out (server-side is ideal). But Covalent is also in the process of ensuring this process happens automatically and more seamlessly in the background of other running tasks. See the aforementioned discussion for more details.
> 
> This is a complexity for most workflow engines. For a more technical breakdown, you can refer to [a blog post](https://parsl-project.org/2023/11/30/passing-complex-data-structures.html) I wrote for Parsl on this topic as well.

My workflow is for now very simple (only one job):

```python
atoms = bulk('Cu', 'fcc')

input_data = {
    'occupations': 'smearing',
    'smearing': 'cold',
    'degauss': 0.01,
    'ecutwfc': 40,
    'mixing_beta': 0.7,
    'mixing_mode': 'plain'
}

parallel_info = {'srun': True,
                    '--hint=nomultithread': True,
                    '--distribution=block :block': True}

@flow(executor=executor)
def workflow(atoms):
    results = static_job(
        atoms,
        'sssp_1.3.0_pbe_efficiency',
        parallel_info=parallel_info,
        input_data=input_data,
        kspacing=0.2)
    return results
```

I probably do something wrong because the two dict are running as HPC jobs:

Screenshot 2023-12-04 at 21 39 30

I guess we can talk about it tomorrow (I post it here in case someone has the same problem one day and can find it here)

src/quacc/calculators/espresso/utils.py
tomdemeyere: I created a MR on ase master, also to add the 'keys' parameter the construct_namelist function, so that we avoid copying code here. This should not affect ASE at all.

Andrew-S-Rosen: Great, thanks! I will review that and get it merged. In the meantime, you can keep what you have here and when it's merged in ASE we can swap it out.

Andrew-S-Rosen:
@tomdemeyere: Huh, that's interesting. It seems that I'm mistaken and the `workflow_executor` stuff influences _post_-processing steps, which begs the question of how one can ensure pre-processing steps are executed server-side (or, in this case, not executed at all since they never have to be resolved). This isn't something that'd be in quacc but rather a Covalent issue/question. I have asked the question in the Covalent Slack channel and will report back. If it's a bug, I'll open an issue and tag you.

Andrew-S-Rosen:
@tomdemeyere: I fixed the merge conflicts and merged `main` into your branch. The tests are now triggered. As you can see, there are a things broken that will need to be addressed, e.g. the YAML file reader update did not necessarily work as you expected (and DeepSource has some warnings as well). These are easy to fix. We can discuss more on the call.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _kwargs_handler(self, binary, **kwargs):
+        keys = ALL_KEYS[binary]
+        kwargs["input_data"] = construct_namelist(kwargs.get("input_data"), keys=keys)
+        self.calc_defaults["input_data"] = construct_namelist(
+            self.calc_defaults["input_data"], keys=keys
+        )
+        # Would be nice to change the merge_dict function so that
+        # it is fully compatible with the Namelist class. I believe
+        # changing 'dict or {}' would do.
```
Andrew-S-Rosen, line 90: Could you clarify what you mean? Let's remove the comment once addressed.

src/quacc/utils/files.py
```diff
@@ -60,13 +62,51 @@ def copy_decompress(source_files: list[str | Path], destination: str | Path) ->
     -------
     None
     """
+
+    # Proposal to make the copy file more flexible
+    # i.e add glob support. (Dangerous maybe?)
+    # Just testing for now
```
Andrew-S-Rosen, line 68: Looks like this comment can be removed. For more flexible glob support, let's try that in a follow-up PR since it's independent of QE.

tests/local/recipes/espresso_recipes/Li.upf
Andrew-S-Rosen, line 1: Can you `gzip` the .upf` files to make the repo smaller? Note that you'll have to then gunzip them in your test suite.

src/quacc/settings.py
```diff
@@ -301,7 +335,18 @@ def resolve_and_make_paths(cls, v):
...
+        "ESPRESSO_PP_PATH",
+        "GAUSSIAN_CMD",
+        "GULP_CMD",
+        "GULP_LIB",
+        "ORCA_CMD",
+        "QCHEM_LOCAL_SCRATCH",
+        "NEWTONNET_MODEL_PATH",
+        "VASP_PRESET_DIR",
+        "VASP_PP_PATH",
+        "VASP_VDW",
```
Andrew-S-Rosen, line 349: I haven't included `ESPRESSO_BIN_PATH` here because it's a dict, but one can imagine doing something similar with that.

src/quacc/settings.py
```diff
@@ -118,6 +118,37 @@ class QuaccSettings(BaseSettings):
         ),
     )
 
+    # ---------------------------
+    # ESPRESSO Settings
+    # ---------------------------
+    ESPRESSO_CMD: Path = Field(
+        Path("pw.x"), description=("Path to the espresso executable.")
+    )
+    ESPRESSO_BIN_PATH: dict[str, Path] = Field(
```
Andrew-S-Rosen, line 127: Should we rename this `ESPRESSO_BIN_PATHS` (plural)?

tests/local/recipes/espresso_recipes/test_espresso.py
```diff
@@ -0,0 +1,98 @@
...
+    results = static_job(
+        atoms, input_data=input_data, pseudopotentials=pseudopotentials, kspacing=0.5
+    )
+
+    input_data = dict(construct_namelist(input_data))
+
+    assert np.allclose(results["atoms"].positions, atoms.positions, atol=1.0e-4)
+
+    assert np.allclose(results["atoms"].cell, atoms.cell, atol=1.0e-3)
+    assert (results["atoms"].symbols == atoms.symbols).all()
```
Andrew-S-Rosen, line 45: Note that there is a [`numpy.testing.assert_allclose` command](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html), but it doesn't make a big difference to me.

tests/local/recipes/espresso_recipes/test_espresso.py
Andrew-S-Rosen, line 1: Before getting this merged, we'll need to increase the code coverage a bit more such that everything passes. Right now, you have added some nice tests for the recipes. However, it would be good to test the calculator directly as well (e.g. in the tests/local/calculators` module) to catch a few other untested lines. I would not necessarily suggest testing all the IO functionality explicitly though if that's going to get moved to ASE though.

Given how much you have contributed already, I won't force you to add more unit tests in addition to the more "functional" tests for the recipes, so I can help out with that before the merge if necessary. It may take a bit of time for me to get to it though. 

Note: You can see the untested lines in the "details" of the codecov runner.

Andrew-S-Rosen (changes requested):
@tomdemeyere: Alright, I have done one final review! Thank you again for all of your very nice work on this and for being a pleasure to work with.

I have left mostly minor comments here. I consider this good to go once they are addressed! 

P.S. For the deepsource issue, I just ignored it permanently in the system since I don't think it's valid for us.

Andrew-S-Rosen:
@buildbot-princeton ok to test

tomdemeyere:
Here you go!

One small detail left: I still didn't change the quacc settings variables. I like the dictionary idea but at the same time I would like the possibility to change the path for each binary specifically in quacc.yaml

Andrew-S-Rosen:
> Here you go!
> 
> One small detail left: I still didn't change the quacc settings variables. I like the dictionary idea but at the same time I would like the possibility to change the path for each binary specifically in quacc.yaml

@tomdemeyere: Fantastic! Thank you for your very detailed revision. I really appreciate the time you put into this. There was one comment lost in the pile that I wanted to re-raise here. Namely, I find the name `ph_job` to be a bit confusing (since it relies on the name of an executable, which is not necessarily known by users new to Espresso). I assume this is a phonon job? If so, can we rename it `phonon_job`? Additionally, the `additional_fields` for that job is `{"name": "ph.x static"}`. Is this correct? I don't know that I'd call it a static calculation per se.

Andrew-S-Rosen:
@tomdemeyere: Alright, this looks great to me! I am going to go ahead and have this merge once tests pass. Congratulations on this very nice PR, and thank you again for all of your insight. Please feel free to make further modifications as-needed in follow-up PRs.

tomdemeyere:
Thanks a lot for taking care of all of that! Looking at the commits it seems that I forgot quite a lot... I have a few branches that I might use to open follow-up PRs:

- More espresso presets
- First complex ph.x workflow (grid technique), using different executors
- Espresso outdirs handling

In the meantime I also opened a merge request on the ASE master about our in house code ONETEP to prepare the integration with quacc.

Andrew-S-Rosen:
> Thanks a lot for taking care of all of that! Looking at the commits it seems that I forgot quite a lot..

Not a problem! And the remaining stuff was mostly minor --- was happy to clean it up since you put in a lot of time and care into it as-is.

> I have a few branches that I might use to open follow-up PRs:

Sounds good to me!

> In the meantime I also opened a merge request on the ASE master about our in house code ONETEP to prepare the integration with quacc.

Awesome!! I had been wanting to add ONETEP recipes for a while. I have it compiled but never ran any test calculations myself. I'll be sure to review the ASE MR if Ask or someone else doesn't get to it in the next day or two.

PR #1299, opened 2023-12-04
Use `deepcopy` in `merge_dicts`

### LINKED ISSUE #1298: Deepcopy should be used in `merge_dicts`

- **quacc version**: 0.4.2
- **Python version**: 3.9+

### What is the issue?

As initially demonstrated in #1239, the `quacc.utils.dicts.merge_dicts` function uses `.copy()` when in reality, `deepcopy()` would be more suitable to ensure nested dictionaries are not mutated.

https://github.com/Quantum-Accelerators/quacc/blob/44fdc720a8af621503e851934a0c909991000c47/src/quacc/utils/dicts.py#L37-L39

The same holds true for the `merge_several_dicts` function.

The challenge, however, is that not all items in the merged dictionaries are (un)pickleable, causing `deepcopy()` to fail. The need for `deepcopy()` hasn't emerged because we never continue to use the input `dict`, but we should do it the right way.

Described above.

SarojArchana:
## Required Checklist

- [x ] I have read the [Contributing Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉
- [x ] My PR is on a custom branch and is _not_ named `main`.
- [ ] I have added relevant unit tests. Note: Your PR will likely not be merged without proper tests.

## Summary of Changes
I have Implement a Custom Copy Function to solve this issue. The Custom Copy Function performs a shallow copy of the top-level dictionary and recursively copies nested dictionaries using a try-except block to handle non-pickleable objects gracefully. This way, we can achieve a balance between deep copying and handling exceptions.

Closes #1298.


buildbot-princeton:

Andrew-S-Rosen:
@SarojArchana: Thanks! This seems like it might be a bit over-engineered for the task.

This is what I get from ChatGPT which seems reasonable to me at a quick glance (was yours ChatGPT too? 🙂). Could you review this, adjust it as needed, and see if it makes sense to use instead? Of course, the variable names may need to be updated to match what's already been used in quacc.

```python
import copy
import pickle

def custom_deepcopy(obj):
    try:
        return copy.deepcopy(obj)
    except (TypeError, pickle.PicklingError):
        if isinstance(obj, dict):
            return {key: custom_deepcopy(value) for key, value in obj.items()}
        elif isinstance(obj, (list, tuple)):
            return type(obj)(custom_deepcopy(item) for item in obj)
        else:
            # For other types, use the default copy.copy
            return copy.copy(obj)

def merge_dicts(dict1, dict2):
    result = custom_deepcopy(dict1)

    for key, value in dict2.items():
        if key in result and isinstance(result[key], dict) and isinstance(value, dict):
            result[key] = merge_dicts(result[key], value)
        else:
            result[key] = value

    if remove_nones:
        merged = remove_dict_nones(merged)

    return result
```

SarojArchana:
Dear Dr. @Andrew-S-Rosen ,
Thanks for the feedback.
The code provided by you looks good. I have tested the working of the above code by generating multiple test cases, and it works fine. Therefore, I  have made all the changes you suggested.
Please review this PR now and let me know If anything else needs to be added.

Andrew-S-Rosen:
@buildbot-princeton test this please.

Andrew-S-Rosen:
@SarojArchana: Thank you for your contribution. This looks good, but before I can merge this, you are going to need to add tests. As you can see below, the coverage for your changes is 53.84%, which means some lines are not covered (click the Details for which lines). Please ensure that all added lines of code are covered by a dedicated unit test, to be added to tests/local/utils/test_dicts.py.

SarojArchana:
Dear Dr. @Andrew-S-Rosen ,
I have added the extra test cases. Despite trying multiple test case in am not able to cover line 20 and line 58.
will it be possible for you to review it once and check if it is due to my test case design and due to some error in codecov/patch code.

Andrew-S-Rosen:
@SarojArchana: Your lines aren't being tested because the deepcopy works as expected. If you put something unpickleable, like a module (maybe?), you should be able to trigger the exception. You should keep the tests you already added though, which are nice.

Andrew-S-Rosen:
@SarojArchana: Looks like the tests are failing, however.

SarojArchana:
Are you taking about "tests / tests-core (3.9, ubuntu-latest) (pull_request) Failing after 3m"

PR #1306, opened 2023-12-05
New mechanism for copying files between jobs

tomdemeyere:
## Required Checklist

- [x ] I have read the [Contributing Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉
- [ ] I have added relevant unit tests. Note: Your PR will likely not be merged without proper tests.

## Summary of Changes

Changes about the copy_file mechanism that we discussed about, I use a recursive function here but we don't have to keep it that way.

I am not sure how to add a test for this as it will involve creating files/dir ?

### DIFF
--- a/src/quacc/utils/files.py
+++ b/src/quacc/utils/files.py
@@ -45,7 +45,7 @@ def check_logfile(logfile: str, check_str: str) -> bool:
     return False
 
 
-def copy_decompress(source: list[str | Path], destination: str | Path) -> None:
+def copy_decompress(source: str | Path, destination: str | Path) -> None:
     """
     Copy and decompress files from source to destination.
 
--- a/src/quacc/utils/files.py
+++ b/src/quacc/utils/files.py
@@ -45,14 +45,14 @@ def check_logfile(logfile: str, check_str: str) -> bool:
     return False
 
 
-def copy_decompress(source_files: list[str | Path], destination: str | Path) -> None:
+def copy_decompress(source: list[str | Path], destination: str | Path) -> None:
     """
     Copy and decompress files from source to destination.
 
     Parameters
     ----------
-    source_files
-        List of files to copy and decompress.
+    source
+        Directory to walk and copy files from.
     destination
         Destination directory.
 
@@ -60,13 +60,24 @@ def copy_decompress(source_files: list[str | Path], destination: str | Path) -> 
     -------
     None
     """
-    for f in source_files:
-        z_path = Path(zpath(f))
-        if z_path.exists():
-            copy(z_path, Path(destination, z_path.name))
-            decompress_file(Path(destination, z_path.name))
-        else:
-            warnings.warn(f"Cannot find file: {z_path}", UserWarning)
+    src, dst = Path(source), Path(destination)
+
+    if src.is_dir():
+        for f in src.iterdir():
+            # Symlink have to be explicitly passed
+            # as they are considered as 'dir' and 'file'
+            # otherwise and can seriously mess up the
+            # copying process
+            if f.is_symlink():
+                pass
+            elif f.is_file():
+                copy(src / f, dst / f.name)
+                decompress_file(dst / f.name)
+            elif f.is_dir:
+                (dst / f.name).mkdir(exist_ok=True)
+                copy_decompress(src / f, dst / f.name)
+    else:
+        warnings.warn(f"Cannot find dir: {src}", UserWarning)
 
 
 def make_unique_dir(base_path: str | None = None) -> Path:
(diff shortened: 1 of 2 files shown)


buildbot-princeton:

Andrew-S-Rosen:
@tomdemeyere: Thanks! A few comments.

1. The current test suite is failing because of the changed behavior. It would be nice to still support copying specific files _in addition_ to just copying a whole directory of contents, especially if the user or code developer knows they only want a subset of files. To permit this, maybe it makes sense to keep the first positional arg as `list[str | Path]` and if any entry in the list `.is_dir()` then we treat it as such. Open to other ideas, but we'll want to still support the existing behavior as an option.

2. To add tests that involve files or directories created solely for the tests, you can mimic what was done in `tests/local/runners/test_ase.py`. Note that there is a function, `prep_files()` that does some initial file staging operations. Then there is a function `teardown_function()` that resets everything. The methods `setup_function()` and `teardown_function()` are special pytest functions that do not need to be explicitly invoked --- they will automatically run at the start and end of a test function, respectively.

Sidenote: There are likely some potentially useful utilities in `monty.shutil`, which we are already importing, like [`copy_r`](0. You might benefit from using [`copy_r`](https://github.com/materialsvirtuallab/monty/blob/5d37679e1f0cd58a8d165d76f3a095cb71266831/monty/shutil.py#L14). However, I don't know that it plays nicely with symlinks. I might open a PR there afterwards if it makes sense.

tomdemeyere:
> @tomdemeyere: Thanks! A few comments.
> 
> 1. The current test suite is failing because of the changed behavior. It would be nice to still support copying specific files _in addition_ to just copying a whole directory of contents, especially if the user or code developer knows they only want a subset of files. To permit this, maybe it makes sense to keep the first positional arg as `list[str | Path]` and if any entry in the list `.is_dir()` then we treat it as such. Open to other ideas, but we'll want to still support the existing behavior (copying specific files) as an option in some shape or form.
> 2. To add tests that involve files or directories created solely for the tests, you can mimic what was done in `tests/local/runners/test_ase.py`. Note that there is a function, `prep_files()`, that does some initial file staging operations. Then there is a function `teardown_function()` that resets everything. The methods `setup_function()` and `teardown_function()` are [special pytest functions](https://docs.pytest.org/en/6.2.x/xunit_setup.html#method-and-function-level-setup-teardown) that do not need to be explicitly invoked --- they will automatically run at the start and end of a test function, respectively.
> 
> Sidenote: There are likely some potentially useful utilities in `monty.shutil`, which we are already importing, like [`copy_r`](https://github.com/materialsvirtuallab/monty/blob/5d37679e1f0cd58a8d165d76f3a095cb71266831/monty/shutil.py#L14). However, I don't know that it plays nicely with symlinks. I might open a PR there afterwards if it makes sense to adopt it later in place of some stuff here.

The problem with how things work now is that there is no way to deal with nested directories because we never know what is the 'base' directory. The function in this PR is made to typically be able to do:

```python
my_job(copy_files = old_job['dir_name'])
```
which solves the problem by unpacking everything and respecting the tree structure that way. If we allow the user to specify specific files with the way things are now there is no way to respect the tree structure (I think for now everything gets copied down to the parent directory)

Andrew-S-Rosen:
@tomdemeyere: I think we're on the same page actually, but it's possible there's just a slight miscommunication.

I agree that we should add functionality to (recursively) copy files from a source directory for the reasons you brought up. No hesitation about that.

I simply meant that if a user _does_ want to copy a specific file from `/path/to/my/staging/area/WAVECAR`, they should be able to do so. In other words, it'd be nice to have the existing capabilities in addition to the one you're proposing. There are several reasons for this. The first is that if a user wants to call `relax_job` and knows that a given file (e.g. WAVECAR) is in some arbitrary directory, they should be able to copy that directly if they so choose without worrying about isolating it in its own directory and then copying that. Only allowing `copy_decompress` to (recursively) copy a directory is also a breaking change that will require modifications to other recipes/tests to ensure that they continue function as expected (hence the failing test in the GitHub Actions runners).

Summary: Keep what you have. It is good. But have it structured like:

```python
def copy_decompress(sources: str | Path | list[str | Path], destination: str | Path):
    if isinstance(sources, (str, Path)):
        sources = [sources]
    for src in sources:
        if src.is_dir():
            ...
        elif src.is_file():
           # use the existing logic from main, i.e., copy the specific file
        else:
           warnings.warn()
```

That should still let you do:
```python
# example:
my_job(copy_files = old_job['dir_name'])
```

while supporting the pre-existing behavior.

Does this make sense, or am I misunderstanding an aspect?

Andrew-S-Rosen:
Closing this as duplicate of #1238.

PR #1313, opened 2023-12-07
`mv *` files from `SCRATCH_DIR` to `RESULTS_DIR` instead of `cp -r`, and make `SCRATCH_DIR` optional

Andrew-S-Rosen:
## Required Checklist

- [X] I have read the [Contributing Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉

## Summary of Changes

The default `SCRATCH_DIR` was changed to `None`, such that everything is done in the same directory as `RESULTS_DIR` unless otherwise specified. This should make the "first calculation" easier on the new user, as everything will happen "before their eyes."

Additionally, an `mv *` style of operation is done instead of `cp -r` when transferring files from `SCRATCH_DIR` to `RESULTS_DIR`, assuming the former is specified. This should speed up the process substantially.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.3]
+
+### Changed
+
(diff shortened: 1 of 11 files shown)


Andrew-S-Rosen:
@tomdemeyere: As described in the above summary, I modified the behavior of the I/O to address a few of your concerns. Please don't hesitate to let me know if you recommend any modifications.

PR #1335, opened 2023-12-11
Espresso: managing various outdirs

tomdemeyere:
## Required Checklist

## Summary of Changes

Edit: No need to be that complicated, this PR just makes sure that chosen outdirs exist if inside the calc_dir.

If the user decides to ask for outdirs that are outside the calc_dir via absolute paths, then there is nothing we should do (?). It is the user responsibility to deal with that data (this PR add a dictionary to keep track of that data but does nothing with it).


buildbot-princeton:

src/quacc/calculators/espresso/espresso.py
```diff
@@ -101,10 +110,55 @@ def read_results(self, directory: Path | str) -> dict[str, Any]:
...
+        Parameters
+        ----------
+        **kwargs
+            User-supplied kwargs
+
+        Returns
+        -------
+        kwargs
+            The merged kwargs
+        """
```
Andrew-S-Rosen, line 137: Should be `parameters` for the name of the parameters and `dict[str, Any]` for the name of the returns.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -180,7 +234,15 @@ def __init__(
             binary=str(bin_path), parallel_info=parallel_info, pseudo_path=pseudo_path
         )
 
-        super().__init__(profile=profile, parallel_info=parallel_info, **kwargs)
+        if kwargs.pop("directory", None):
+            warn(
+                "It is highly discouraged to use the 'directory' parameter when using quacc."
+            )
```
Andrew-S-Rosen, line 240: Let's change this to a raised error since it would do absolutely gnarly things. Can simply say it's not supported.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -101,10 +110,55 @@ def read_results(self, directory: Path | str) -> dict[str, Any]:
...
+        input_data = parameters.get("input_data", {})
+
+        cwd = Path.cwd()
+
+        for section in input_data:
+            for d_key in self.outdirs.keys():
+                if d_key in input_data[section]:
+                    path = Path(input_data[section][d_key])
+                    if path.is_absolute():
+                        path = path.resolve()
```
Andrew-S-Rosen, line 148: Can you replace with `.expanduser().resolve()`?

Andrew-S-Rosen (commented):
Thanks for adding these sanity checks! I left a few small comments, and we'll also need some more tests to evaluate the behavior of the new functionality, as you know.

> If the user decides to ask for outdirs that are outside the calc_dir via absolute paths, then there is nothing we should do (?). It is the user responsibility to deal with that data (this PR add a dictionary to keep track of that data but does nothing with it).

Should we simply forbid setting of those `outdirs` then?

tomdemeyere:
> Thanks for adding these sanity checks! I left a few small comments, and we'll also need some more tests to evaluate the behavior of the new functionality, as you know.
> 
> > If the user decides to ask for outdirs that are outside the calc_dir via absolute paths, then there is nothing we should do (?). It is the user responsibility to deal with that data (this PR add a dictionary to keep track of that data but does nothing with it).
> 
> Should we simply forbid setting of those `outdirs` then? Perhaps that's a simpler solution altogether.

I don't think we should forbid it as it will never make anything crash if used properly (with this patch), and it would remove some flexibility. The only cases where it might cause an issue are:

- If the user specify a path outside quacc that does not exist.
- If the user is inconsistent in his choice of outdir in the same flow.

But I would argue that these are cases where the calculation deserves to crash anyway. Also, this gives an alternative to copying files around, for people with a very small scratch space this might be useful. If we keep it, maybe I should add paragraph in the doc to explain all that.

Andrew-S-Rosen:
> I don't think we should forbid it as it will never make anything crash if used properly (with this patch), and it would remove some flexibility. The only cases where it might cause an issue are:

I see. That makes sense then. Thanks for clarifying.

>  Also, this gives an alternative to copying files around, for people with a very small scratch space this might be useful. If we keep it, maybe I should add paragraph in the doc to explain all that.

In this scenario, the user could simply not set a `SCRATCH_DIR` setting, and then there will be no file shuttling. If set, then files are moved (but not copied).

Certainly, if desired, additional details can be proved in a docstring, which is rendered in the API documentation.

Andrew-S-Rosen:
Thanks for the really great test! To get this one merged, there's just a few more lines that'd be great to cover. To make the process quicker, no need to do a "real" calculation to address the remaining uncovered lines. You can simply add the following to `tests/core/calculators/espresso/test_espresso.py`:

```python
from quacc.calculators.espresso.espresso import EspressoTemplate

def test_espresso_write_input(monkeypatch, tmp_dir):
    monkeypatch.chdir(tmp_dir)
    params = {...}
    new_params = EspressoTemplate._outdir_handler(params)
    # do assertions on `new_params`
```

src/quacc/calculators/espresso/espresso.py
```diff
@@ -101,11 +109,56 @@ def read_results(self, directory: Path | str) -> dict[str, Any]:
...
+
+        Returns
+        -------
+        dict[str, Any]
+            The merged kwargs
+        """
+
+        input_data = parameters.get("input_data", {})
+
+        cwd = Path.cwd()
```
Andrew-S-Rosen, line 140: I fear this may cause some confusion, and I generally recommend not relying on the `cwd` in quacc. Everything should be absolute whenever possible.

Let's say you want to run a two subsequent jobs. In quacc, what will happen is the following:

1. When `calc_setup()` is called, It'll `cd` into a custom directory to run Job 1 (the `tmpdir`).
2. The job will finish, and it will `cd` back into the directory where the Job 1 results are stored (`job_results_dir`). 
3. Job 2 will start, and it'll `cd` into a custom directory to run Job 2 _when `calc_setup` is called_.

But when the `Espresso` calculator is instantiated, you're still in the directory from mentioned in Step 2, which the user doesn't know about. There are also other, much more complicated concerns with using relative paths or `cwd` when running parallelized calculations with many workflow engines (notably parsl) that I won't bore you with.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -101,11 +109,56 @@ def read_results(self, directory: Path | str) -> dict[str, Any]:
...
         return results
 
+    def _outdir_handler(self, parameters: dict[str, Any]) -> dict[str, Any]:
+        """
+        Function that handles the various outdir of espresso binaries.
+        If the user-supplied path are absolute, they are resolved and checked
+        against the scratch_dir and results_dir. If they are relative, they are
+        resolved against the scratch_dir and results_dir. If they are not in
+        either of those directories, they are ignored. This case is discouraged,
+        are encouraged to use quacc's scratch_dir and results_dir.
```
Andrew-S-Rosen, line 125: I don't see any mention of the scratch_dir or results_dir in this function. Not sure how the docstring is related there.

src/quacc/calculators/espresso/espresso.py
Andrew-S-Rosen, line 140: I try to avoid relying on `cwd` whenever possible because it can be quite unexpected at times. Any reason why you can't use `Path(directory)` that was already present in `write_input`?

tomdemeyere: No reason to use directory in this case (it will always be equal to '.' as we force it to be), a simple cwd = Path().expanduser().resolve() works, but isn't what .cwd() is doing under the hood anyway?

I suspect you might like the last commit better?

Andrew-S-Rosen: @tomdemeyere: I had suggested `directory` just in case we eventually/somehow supported the `directory` kwarg, so that the change would be reflected elsewhere. But you're right, `directory` is forced to be `"."`, which is `cwd`, so it does not really matter in the end.

`Path().expanduser().resolve()` will do the same as `.cwd()`, as you noted. I suppose it does not matter all that much in the end since we probably won't ever support the `directory` kwarg.

src/quacc/calculators/espresso/espresso.py
Andrew-S-Rosen: I tried updating this.

Andrew-S-Rosen:
@tomdemeyere: Is this ready to merge? I made some very minor changes to reflect the open comments. Thanks for this nice PR!

@buildbot-princeton: test this please.

tomdemeyere:
Ready to merge

PR #1354, opened 2023-12-13
Deprecate old Q-Chem recipes and use `terminate_on_nonzero_returncode`

Andrew-S-Rosen:
@samblau


Andrew-S-Rosen:
Starting over because the rename is dsajkldsjladsakl

PR #1379, opened 2023-12-16
Espresso: improved preset handler + more presets

tomdemeyere:
### Checklist

- [ ] I have added relevant unit tests.

### Notes

- Your PR will likely not be merged without proper and thorough tests, even if it's for a "simple" fix.

## Summary of Changes

More presets for the espresso calculator, I am not done yet but I open this PR to discuss a little bit more about presets. For now I created new presets for specific cases (metals, semiconductors) where I copied the pseudopotentials part from the previous two presets.

Should it be possible for the user to specify more than one preset i.e:

``` python

(..., preset = 'my-param-preset, pseudopotentials = 'my-pp-preset', ...)
(..., presets = ['my-param-preset, 'my-pp-preset'], ...)
(..., preset = 'my-big-preset-containing-both--pp--and-param', ...)

```


buildbot-princeton:

src/quacc/calculators/espresso/presets/bulk_metal_precision.yaml
Andrew-S-Rosen, line 1: Rather than duplicating information between presets, like the pseudopotentials, you should have a separate YAML for the pseudopotentials and then have that be inherited by all child YAML files that use it. See [this example](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/calculators/vasp/presets/BulkSet.yaml). Any parameter with `parent_` in the name will refer to a parent YAML file that will be inherited as the base set. This is mentioned in the `load_yaml_calc` [docstring](https://github.com/Quantum-Accelerators/quacc/blob/9f97088f3a8817750dfb68ef62340d7979ce319c/src/quacc/utils/files.py#L129-L134).

Andrew-S-Rosen (commented):
@tomdemeyere: Thanks for this. I left a comment here explaining the situation. In quacc, it is possible to have one YAML file inherit from another one, so there is never a need to duplicate parameters (see comment for details).

Something like the following would do what you need without duplication:

```yaml
input_data:
  system:
    occupations: smearing
    smearing: marzari-vanderbilt
    degauss: 0.005
  electrons:
    conv_thr: 1.0e-12
    mixing_mode: TF
    mixing_beta: 0.7

kspacing: 0.10

parent_pseudos: ssp_1.3.0_pbe_efficiency.yaml
```

Additionally, if you want the user to be able to mix-and-match pseudopotential sets and input parameter-based YAML files, you can add a custom `pseudopotentials` kwarg to the calculator. This is how the `Vasp` calculator works. The user can load in a preset of settings, which may or may not include pseudopotentials, and they can also override those pseudopotentials by setting the `setups` parameter in the `Vasp` calculator (`setups` is ASE jargon for pseudopotentials in VASP).

src/quacc/calculators/espresso/presets/bulk_metal_efficiency.yaml
```diff
@@ -0,0 +1,635 @@
...
+    smearing: marzari-vanderbilt
+    degauss: 0.01
+  electrons:
+    conv_thr: 1.0e-10
+    mixing_mode: TF
+    mixing_beta: 0.7
+
+kspacing: 0.25
+
+pseudopotentials:
```
Andrew-S-Rosen, line 17: Can probably be `parent_pseudopotentials: sssp_1.3.0_pbe_efficiency.yaml` if I understand correctly.

src/quacc/calculators/espresso/presets/semiconductors_precision.yaml
```diff
@@ -0,0 +1,635 @@
...
+    smearing: gaussian
+    degauss: 0.005
+  electrons:
+    conv_thr: 1.0e-12
+    mixing_mode: local-TF
+    mixing_beta: 0.6
+
+kspacing: 0.20
+
+pseudopotentials:
```
Andrew-S-Rosen, line 17: Can probably be `parent_pseudopotentials: sssp_1.3.0_pbe_precision.yaml`.

src/quacc/calculators/espresso/presets/sssp_1.3.0_pbe_precision.yaml
```diff
@@ -0,0 +1,623 @@
+# SSSP PBE Efficiency v1.3.0
+# https://doi.org/10.1038/s41524-018-0127-2
+# https://www.materialscloud.org/discover/sssp/table/efficiency
```
Andrew-S-Rosen, line 3: Docstring should presumably be updated to the "precision" version.

tomdemeyere:
Thanks for the info, indeed the 'parent' mechanism is all I was looking for, I used it in pretty much every new presets. As you will see I changed the function the parse_pp_and_cutoff in espresso.utils to reflect the new changes in presets.

Something you might not like is the new ability to modify the Atoms object directly by a preset, If not I can probably find another way to do what I want.

I added a test to make sure everything is working correctly and that the priority between calc_default, user defined kwargs and presets kwargs is respected.

Andrew-S-Rosen:
@tomdemeyere: Great! Glad that feature could help streamline things.

Would you be able to walk me through the logic for having the `pbc: [True, True, True]` in the preset? Maybe once I understand the context a bit more, I can help make a suggestion that is a bit less roundabout. It would indeed be nice to mess with the `Atoms` object if possible, but more-so I am looking to understand the use case you have in mind.

tomdemeyere:
Specifying pbc: [True, True, **False**] would make things work better with ase.io.write_espresso_in for example. If you specify the kspacing (which is a single float), the k-grid generated with the specified pbc's would have no k-point along the third dimension.

This is completely avoidable, for example by manually calling the function kspacing_to_grid from ase as a workaround. I just thought that changing Atoms inside presets might be an interesting idea that you might want to hear about, but after second thoughts changing it might raise a few other problems

Andrew-S-Rosen:
Ah thanks for clarifying! Yeah, I think I'd prefer this to be taken care of outside the presets. After all, the user may set no preset and also this behavior is not immediately obvious to a new user. In general, I like to abide by the principle of least surprise. Let's keep setups to just input parameters to the calculator.

tomdemeyere:
@Andrew-S-Rosen Presets still contains atoms_info but it is not used to modify the Atoms object, instead, it creates an atoms copy and calls spacing_to_grid manually.

Creating an entire copy might be overkill for now, but will be useful in the future if other checks are performed

This should be pretty good to go as it is, I suspect there are some details you might want to check

Andrew-S-Rosen:
> @Andrew-S-Rosen Presets still contains atoms_info but it is not used to modify the Atoms object, instead, it creates an atoms copy and calls spacing_to_grid manually.
> 
> Creating an entire copy might be overkill for now, but will be useful in the future if other checks are performed
> 
> This should be pretty good to go as it is, I suspect there are some details you might want to check
> 
> EDIT: re-reading your last comment I realise you probably want no "atoms_info" in presets at all, I will find a way to delete them

Thanks for the update!

One option that you might prefer is to create a separate module called `slabs.py` where the default preset is set to one of your slab presets, and in that recipe you could also set `atoms.pbc` as needed. See the [analogous module for VASP](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/vasp/slabs.py). This is entirely optional though. I just wanted to let you know of an alternate way to think about this.

Regarding the atoms info, a reason that I'm not a huge fan of the `atoms_info` key in the YAML file is that it is effectively a hidden feature. I try to ensure that the YAML parameters have 1:1 correspondence with calculator parameters, so it's obvious what can and can't be set. Here, the YAML is supporting some "hidden convenience magic". Also, one might think it would be `atoms.info` that is being set when in reality, that's not the case.

I'm not sure off-hand if there are other parameters behaving similarly, e.g. `kspacing`. If there are any parameters set in the YAML that are not direct arguments for the calculator, please take a moment to add them as defined, optional keyword arguments to the `Espresso` class that way they are documented and more generic.

Otherwise, this looks good and I'm happy to merge once that minor detail is sorted out. Thank you!

Andrew-S-Rosen:
@tomdemeyere: For a simple fix, we can consider adding and documenting a kwarg to the Espresso calculator that is `slab: bool = False`. If set to `True`, then the PBCs can be set appropriately. Then in the YAML, you could have `slab: true`.

tomdemeyere:
> @tomdemeyere: For a simple fix, we can consider adding and documenting a kwarg to the Espresso calculator that is `slab: bool = False`. If set to `True`, then the PBCs can be set appropriately. Then in the YAML, you could have `slab: true`.

With these solutions, a user who uses the ESM presets without setting this additional parameter or without using the specific workflow/job would be very likely to create a non-functional calculation (ESM in QE refuses any k-points that is not in the x y 0 plane) which is not ideal for a preset.

So far I am trying to find a solution for a preset that is:

- Self-contained (no need for additional parameters for a calculation to be functional)
- Does not use any kind of hidden features (although we could argue that input_atoms is a parameter of the Espresso calc)

Additionally, I think it is very important that users can feel they have a global grasp of what is going on when using quacc (and by that, I mean which parameters are doing what and when), and adding such a specific "slab" might be confusing in this regard. When I was trying a different workflow engine, I quickly got overwhelmed by how many parameters there were, and understanding the overlap between them was complicated and time-consuming.

For now, I am thinking about the possibility of turning kspacing into a 3-element list, that would allow the preset to specify the spacing in each direction. It's not ideal because it does not respect the original definition of the calculator though...

Andrew-S-Rosen:
It sounds like this might be resolved by having the dedicated slab workflows. But I will trust your best judgment in the end.

I agree that minimizing the number of non-standard parameters is ideal for the sake of the user.

tomdemeyere:
No satisfying solution here in my opinion, slab presets will not have any k-point information. It will be up to the user to specify them.

It should be good to go if you are happy with it

Andrew-S-Rosen:
@tomdemeyere: Thanks for this nice PR. I am happy to merge as-is. We can always revisit things later if desired.

PR #1403, opened 2023-12-23
Espresso: first complex flow

tomdemeyere:
### Checklist

### Notes

## Summary of Changes

First complex workflow with ph.x. It aims to make phonon calculations embarrassingly parallel by allowing parallelization at the representation level, requirements:

- [x] Takes two arguments, a static job (pw.x) and a ph.x job (ph.x)
- [x] quacc must be able to run 'test_runs' with both of these binaries
- [ ] As a result, these two jobs will have variable computational loads at times. The flow should be able to run these jobs using different executors or finding a workaround to mimic that.
- [ ] We should be able to copy a list of directories (I have made a workaround for now but it surely can be better)

Extra to-do:

- [ ] Allowing an extra argument "blocks" which is an integer and allows the user to regroup multiple representations into one calculation, this is useful in case of limited scratch space. if "blocks" == 0 this is the extreme case where we fall back on q-point parallelization only.
- [ ] Some calculations will be able to be done very quickly because of symmetry, either find a way to trick QE into doing them at the end during the "recover" run or be able to specify another executor for them to do it either on the login node or a scavenger/serial HPC queue.
- [ ] Write tests.
- [ ] Write docs.


buildbot-princeton:

PR #1404, opened 2023-12-24
Espresso: Dynamic phonon workflow

tomdemeyere:
### Checklist

### Notes

## Summary of Changes

First complex workflow with ph.x. It aims to make phonon calculations embarrassingly parallel by allowing parallelization at the representation level, requirements:

- [x] Takes two arguments, a static job (pw.x) and a ph.x job (ph.x)
- [x] quacc must be able to run 'test_runs' with both of these binaries
- [x] As a result, these two jobs will have variable computational loads at times. The flow should be able to run these jobs using different executors or finding a workaround to mimic that.
- [x] We should be able to copy a list of directories (I have made a workaround for now but it surely can be better)

Extra to-do:

- [x] Allowing an extra argument "blocks" which is an integer and allows the user to regroup multiple representations into one calculation, this is useful in case of limited scratch space. if "blocks" == 0 this is the extreme case where we fall back on q-point parallelization only.
- [x] Write tests.
- [ ] Write docs.


buildbot-princeton:

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Function to carry out a basic ph.x calculation.
+
+    Parameters
+    ----------
+    preset
+        The name of a YAML file containing a list of parameters to use as
+        a "preset" for the calculator. quacc will automatically look in the
+        `ESPRESSO_PRESET_DIR` (default: quacc/calculators/espresso/presets).
+
```
Andrew-S-Rosen, line 36: As a head's up, there can't be line breaks between the arguments otherwise it wouldn't render in the `mkdocs` documentation. A bit annoying, but it is what it is.

Also, please add `test_run` and any other kwargs to the docstring when you're done. Thanks!

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+        calc_defaults=calc_defaults,
+        calc_swaps=calc_kwargs,
+        parallel_info=parallel_info,
+        additional_fields={"name": "ph.x static"},
+        copy_files=copy_files,
+    )
+
+
+@flow
+def grid_phonon(pw_job: Job = static_job, phonon_job: Job = phonon_job):
```
Andrew-S-Rosen, line 97: 1. As a matter of convention, all flows in quacc are named with `_flow` in the name. For instance, `grid_phonon_flow`. When you're done, also make sure to update the [list of recipes](https://quantum-accelerators.github.io/quacc/user/recipes/recipes_list.html#quantum-espresso) in the documentation and a docstring describing the flow.

2. A subtle comment that I'd be happy to get your thoughts on. I also originally thought that having the `Job`s be the kwargs themselves in `Flow`s would be ideal since it's natural. However, then in test runs, I ran into confusing situations because even though they _look_ like functions, when a workflow engine is specified, they become instantiated classes that are mutable. This can lead to unexpected behavior when modifying the class attributes. 

To resolve this, I ended up doing what's shown [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/emt/slabs.py): the kwargs would be `custom_pw_job: Job | None = None` and `custom_phonon_job: Job | None = None`, and then in the function itself, you'd do a check for `pw_job = static_job if custom_pw_job is None else custom_phonon_job`. This protects against mutability of keyword arguments. However, it's sadly at the expense of a slightly more cumbersome name for the keyword argument. I would recommend adopting the same here, unless you can think of a better solution.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -28,7 +28,7 @@ class EspressoTemplate(EspressoTemplate_):
     of other binaries such as pw.x, ph.x, cp.x, etc.
     """
 
-    def __init__(self, binary: str = "pw") -> None:
+    def __init__(self, binary: str = "pw", test_run: bool = False) -> None:
```
Andrew-S-Rosen, line 31: When you're done, make sure to add the docstring :) 

Note: There's a small typo below ("we fall bacl on" --> "we fall back to")

src/quacc/calculators/espresso/espresso.py
```diff
@@ -95,6 +100,34 @@ def write_input(
             **parameters,
         )
 
+    def _test_run(self, parameters: dict[str, Any], directory: Path):
```
Andrew-S-Rosen, line 103: Add in `-> dict[str, Any]` for the return. Also, please turn the comment into a docstring even though it's a private function :)

src/quacc/calculators/espresso/utils.py
```diff
@@ -43,3 +46,40 @@ def parse_pp_and_cutoff(config: dict[str, Any], atoms: Atoms) -> dict[str, Any]
     tmp_input_data = {"system": {"ecutwfc": wfc_cutoff, "ecutrho": rho_cutoff}}
 
     return {"pseudopotentials": pseudopotentials, "input_data": tmp_input_data}
+
+
+def parse_ph_patterns(root_dir):
```
Andrew-S-Rosen, line 51: `parse_ph_patterns(root_dir: str | Path) -> dict:`

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+        calculation. See the ASE documentation for more information.
+
+    **calc_kwargs
+        calc_kwargs dictionary possibly containing the following keys:
+
+        - input_data: dict
+        - qpts: list[list[float]] | list[tuple[float]] | list[float]
+        - nat_todo: list[int]
+
+        See the docstring of quacc.calculators.espresso.io.write_espresso_ph for more
```
Andrew-S-Rosen, line 61: Writing this as `[quacc.calculators.espresso.io.write_espresso_ph][]` will enable it to automatically hyperlink in the online documentation.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+                "verbosity": "high",
+            }
+        },
+        "qpts": (0, 0, 0),
+    }
+
+    template = EspressoTemplate("ph", test_run=test_run)
+
+    return base_fn(
+        Atoms(),
```
Andrew-S-Rosen, line 85: No need to explicitly specify `Atoms()` since `base_fun` defaults to setting it if not specified.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+        preset=preset,
+        template=template,
+        calc_defaults=calc_defaults,
+        calc_swaps=calc_kwargs,
+        parallel_info=parallel_info,
+        additional_fields={"name": "ph.x static"},
```
Andrew-S-Rosen, line 91: Does "static" make much sense in this context? Looks like this might have changed a bit from the existing one in `main`.

src/quacc/recipes/espresso/phonons.py
Andrew-S-Rosen, line 1: Looks like some changes have been reverted from the existing `phonon_job` in `main`. I commented a few of those instances, but just do a double check. I also explained some of the differences in case you're unsure.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+    # First ph job to generate the patterns
+    ph_job_partial = partial(
+        phonon_job, test_run=True, copy_files=pw_job_results["dir_name"]
+    )
+    # We acutally run the job
+    ph_job_results = ph_job_partial()
+    # We parse the patterns
+    ph_patterns = parse_ph_patterns(ph_job_results["dir_name"])
+    # We need to get the input_data from the ph job that was sent in...
+    input_data = ph_job_results["parameters"]["input_data"]  # Better way to do that?
```
Andrew-S-Rosen, line 111: This seems fine to me.

Andrew-S-Rosen (commented):
Thanks, @tomdemeyere! This is already looking great. I added some preliminary comments for you (I know you're still working on it). Overall, the structure of the `Flow` makes sense.

> Some calculations will be able to be done very quickly because of symmetry, either find a way to trick QE into doing them at the end during the "recover" run or be able to specify another executor for them to do it either on the login node or a scavenger/serial HPC queue.

You can strip the decorator off a `Job` to turn it into a regular function via `Job.__wrapped__`. That might come in handy here, if you know in advance which ones are trivial enough to run outside of the job scheduler system.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,142 @@
...
+        input_data["inputph"]["start_q"] = pattern
+        input_data["inputph"]["last_q"] = pattern
+        n_repr = ph_patterns[pattern]
+        for representation in range(1, n_repr + 1):
+            input_data["inputph"]["start_irr"] = representation
+            input_data["inputph"]["last_irr"] = representation
+            ph_job_partial = partial(
+                phonon_job, input_data=input_data, copy_files=pw_job_results["dir_name"]
+            )
+            grid_results.append(ph_job_partial())
```
Andrew-S-Rosen, line 126: So, this may technically work in Covalent but is not a general approach. The reason for this is that `ph_patterns` is effectively a `Future` with unknown length until runtime, and so it is impossible to know _a priori_ how many iterations to loop over. This kind of pattern is described by the `@subflow` decorator in quacc, which meant when you are iterating over a bunch of `Job`s that you want to then collect.

Subflows (called sublattices in Covalent, join apps in Parsl, and subworkflows elsewhere) are generally the most difficult part of workflow management, so I'll take care of refactoring this appropriately when you're ready. You can see an example [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/common/slabs.py#L18).

tomdemeyere:
Thanks for your comments, this should be the last "core" PR for the espresso calc. If everything is working as intended, I will start to actually run production calculations using quacc. Other recipes and bug fixes should then come naturally without requiring much core changes

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+
+@flow
+def grid_phonon_flow(
+    custom_pw_job: Job | None = None,
+    custom_phonon_job: Job | None = None,
+    nblocks: int = 1,
+):
+    """
```
Andrew-S-Rosen, line 100: There are some additional arguments that should be included here, namely the `atoms`, `preset`, and `parallel_info`. 

This would then be passed to the appropriate jobs in the flow. The recipe should "just work" even if called with default settings, but right now for instance, no `Atoms` object is ever passed to `pw_job()`.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -0,0 +1,178 @@
...
+    ph_job = phonon_job if custom_phonon_job is None else custom_phonon_job
+
+    pw_job_results = pw_job()
+
+    # First ph job to generate the patterns
+    ph_job_partial = partial(
+        ph_job, test_run=True, copy_files=pw_job_results["dir_name"]
+    )
+    # We actually run the job
+    ph_job_results = ph_job_partial()
```
Andrew-S-Rosen, line 141: There is no benefit of using a partial function here. The same can be done as:

```python
ph_job_results = ph_job(test_run=True, copy_files=pw_job_results["dir_name"])
```

tomdemeyere: Oh ok, would it still take parameters sent to the original pw_job by the user or override them in this case?

src/quacc/recipes/espresso/phonons.py
Andrew-S-Rosen, line 1: @tomdemeyere: Here is a very rough sketch of the proposed changes to get you closer to what you're trying to accomplish. I tried to open a PR to your branch, but I messed something up. I'm just going to post it here as a comment instead.

```python
"""Core recipes for espresso."""
from __future__ import annotations

from pathlib import Path
from typing import TYPE_CHECKING

import numpy as np
from quacc import Job, flow, job, subflow
from quacc.calculators.espresso.espresso import EspressoTemplate
from quacc.calculators.espresso.utils import parse_ph_patterns
from quacc.recipes.espresso._base import base_fn
from quacc.recipes.espresso.core import static_job

if TYPE_CHECKING:
    from ase.atoms import Atoms
    from quacc.schemas._aliases.ase import RunSchema


@job
def phonon_job(
    preset: str | None = None,
    copy_files: str | list[str] | None = None,
    parallel_info: dict[str] | None = None,
    test_run: bool = False,
    **calc_kwargs,
) -> RunSchema:
    """
    Function to carry out a basic ph.x calculation.

    Parameters
    ----------
    preset
        The name of a YAML file containing a list of parameters to use as
        a "preset" for the calculator. quacc will automatically look in the
        `ESPRESSO_PRESET_DIR` (default: quacc/calculators/espresso/presets).
    copy_files
        List of files to copy to the calculation directory. Almost always needed
        for ph.x calculations. This parameter can either be a string or a list of
        strings.

        If a string is provided, it is assumed to be a path to a directory,
        all of the child tree structure of that directory is going to be copied to the
        scratch of this calculation. For phonon_job this is what most users will want to do.

        If a list of strings is provided, each string point to a specific file. In this case
        it is important to note that no directory structure is going to be copied, everything
        is copied at the root of the temporary directory.
    parallel_info
        Dictionary containing information about the parallelization of the
        calculation. See the ASE documentation for more information.
    test_run
        If True, a test run is performed to check that the calculation input_data is correct or
        to generate some files/info if needed.
    **calc_kwargs
        calc_kwargs dictionary possibly containing the following keys:

        - input_data: dict
        - qpts: list[list[float]] | list[tuple[float]] | list[float]
        - nat_todo: list[int]

        See the docstring of [quacc.calculators.espresso.io.write_espresso_ph][] for more information.

    Returns
    -------
    RunSchema
        Dictionary of results from [quacc.schemas.ase.summarize_run][]
    """

    calc_defaults = {
        "input_data": {
            "inputph": {
                "tr2_ph": 1e-12,
                "alpha_mix(1)": 0.1,
                "nmix_ph": 12,
                "verbosity": "high",
            }
        },
        "qpts": (0, 0, 0),
    }

    template = EspressoTemplate("ph", test_run=test_run)

    return base_fn(
        preset=preset,
        template=template,
        calc_defaults=calc_defaults,
        calc_swaps=calc_kwargs,
        parallel_info=parallel_info,
        additional_fields={"name": "ph.x Phonon"},
        copy_files=copy_files,
    )


@flow
def grid_phonon_flow(
    atoms: Atoms,
    preset: str | None = None,
    parallel_info: dict[str] | None = None,
    custom_pw_job: Job | None = None,
    custom_phonon_job: Job | None = None,
    nblocks: int = 1,
):
    """
    Function to carry out a grid parallelization of a ph.x calculation. Each representation of each
    q-point is calculated in a separate job. This allow the calculations to be run on different machines
    at possibly different times. The only drawback is that the data of the prior pw.x calculation has to
    be copied to each of the jobs. Thus the total amount of data will be n*m times the size of the pw.x
    calculation where n is the number of q-points and m is the number of representations. Plus the data
    produced by each ph.x calculations. This can becomes very large if your system has a lot of atoms.
    To provide a way to control the amount of data produced, the user can provide an additional argument
    "nblocks" which will group multiple representations together in a single job. This will reduce the
    amount of data produced by a factor of nblocks.

    Parameters
    ----------
    atoms
        Atoms object
    preset
        The name of a YAML file containing a list of parameters to use as
        a "preset" for the calculator. quacc will automatically look in the
        `ESPRESSO_PRESET_DIR` (default: quacc/calculators/espresso/presets).
    parallel_info
        Dictionary containing information about the parallelization of the
        calculation. See the ASE documentation for more information.
    custom_pw_job
        A custom job to run the pw.x calculation. This job should return a RunSchema dictionary.
        The default job is [quacc.recipes.espresso.core.static_job][].
    custom_phonon_job
        A custom job to run the ph.x calculation. This job should return a RunSchema dictionary.
        The default job is [quacc.recipes.espresso.phonons.phonon_job][].
    nblocks
        The number of representations to group together in a single job. This will reduce the amount
        of data produced by a factor of nblocks. If nblocks = 0 each job will contain all the representations
        for a single q-point.

    Returns
    -------
    RunSchema
        Dictionary of results from [quacc.schemas.ase.summarize_run][]
    """

    pw_job = static_job if custom_pw_job is None else custom_pw_job
    ph_job = phonon_job if custom_phonon_job is None else custom_phonon_job

    # We run a first pw job (single point or relax) depending on the input
    # ASR: Where is the relax job??
    pw_job_results = pw_job(atoms, preset=preset, parallel_info=parallel_info)

    # Run the phonon subflow
    grid_results = phonon_subflow(
        ph_job, nblocks=nblocks, pw_job_results_dir=pw_job_results["dir_name"]
    )

    # We have to copy back the files from all of the jobs
    copy_back = [result["dir_name"] for result in grid_results]

    # We run a 'recover' ph job to diagonalize the dynamical matrix
    # for each q-point.
    input_data = grid_results[-1]["parameters"]["input_data"]
    input_data["inputph"]["recover"] = True

    return ph_job(input_data=input_data, copy_files=copy_back)


@subflow
def phonon_subflow(ph_job: Job, pw_job_results_dir: str | Path, nblocks: int = 1):
    # We loop over the q-points and representations and run a ph job for each
    # of them

    # Run a test phonon job
    ph_test_job_results = ph_job.__wrapped__(
        test_run=True, copy_files=pw_job_results_dir
    )
    input_data = ph_test_job_results["parameters"]["input_data"]
    ph_patterns = parse_ph_patterns(ph_test_job_results["dir_name"])

    grid_results = []
    for pattern in ph_patterns:
        input_data["inputph"]["start_q"] = pattern
        input_data["inputph"]["last_q"] = pattern
        n_repr = ph_patterns[pattern]
        this_block = nblocks if nblocks > 0 else n_repr
        repr_to_do = np.array_split(
            np.arange(1, n_repr + 1), np.ceil(n_repr / this_block)
        )
        for representation in repr_to_do:
            input_data["inputph"]["start_irr"] = representation[0]
            input_data["inputph"]["last_irr"] = representation[-1]
            ph_job_results = ph_job(
                input_data=input_data, copy_files=pw_job_results_dir
            )
            grid_results.append(ph_job_results)
    return grid_results
```

src/quacc/recipes/espresso/phonons.py
Andrew-S-Rosen: If the user supplies a `custom_ph_job` with pre-specified arguments, they will be carried over and not overidden, which is the main purpose of the `partial` function. See an example [here](https://quantum-accelerators.github.io/quacc/user/recipes/recipes_intro.html#modifying-jobs-in-a-dynamic-workflow).

Andrew-S-Rosen:
Alright, I was able to run this and confirm at least the basics work (i.e. I got an output, and it didn't crash), so I merged it in the most recent commit. Feel free to build from there. I also incorporated the small changes from #1415.

tomdemeyere:
@Andrew-S-Rosen Thanks for all the work, I updated the todo list. As far as I understand, I can specify different executors for the pw job and the ph job easily. For the test and recovery runs this should be probably passed with "wrapped" as it should run fine on any computer with the binaries. Although it would be better to be able to specify different executor i.e HPC executor on a fast queue?

Andrew-S-Rosen:
@tomdemeyere: Happy to help.

> As far as I understand, I can specify different executors for the pw job and the ph job easily

Correct. That will be no issue.

> For the test and recovery runs this should be probably passed with "wrapped" as it should run fine on any computer with the binaries. 

This is technically true, but in practice we don't want to forcibly run any task involving an executable "locally". If the server (i.e. job dispatching process) is running on the login node of an HPC machine, then "local" here is fine. However, most of the workflow engines support having the job dispatching scheduler on a location external to the HPC resources themselves, e.g. a laptop or some low-cost server. In this case, we can only rely on the fact that the server environment has the `quacc` dependencies installed. Requiring the user to install Espresso in the server environment is not desirable as a general solution. 

> Although it would be better to be able to specify different executor i.e HPC executor on a fast queue?

And so with the above comment in mind, the answer to this question is "yes".

As for how to best do this from a software management standpoint, that's certainly up for discussion. It's not immediately clear to me the best approach, and I am certainly open to suggestions.

tomdemeyere:
Really just throwing an idea here without thinking about the practicalities: Would it be possible to implement presets for hpc executor (or any executor really)? Surely requiring the user to send additional executor parameters for different part of a flow will quickly get cumbersome and repetitive...

As for the implementation details I am afraid I will be of no help since I don't know much about how executors work.

Andrew-S-Rosen:
I do think something like that is the ideal way to go. Basically, have the user define a "low_cost" configuration for the executor, and we use that by default (if specified by the user as a quacc setting). I'll think about this over the next day or two. There are also one or two minor/subtle changes that need to be made to the existing flow to make it compatible with Covalent, which I'll also try and address.

tomdemeyere:
Happy that you had the same idea, some more thoughts:

- Could these preset files mix nicely with the new ASE config files?

- Should the user be allowed to pass kwargs for executor as well? What I am thinking, is to simply allow the user to pass <name_of_the_(job/flow/subflow)>_executor = ... as parameters to the parent flow. This would work for any job/flow/subflow functions that are called inside this flow and therefore allow the user complete flexibility on how each part of a flow is executed.

Andrew-S-Rosen:
> Could these preset files mix nicely with the new ASE config files?

I don't have much faith in the stability of the ASE config approach to be quite honest, so I'd personally prefer to avoid it as much as possible.

> Should the user be allowed to pass kwargs for executor as well? What I am thinking, is to simply allow the user to pass <name_of_the_(job/flow/subflow)>_executor = ... as parameters to the parent flow. This would work for any job/flow/subflow functions that are called inside this flow and therefore allow the user complete flexibility on how each part of a flow is executed.

This is a very good idea. I have some rough ideas of what this could look like. I'll give it a shot.

Andrew-S-Rosen:
@tomdemeyere:

> As a result, these two jobs will have variable computational loads at times. The flow should be able to run these jobs using different executors or finding a workaround to mimic that.

This should now be taken care of with the introduction of #1432 and the changes made in this PR. See the new sections of the documentation for details:
- [Modifying recipes](https://quantum-accelerators.github.io/quacc/user/recipes/modifications.html)
- [Modifying decorators](https://quantum-accelerators.github.io/quacc/user/wflow_engine/wflow_engines3.html).

As always, I am open to suggestions if this is not suitable.

As for this PR, the next steps from your end to get this merged would basically be:

1. Add tests for running this locally without a workflow engine, like you did with `static_job`.
2. Add any missing docstrings.

Then I will be happy to:

1. Ensure it all works with Covalent and add tests for that. There will probably be one or two minor changes needed.

Other considerations:

1. Perhaps consider adding a `relax_job` in `quacc.recipes.espresso.core` to complement `static_job`. This could be a separate PR.
2. The `phonon_job` should probably include a `conv_thr` parameter in the defaults that is tighter than the default since you often want this for phonon calculations.
3. Re-introducing nested parameter support is independently handled in #1429.

tomdemeyere:
> > Could these preset files mix nicely with the new ASE config files?
> 
> I don't have much faith in the stability of the ASE config approach to be quite honest, so I'd personally prefer to avoid it as much as possible.
> 
> > Should the user be allowed to pass kwargs for executor as well? What I am thinking, is to simply allow the user to pass <name_of_the_(job/flow/subflow)>_executor = ... as parameters to the parent flow. This would work for any job/flow/subflow functions that are called inside this flow and therefore allow the user complete flexibility on how each part of a flow is executed.
> 
> This is a very good idea. I have some rough ideas of what this could look like. I'll give it a shot in the next day or two.

I looked at the latest commit on the dynamic phonon workflow and for me, this looks like the last missing key to use quacc for production calculations.

If your goal is to be beginner-friendly: maybe it is better to 'hide' this decorator parameter into the kwargs? A little bit as I described before? I am thinking about the 'easy to use, hard to master' philosophy where the user is nicely brought to more and more complex concepts without even realizing it.

Of course, there would be a section in the documentation describing all of that, so it would not be truly 'hidden', beginners would just not see it immediately, hopefully only when they need it.

Additionally, these customizing operations could be done in the base_fn, this way if a user creates a flow, they don't even have to think about these details, it just works.

Again, just throwing ideas without thinking about the practicalities, hopefully, these inspire you.

Andrew-S-Rosen:
> > > Could these preset files mix nicely with the new ASE config files?
> > 
> > 
> > I don't have much faith in the stability of the ASE config approach to be quite honest, so I'd personally prefer to avoid it as much as possible.
> > > Should the user be allowed to pass kwargs for executor as well? What I am thinking, is to simply allow the user to pass <name_of_the_(job/flow/subflow)>_executor = ... as parameters to the parent flow. This would work for any job/flow/subflow functions that are called inside this flow and therefore allow the user complete flexibility on how each part of a flow is executed.
> > 
> > 
> > This is a very good idea. I have some rough ideas of what this could look like. I'll give it a shot in the next day or two.
> 
> I looked at the latest commit on the dynamic phonon workflow and for me, this looks like the last missing key to use quacc for production calculations.
> 
> If your goal is to be beginner-friendly: maybe it is better to 'hide' this decorator parameter into the kwargs? A little bit as I described before? I am thinking about the 'easy to use, hard to master' philosophy where the user is nicely brought to more and more complex concepts without even realizing it.
> 
> Of course, there would be a section in the documentation describing all of that, so it would not be truly 'hidden', beginners would just not see it immediately, hopefully only when they need it.
> 
> Additionally, these customizing operations could be done in the base_fn, this way if a user creates a flow, they don't even have to think about these details, it just works.
> 
> Again, just throwing ideas without thinking about the practicalities, hopefully, these inspire you.

@tomdemeyere: Great to hear this likely addresses your request. I certainly think there may be ways to make the current approach a bit less intrusive, particularly so that new recipe developers who want to contribute back to `quacc` don't have to explicitly include the `customize_funcs()` call in every `@flow` they make. I haven't quite figured out a way to achieve that but will give it thought and am happy to review specific suggestions. Personally, I am not too bothered by having `job_decorators` as an optional keyword argument, although I always would prefer to minimize code duplication where possible.

tomdemeyere:
I've added a test and can run it locally if I only run this one. If I run the full suit the previous ph.x test files break the grid test by giving the wrong pattern files. This can be fixed by using UNIQUE_DIR = True I believe. Being a potential real problem I added a parameter to the parse_pattern function that looks for the prefix of the actual calculation. If the user always uses the same dir and the same prefix this will still be an issue though.

I don't know if having both calc_defaults and a preset will make sense for ph.x here, or even for pp.x a few notes on that:

- All espresso binaries have distinct input parameters, ph.x : https://www.quantum-espresso.org/Doc/INPUT_PH.html and pw.x: https://www.quantum-espresso.org/Doc/INPUT_PW.html For example ph.x does not have 'conv_thr'.

If you insist on having a 'basic' preset for all binaries I can take care of the binary that I know.

On the relax job comment, I can add a relax_job in core.py. If it does not add real value other than being a static_job with input_data = {'calculation': 'relax'} I can say that I will almost certainly always fall back on static_job. If it does add some value, for example allowing me to use different optimizers (the ASE ones for example) seamlessly then it's another story.

Andrew-S-Rosen:
@tomdemeyere: Great, thanks! I hope the info below provides further context. If anything is unclear, just let me know. If you'd like me to sketch out a proof-of-concept, I'm happy to do that as well for any of the points below. Just let me know.

> I've added a test and can run it locally if I only run this one. If I run the full suit the previous ph.x test files break the grid test by giving the wrong pattern files. This can be fixed by using UNIQUE_DIR = True I believe. Being a potential real problem I added a parameter to the parse_pattern function that looks for the prefix of the actual calculation. If the user always uses the same dir and the same prefix this will still be an issue though.

Thanks for the head's up. This is good to know. If you have a suggestion for how to best address this, let me know. Do you think it would be beneficial from a user perspective to have the `CREATE_UNIQUE_DIR` setting enabled by default? Of course, most production calculations will require some sort of task isolation to prevent file clobbering, although new users may expect results to be written out to the current working directory upon job completion so I can see both sides of the argument.

> I don't know if having both calc_defaults and a preset will make sense for ph.x here, or even for pp.x a few notes on that:

Ah, thank you for clarifying. This is something we should sort out. In general, my philosophy has been as follows:
- All recipes should simply work "as-is" given the default parameters. It may not be the best settings, but it should run to completion for a typical material. Ideally though, the defaults should be as reasonable as possible.
- All recipes should, given the defaults, be compatible. In other words, energies from one job should be comparable to energies from another job unless the job's purpose explicitly prohibits that (e.g. `hybrid_relax_job` or something).

If each binary takes different arguments, however, then having a common default preset may not make sense here. How would you, for instance, be able to specify things like the smearing though in a phonon job? It's not part of the `&INPUTPH` block, so is it simply that this information is parsed from the prior `pw.x` calculation? I'm assuming the answer is yes based on [this page](https://pranabdas.github.io/espresso/hands-on/phonon/).

If my understanding is correct, perhaps what makes the most sense is to have the following:
- We rename "basic.yaml" to "basic_pw.yaml" to specify its purpose,
- We set the default value for the `preset` keyword to be "basic_pw" for all calculations involving `pw.x`, such that the minimal details needed to run a `pw.x` calculation are provided. Feel free to propose changes to my example [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/calculators/espresso/presets/basic.yaml).
- For calculations not running the `pw.x` binary directly, we set `preset = None` as the default, unless it makes sense to have something there (probably not needed for `pp.x` or `ph.x` if I understand correctly).

> For example ph.x does not have 'conv_thr'.

This is fine. However, for accurate phonon calculations, we want to make sure that the `pw.x` calculation has tight convergence of the energies (and forces) by default, unless the user explicitly requests otherwise. So, in the `grid_phonon_flow`, we shouldn't call `static_job` with default arguments by default. It should instead look something more like:

```python
calc_defaults = {"static_job": {"input_data": {"electrons": {"conv_thr": 1e-8}}}}
job_params = recursive_merge_dict(calc_defaults, job_params)

pw_job, ph_job, recover_ph_job = customize_funcs(
    ["pw_job", "ph_job", "recover_ph_job"],
    [static_job_, phonon_job, phonon_job],
    decorators=job_decorators,
    parameters=job_params,
)
```
or
```python
static_job_ = partial(static_job, input_data={"electrons": {"conv_thr": 1e-8}})
pw_job, ph_job, recover_ph_job = customize_funcs(
    ["pw_job", "ph_job", "recover_ph_job"],
    [static_job_, phonon_job, phonon_job],
    decorators=job_decorators,
    parameters=job_params,
)
```
both of which should be equivalent I think. I would recommend the former.

> On the relax job comment, I can add a relax_job in core.py. If it does not add real value other than being a static_job with input_data = {'calculation': 'relax'} I can say that I will almost certainly always fall back on static_job.

I do think it is valuable to have a standalone `relax_job`. There will be many times where we want specifically call a `relax_job` in a workflow, and simply being able to use the function as-is without needing to patch in `{'calculation': 'relax'}` will prevent mistakes. It also is helpful for users who may use the recipe without know much about Espresso just yet. On top of that, it is a bit awkward to run a relaxation on a recipe named `static_job`, which implies no relaxation. 

Another benefit of having a standalone `relax_job` is making it easier for the user to do common things. For instance, I would imagine there being an additional keyword argument `relax_cell: bool = False`. Then you would have something like:

```python
calc_defaults = {"calculation": "vc-relax" if relax_cell else "relax"}
```

If you added `relax_job` as a preliminary step in the `grid_phonon_flow` (in place of or preceeding `static_job`), we would want to make sure that the force tolerance (`force_conv_thr`) is set sufficiently tight just like I proposed with `conv_thr` in the `static_job` step in the flow.

>  If it does add some value, for example allowing me to use different optimizers (the ASE ones for example) seamlessly then it's another story.

This is easy to do but would be a standalone job, e.g. `ase_relax_job`. No need to worry about that right now, but it would only take me a minute or two to add once you have an existing `relax_job`. You can see [here](https://github.com/Quantum-Accelerators/quacc/blob/cee0047ceb43837d93058ba78e4721b871857696/src/quacc/recipes/qchem/core.py#L96) that it is as simple as calling a different function in `_base.py` that calls `run_opt()` instead of `run_calc()` and uses `summarize_opt_run()` instead of `summarize_run()`. That's all there is to it. I can add this in at a later PR if that would be helpful for you.

tomdemeyere:
> @tomdemeyere: Great, thanks! I hope the info below provides further context. If anything is unclear, just let me know. If you'd like me to sketch out a proof-of-concept, I'm happy to do that as well for any of the points below.
> 
> > I've added a test and can run it locally if I only run this one. If I run the full suit the previous ph.x test files break the grid test by giving the wrong pattern files. This can be fixed by using UNIQUE_DIR = True I believe. Being a potential real problem I added a parameter to the parse_pattern function that looks for the prefix of the actual calculation. If the user always uses the same dir and the same prefix this will still be an issue though.
> 
> Thanks for the head's up. If you have a suggestion for how to best address this, let me know. Do you think it would be beneficial from a user perspective to have the `CREATE_UNIQUE_DIR` setting enabled by default? Of course, most production calculations will require some sort of task isolation to prevent file clobbering, although new users may expect results to be written out to the current working directory upon job completion so I can see both sides of the argument.
> 
> > I don't know if having both calc_defaults and a preset will make sense for ph.x here, or even for pp.x a few notes on that:
> 
> If each binary takes different arguments, having a common default preset may not make sense here. How would you, for instance, be able to specify things like the smearing though in a phonon job? It's not part of the `&INPUTPH` block, so is it simply that this information is parsed from the prior `pw.x` calculation? I'm assuming the answer is yes based on [this page](https://pranabdas.github.io/espresso/hands-on/phonon/).
> 
> If my understanding is correct, perhaps what makes the most sense is to have the following:
> 
> * We rename "basic.yaml" to "basic_pw.yaml" to specify its purpose,
> * We set the default value for the `preset` keyword to be "basic_pw" for all calculations involving `pw.x`, such that the minimal details needed to run a `pw.x` calculation are provided and results are cross-compatabile by default. Feel free to propose changes to my proposal [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/calculators/espresso/presets/basic.yaml).
> * For calculations not running the `pw.x` binary directly, we set `preset = None` as the default, unless it makes sense to have something there (probably not needed for `pp.x` or `ph.x` if I understand correctly). Or perhaps it makes more sense to have no `preset` kwarg for `post_processing_job` or `phonon_job` since all of the presets are really only compatible with `pw.x` as it is.
> 
> > For example ph.x does not have 'conv_thr'.
> 
> This is fine. However, for accurate phonon calculations, we want to make sure that the `pw.x` calculation has tight convergence of the energies (and forces) by default, unless the user explicitly requests otherwise. So, in the `grid_phonon_flow`, we shouldn't call `static_job` with default arguments by default. It should instead look something more like:
> 
> ```python
> calc_defaults = {"static_job": {"input_data": {"electrons": {"conv_thr": 1e-8}}}}
> job_params = recursive_dict_merge(calc_defaults, job_params)
> 
> pw_job, ph_job, recover_ph_job = customize_funcs(
>     ["pw_job", "ph_job", "recover_ph_job"],
>     [static_job_, phonon_job, phonon_job],
>     decorators=job_decorators,
>     parameters=job_params,
> )
> ```
> 
>  > On the relax job comment, I can add a relax_job in core.py. If it does not add real value other than being a static_job with input_data = {'calculation': 'relax'} I can say that I will almost certainly always fall back on static_job.
> 
> I do think it is valuable to have a standalone `relax_job`. There will be many times where we want specifically call a `relax_job` in a workflow, and simply being able to use the function as-is without needing to patch in `{'calculation': 'relax'}` will prevent mistakes. It also is helpful for users who may use the recipe without know much about Espresso just yet. On top of that, it is a bit awkward to run a relaxation on a recipe named `static_job`, which implies no relaxation.
> 
> Another benefit of having a standalone `relax_job` is making it easier for the user to do common things. For instance, I would imagine there being an additional keyword argument `relax_cell: bool = False`. Then you would have something like:
> 
> ```python
> calc_defaults = {"calculation": "vc-relax" if relax_cell else "relax"}
> ```
> 
> If you added `relax_job` as a preliminary step in the `grid_phonon_flow` (in place of or preceeding `static_job` if requested by the user), we would want to make sure that the force tolerance (`force_conv_thr`) is set sufficiently tight just like I proposed with `conv_thr` in the `static_job` step in the flow.
> 
> I'm happy to add this as a separate PR when we're done here.
> 
> > If it does add some value, for example allowing me to use different optimizers (the ASE ones for example) seamlessly then it's another story.
> 
> This is easy to do but would be a separate, standalone job, e.g. `ase_relax_job`. No need to worry about that right now, but it would only take me a minute or two to add once you have an existing `relax_job`. You can see [here](https://github.com/Quantum-Accelerators/quacc/blob/cee0047ceb43837d93058ba78e4721b871857696/src/quacc/recipes/qchem/core.py#L96) that it is as simple as calling a different function in `_base.py` that calls `run_opt()` instead of `run_calc()` and uses `summarize_opt_run()` instead of `summarize_run()`. That's all there is to it. I can add this in at a later PR if that would be helpful for you.

This is something I discussed with @Nekkrad some time ago, I think we reached some kind of conclusion:

- For some jobs files might be important as well, so it should be easy to find them. This means that if possible they should have a different directory (UNIQUE_DIR). 
- We should be able to somewhat name them, or they should be recognizable even months later, If I run a whole lot of calculations for my PhD, how do I easily find which one is which one? In a perfect world, I understand that we might only want to keep the 'result' dict python returns but in practice, there will still be a lot of trial and error, and checking files will be necessary in one way or another.

This is when I started to play with the 'tmpdir' argument which for me at the time solved all problems (having a unique_dir, being able to run all calculations of a flow in this unique dir without moving files, and having a directory with a given name), but this is not a desirable solution. I cannot speak for all users (I don't know how easy it is for you to have that kind of input but you might want to ask around) but I would say that the top priorities are:

- No files are lost because another job overwrote them.
- No jobs fail because they read previous, unrelated files (like for the tests)

I suspect some users will expect Quacc to be production-ready without changing many parameters.

For the relax job: I just want to make sure it is not a mandatory step of any phonon workflow. Sometimes the user will send a geometry that is already at the bottom of the PES, or a transition state, or any point of the PES where we want to know the Hessian, for ML or anything else.

Finally on the presets I think your idea 'basic_pw' for pw.x and None for the rest is best.

Andrew-S-Rosen:
> We should be able to somewhat name them, or they should be recognizable even months later, If I run a whole lot of calculations for my PhD, how do I easily find which one is which one? In a perfect world, I understand that we might only want to keep the 'result' dict python returns but in practice, there will still be a lot of trial and error, and checking files will be necessary in one way or another.

This is somewhat intentional. Quacc has primarily been designed to be used with a workflow engine and/or database. As is very common when using a workflow engine and running calculations at scale, the user is predominantly interacting with the results via querying of a database. Details can be found [here](https://quantum-accelerators.github.io/quacc/user/advanced/database.html), although I should probably make it more prominent. With a database-first approach, the user really only interacts with the filesystem for debugging purposes or the occasional in-depth investigation. Each entry in the database has a field that mentions where the source files are located. This is a common learning curve when using a workflow engine, where one transitions from relying on a very organized directory structure to instead relying on querying a database.

With that said, I understand that this may not be ideal for all users, and flexibility is always nice. The user always has control over this because they can set the `RESULTS_DIR` quacc setting to whatever they like. For instance,

```python
from quacc import RESULTS_DIR

RESULTS_DIR = "/path/to/my/espresso/runs"
# then run quacc workflow for Espresso
```

or, more robustly in a workflow setting, by modifying the `QUACC_RESULTS_DIR` environment variable.

You can then change out that directory for a different project quite easily.

So, to summarize, in my view there is not really a need to change the unique directory name. Instead, the user should simply be changing the base results directory where those files are stored for each project they are interested in, assuming they want to have a more structured filesystem. I updated the documentation to better reflect this.

> No files are lost because another job overwrote them. No jobs fail because they read previous, unrelated files (like for the tests).

I think my conclusion from this is that it would be worthwhile to consider having results isolation enabled by default. There are a few caveats and subtleties to consider with this decision (e.g. some workflow engines, e.g. Covalent and Jobflow, offer a way to do this natively through the workflow engine and we want to respect that without confusing the user), but those are details I can sort out.

> For the relax job: I just want to make sure it is not a mandatory step of any phonon workflow. Sometimes the user will send a geometry that is already at the bottom of the PES, or a transition state, or any point of the PES where we want to know the Hessian, for ML or anything else.

Understood. I don't think it should be made mandatory either.

> Finally on the presets I think your idea 'basic_pw' for pw.x and None for the rest is best.

Done in `main`.

tomdemeyere:
Thanks for the details, I didn't think about using the RESULT_DIR that way to be honest.

> I think my conclusion from this is that it would be worthwhile to consider having results isolation enabled by default.

I completely agree and would like to see the UNIQUE_DIR set to true by default.

Coming back to the PR, this should be close to go. The only issue that I see is the fact that the test_run calls the binaries and is being stripped, which is something you didn't want to see (I've added a comment at the right place), unfortunately this would have to also have it's own executor. On the parameter side they should always be equal to the one of the big ph.x calculations, so nothing to change there. 

Docs should be ok for now. I am planning to write some more general docs about the Espresso calc and how to use it, and most importantly show a few examples about how to use Espresso with Quacc (separate PR)

Andrew-S-Rosen:
> Coming back to the PR, this should be close to go. The only issue that I see is the fact that the test_run calls the binaries and is being stripped, which is something you didn't want to see (I've added a comment at the right place), unfortunately this would also need to have it's own executor. On the parameter side they should always be equal to the one of the big ph.x calculations, so nothing to change there.

Great, thanks for the head's up. I'll plan to do the following in the next day or two:
1. Fix the merge conflicts
2.. Handle any remaining awkwardness with decorators/executors (won't be an issue)
3. Ensure the tests run appropriately in a remote environment with a workflow manager, and I'll add a test for this.

Afterwards, I'll ping you and you can take a look before we proceed with the merge.

> Docs should be ok for now. I am planning to write some more general docs about the Espresso calc and how to use it, and most importantly show a few examples about how to use Espresso with Quacc (separate PR)

Great!

Andrew-S-Rosen:
@tomdemeyere: Alright, I fixed the merge conflict and cleaned up a few things. Could you please do the following?
1. Please ensure that the grid phonon tests run as efficiently as possible, e.g. artificially high `kpts` or `kspacing` and such.
2. You'll notice that the `static_job()` and `relax_job()` tests both have a test that involves minimal parameters. This ensures that the function should "just work" by the user. If you could do the same for one of the grid phonon tests, that would be ideal.
3. The test is currently failing. I have not investigated as to why since it's a `ph.x` error. If I could get your assistance on resolving that, that would be a big help.

Once that is completed, I will copy the quickest `grid_phonon` test to the `tests/covalent` directory and add a Covalent-specific test there plus resolve any remaining issues that come up because of that, although none are immediately apparent to me.

tomdemeyere:
1. I've tried to minimise convergence parameters, tell me if it's still not fast enough
2. Done, FYI: QE is an iterative code, meaning that convergence is never ensured, for bulk systems simple density mixing is often more than enough. For inhomogeneous systems such as isolated molecules the 'local-TF' mixing is often necessary. Which means that there is always one of these cases that will most likely not 'work' from default settings. (Strictly speaking local-TF should work for some bulk systems but will not be efficient at all.)
3. Same reason as before. Because the tests use the same directory, previous phonon calculations write to _ph0/pwscf.phsave/patterns.i.j.xml where i is qpoint number and j is repr number. Because a previous calculation has more q-points it writes patterns.2.j.xml, as a results the parse_patterns read these files and believe that there is more q-point, hence the "wrong q_point" error from ph.x.

I can think about a few tricks fixing 3 if you do not want to put "UNIQUE_DIR" for tests

src/quacc/recipes/espresso/phonons.py
```diff
@@ -56,10 +85,192 @@ def phonon_job(
...
         copy_files=prev_dir,
     )
+
+
+# 1. The name seems confusing to me (this is grid parallelization),
+# overall I would argue to globally
+# change the name of these jobs to distinguish them from other kind of
+# phonon calculations. ph.x is not limited to phonon calculations after all.
+# 2. Shouldn't the test_job also be run in this subflow? It would be
+# easier for users who want to use this for custom workflows.
```
Andrew-S-Rosen, line 102: Putting the `test_job` inside of the `@subflow` function is not possible in a distributed workflow environment. The reason is that, in production, the `test_job` will be distributed to a remote worker and a `Future` (or equivalent) will be returned in place of the results. This `Future` cannot be used for operations like parsing the `ph_patterns` or the numpy operations performed on the expected output from `test_job`. Instead, the results of `test_job` are passed as inputs to `@subflow`, and the underlying workflow engine will use this information to construct a directed acyclic graph of compute jobs, knowing not to run the `@subflow` until the results of `test_job` are available.

This error would be come clear when running the workflow with any of the supported workflow engines.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -56,10 +85,192 @@ def phonon_job(
...
         parallel_info=parallel_info,
         additional_fields={"name": "ph.x Phonon"},
         copy_files=prev_dir,
     )
+
+
+# 1. The name seems confusing to me (this is grid parallelization),
+# overall I would argue to globally
+# change the name of these jobs to distinguish them from other kind of
+# phonon calculations. ph.x is not limited to phonon calculations after all.
```
Andrew-S-Rosen, line 100: Agreed. Feel free to rename, perhaps to `_phonon_forces_subflow` like what is done [here](https://github.com/Quantum-Accelerators/quacc/blob/10f17dda01e0594204a82e89941f5d2c0f6ed539/src/quacc/recipes/common/phonons.py#L72-L82).

The `_` in the front indicates to the user that it is not meant to be used directly (except by expert users), and the recipe will not appear in the API documentation by default. This is intentional.

In fact, one could consider moving this function _inside_ of `grid_phonon_flow` --- the only reason I didn't is in case we need to use it for some other recipe later on.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -56,10 +85,192 @@ def phonon_job(
...
+            }
+        }
+    }
+
+    job_params = recursive_dict_merge(calc_defaults, job_params)
+
+    # I got surprised by the 'basic_pw' that is send to pw_job implicitly,
+    # changing parameters I didn't set.
+    # I guess that users will keep forgetting about it and be surprised
+    # changing their parameters when writing custom flow?
```
Andrew-S-Rosen, line 238: If you find it not intuitive, you might consider adding a `preset: str = "basic_pw"` keyword argument to the `grid_phonon_flow`, which is then passed to `relax_job` directly. This will make it less hidden to the user.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -56,10 +85,192 @@ def phonon_job(
...
+    pw_job, ph_test_job, ph_job, recover_ph_job = customize_funcs(
+        ["pw_job", "cheap_job", "ph_job", "cheap_job"],
+        [relax_job, phonon_job, phonon_job, phonon_job],
```
Andrew-S-Rosen, line 241: Feel free to adjust the names as you see fit. Right now, the `ph_test_job` and `recover_ph_job` both have the same name and therefore will share the same decorator (and parameters) if specified by the user. In hindsight, I don't really think that's necessary. It probably just makes more sense to have four unique names: "pw_job", "ph_test_job", "ph_job", and "recover_ph_job". It's not any more verbose, and it's probably easier on the user. If you agree, feel free to make this change and update the names in the docstring.

Andrew-S-Rosen (commented):
@tomdemeyere: Thank you for your changes and patience! We are almost there.

> I've tried to minimise convergence parameters, tell me if it's still not fast enough

Looks good to me! Thanks!

> Done, FYI: QE is an iterative code, meaning that convergence is never ensured, for bulk systems simple density mixing is often more than enough. For inhomogeneous systems such as isolated molecules the 'local-TF' mixing is often necessary. Which means that there is always one of these cases that will most likely not 'work' from default settings. (Strictly speaking local-TF should work for some bulk systems but will not be efficient at all.)

Thanks, that's good to know. I think that's alright. DFT is complicated, and ultimately, users will be expected to have to put in some legwork into learning how the calculations are meant to be run (although helping them out a bit with sensible defaults is always nice).

> Same reason as before. Because the tests use the same directory, previous phonon calculations write to _ph0/pwscf.phsave/patterns.i.j.xml where i is qpoint number and j is repr number. Because a previous calculation has more q-points it writes patterns.2.j.xml, as a results the parse_patterns read these files and believe that there is more q-point, hence the "wrong q_point" error from ph.x.

Thanks, forgot about that. [This](https://github.com/Quantum-Accelerators/quacc/blob/main/tests/core/.quacc.yaml) is where the default parameters live. We can likely add `CREATE_UNIQUE_DIR` there for now. I will make sure that detail is sorted out. No problem.

Anyway, I left a few comments, mostly addressing your questions. Please "resolve" them when you're done, and then I'll take care of the additional test and such.

src/quacc/recipes/espresso/phonons.py
tomdemeyere: Got it, thanks for the explanation!

src/quacc/recipes/espresso/phonons.py
tomdemeyere: Ok, would it be possible to construct a custom flow with this subflow ? (I wanted to do a example in the doc)

Andrew-S-Rosen: It is certainly possible, but I don't know that I'd recommend it for most users. What is the example you had in mind? If it's to call the grid phonon workflow after a prior job, then you can make `relax_job` an optional first step and then call the `grid_phonon_flow` as a second step in your custom workflow.

Andrew-S-Rosen:
(Sorry, I don't know why the tests aren't running for you in this PR anymore. Going to sort out some issues that arose because of the update to the YAML.)

Andrew-S-Rosen (commented):
@tomdemeyere: There exists one failing test on GitHub Actions (see comment below). I also ran it via a workflow engine (Covalent in this case), which picked up another error in your `@flow` design (commented below). I don't have enough `Espresso` expertise to address them without quite a bit of digging so will need to rely on your input.

This is a simple test:

```bash
covalent start
quacc set WORKFLOW_ENGINE covalent
```

```python
import covalent as ct
from ase.build import bulk
from quacc.recipes.espresso.phonons import grid_phonon_flow

atoms = bulk("Cu")
ct.dispatch(grid_phonon_flow)(atoms)

# Then open the GUI
```

(Note: If you make changes to the code, you'll need to `covalent stop` the server, `pip install -e .` the repo again, and then `covalent start` the server so it recognizes your changes).

src/quacc/recipes/espresso/phonons.py
```diff
@@ -56,10 +81,180 @@ def phonon_job(
...
+    job_params = recursive_dict_merge(calc_defaults, job_params)
+    pw_job, ph_test_job, ph_job, recover_ph_job = customize_funcs(
+        ["pw_job", "ph_test_job", "ph_job", "ph_recover_job"],
+        [relax_job, phonon_job, phonon_job, phonon_job],
+        parameters=job_params,
+        decorators=job_decorators,
+    )
+
+    # Get the ph.x input_data from the input job parameters
+    ph_job_input_data = job_params.get("ph_job", {}).get("input_data", {})
```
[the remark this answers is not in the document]
tomdemeyere: If that's about NoneType not being iterable updating ASE should do the trick?

src/quacc/recipes/espresso/phonons.py
```diff
@@ -56,10 +81,180 @@ def phonon_job(
...
+
+    # Prep for the final recover ph.x job
+    copy_back = [result["dir_name"] for result in grid_results]
+
+    input_data = grid_results[-1]["parameters"]["input_data"]
+    for k in ["start_q", "last_q", "start_irr", "last_irr"]:
+        input_data["inputph"].pop(k)
+    input_data["inputph"]["recover"] = True
+
+    return recover_ph_job(copy_back, input_data=input_data)
```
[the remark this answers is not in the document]
tomdemeyere: New commits should fix that by changing the whole thing simply to

``` python
ph_job_input_data["recover"] = True
```

Which makes even more sense because the same base ph_input_data should be used everywhere. Although I am a little bit in the future because this requires the flat dict come-back.

src/quacc/recipes/espresso/phonons.py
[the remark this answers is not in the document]
Andrew-S-Rosen: You can see the error in the GitHub Actions log [here](https://github.com/Quantum-Accelerators/quacc/actions/runs/7409696373/job/20160455931). It is still present despite using commit 2846aa905050689d19e228696dfba5f510ee8fb4 from ASE, as noted in the install step of the logs.

src/quacc/recipes/espresso/phonons.py
tomdemeyere: Coming back quickly on that: how would you make the relax_job optional? I had a look at the doc but didn't find how to do it

Andrew-S-Rosen:
@tomdemeyere: Great news. I can confirm that this works as expected now.

I am okay with merging this at the given coverage level. When you are happy for this to be merged, please ping me and I will do so. 

Note: I will add workflow engine-specific tests in a separate PR since I have to play around with the dependencies on the GitHub Actions runners to add `Espresso`. I tested it on my machine, and it works as expected.

tomdemeyere:
Thanks for sticking on this PR, and doing quite a bit of general modifications to quacc to get it to work. I wanted to write tests for the uncovered lines but I realised this would require to add the "test_run" parameter to the jobs in core.py. I suspect I might do that in a different, much smaller PR.

Before this is merged, I just want to make sure of something that is Espresso related I will ping as soon as I am done.

Andrew-S-Rosen:
Sounds good. Thank you for helping make quacc better through your feedback and also for sticking with it as we went through this more complex workflow together. It turned out great!

tomdemeyere:
For now three problems:

- When using the "gamma trick" (this happens relatively often) representations contain only one mode, some of them will be "calculated using symmetry". Currently for large systems this will be problematic because quacc will use the 'expensive' executor for trivial jobs.

- The band structure calculation + the electrical part are always recomputed at each representation. For large systems this will represent significant overhead time. I didn't test but I believe this will also happen for Raman spectra which is very expensive.

- For now the routine copy every files, in practice this will not be possible for medium to large systems, quick tests showed me that you easily get to a few terabytes of data on medium sized systems. A large amount of this data could be avoided by copying only specific files.

I understand if you are tired of this PR and want to move along with it (I can open another PR). I can fix the two first problems alone quite quickly given the last point is fixed. As of now there is no function to copy specific files respecting directory trees.

Andrew-S-Rosen:
> When using the "gamma trick" (this happens relatively often) representations contain only one mode, some of them will be "calculated using symmetry". Currently for large systems this will be problematic because quacc will use the 'expensive' executor for trivial jobs.

Ah, okay. If you have a propose for how to handle that, by all means. In practice, these kinds of situations will inevitably happen. In case it's helpful for you or your colleagues, I generally would recommend using a workflow engine like Parsl or Dask for such calculations. The way they work is a bit different than you might be used to --- the user requests a single "pilot" job that will continually pull in jobs over the course of its lifetime. Let's say each Espresso job requires 1 node. The pilot job could be, say, a 4 node Slurm job and it would continually pull in work to run 4 individual jobs simultaneously, and when one finishes, it pulls in another until it has no work left to run.  This kind of scenario is generally much more efficient for extremely heterogeneous compute workflows because: 1) You only queue up once for the pilot job, not for each individual job; 2) If the pilot job is on your high-performance resource, it's fine if a low-cost job runs there since it will be in an out quickly, and another job will be pulled in. This is in contrast with a workflow engine like Covalent, where each job is effectively its own Slurm job tied to a given executor. Now, this isn't to say you shouldn't try to improve the flexibility in the workflow --- after all, there are strong pros and cons for each workflow engine out there. But I am just sharing this information as a practical bit of info for you in case you run production calculations.

> The band structure calculation + the electrical part are always recomputed at each representation. For large systems this will represent significant overhead time. I didn't test but I believe this will also happen for Raman spectra which is very expensive.

I see. So perhaps there can be an optional keyword argument in the `@flow` to enable or disable part of this? Feel free to propose something.

> For now the routine copy every files, in practice this will not be possible for medium to large systems, quick tests showed me that you easily get to a few terabytes of data on medium sized systems. A large amount of this data could be avoided by copying only specific files.

Indeed, I generally recommend only copying specific files for this reason. I assume you are referring to the `prev_dir` keyword argument in `phonon_job` when you say there is no mechanism for selecting specific files. I don't think this would be particularly hard to resolve. You could add an optional keyword argument to `phonon_job` that is a `list[str]` and enumerates the filenames to be copied by default from the `prev_dir`. Then in `_base_fn()` instead of passing the full directory to `copy_files`, you would pass `[prev_dir / filename1, prev_dir / filename2]` and so on, if I understand correctly. I'm not quite sure why the tree structure needs to be maintained if `ph.x` is only looking for files in its current working directory, but I trust this can be solved either way.

> I understand if you are tired of this PR and want to move along with it (I can open another PR). I can fix the two first problems alone quite quickly given the last point is fixed. As of now there is no function to copy specific files respecting directory trees.

It's best to do this right! Happy to defer to what you think is best.

tomdemeyere:
Thanks for the explanation about Parsl, the system you describe seems very handy for that specific task, I might use more than one workflow engine (it seems to me that Parsl cannot submit HPC jobs directly from a local computer? Which is why I wanted to use Covalent)

Back to the task:

The last commit aims to fix all three problems at the same time, of course, I missed the specific keywords 'only_init' in ph.x for grid parallelization... The keywords fix every problem by pre-calculating everything during the 'init' phase (previously the test job). Every other job is then set to recover = .true. and therefore avoid recalculating anything more than once. It also allows to have a better description of what needs to be calculated or not, quacc is now able to skip all the modes calculated using symmetry, they will be all done at the very end in the 'recover_job'. This will require an update on the ASE side (read_espresso_ph) but it is for the best. In theory, everything comes together very nicely.

For this purpose I had to add a case in copy_files if the user sends a dict, this would be some kind of 'expert' mode where keys are base directories and value(s) are paths that will be resolved relative to their key. You are also allowed to specify glob patterns. If you are not happy with it I am open to any alternative route!

Andrew-S-Rosen:
> Thanks for the explanation about Parsl, the system you describe seems very handy for that specific task, I might use more than one workflow engine (it seems to me that Parsl cannot submit HPC jobs directly from a local computer? Which is why I wanted to use Covalent)

Parsl does have a way to submit jobs to the remote machine [via SSH](https://parsl.readthedocs.io/en/stable/stubs/parsl.channels.SSHChannel.html), although I must admit that I have never used it. Dask, which uses a pilot job model like Parsl, does not support remote job submission.

> The last commit aims to fix all three problems at the same time, of course, I missed the specific keywords 'only_init' in ph.x for grid parallelization... The keywords fix every problem by pre-calculating everything during the 'init' phase (previously the test job). Every other job is then set to recover = .true. and therefore avoid recalculating anything more than once. It also allows to have a better description of what needs to be calculated or not, quacc is now able to skip all the modes calculated using symmetry, they will be all done at the very end in the 'recover_job'. This will require an update on the ASE side (read_espresso_ph) but it is for the best. In theory, everything comes together very nicely.

Sounds good. I will formally review the PR shortly.

src/quacc/utils/files.py
```diff
@@ -69,9 +69,57 @@ def copy_decompress_files(
...
+            copy_decompress_files_from_dir(f_path, destination)
         else:
             warnings.warn(f"Cannot find file {f_path}", UserWarning)
 
+def copy_decompress_tree(
+    source_files: dict[str, str | Path | list[str | Path]], destination: str | Path
+) -> None:
+    """
+    Copy and decompress files from source to destination.
+    This function respects the directory tree.
```
Andrew-S-Rosen, line 82: It would be very helpful if you could add unit tests for this function, independent of Espresso. I will probably want to refactor this at some point but would be hesitant to do so without the peace of mind of simple-to-run tests.

src/quacc/calculators/espresso/utils.py
```diff
@@ -43,3 +46,39 @@ def get_pseudopotential_info(
             ecutrho = pp_dict[element]["cutoff_rho"]
         pseudopotentials[element] = pp_dict[element]["filename"]
     return ecutwfc, ecutrho, pseudopotentials
+
+
+def parse_ph_patterns(root_dir: str | Path, prefix: str) -> dict[int, int]:
```
Andrew-S-Rosen, line 51: This function isn't used anywhere anymore, right? If so, please remove the function plus the imports introduced by this change.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -45,21 +72,225 @@ def phonon_job(
...
+    )
+
+    grid_results = _grid_phonon_subflow(
+        ph_job_input_data, ph_test_job_results, ph_job, nblocks=nblocks
+    )
+
+    prev_dirs = {}
+
+    for result in grid_results:
+        prev_dirs[result["dir_name"]] = ["**/dynmat.*.*.xml", "**/tensors.xml"]
```
Andrew-S-Rosen, line 292: While I have not tested it yet, I imagine that this will not work. You are trying to populate a dictionary with a future value that is not known until runtime. In other words, the key `results["dir_name"]` is not known _a priori_ and will not behave as expected. Assignments like this should not be done within a `@flow`.

The previous scenario was:

```python
copy_back = [result["dir_name"] for result in grid_results]
ph_job_input_data["recover"] = True
return recover_ph_job(copy_back, input_data=ph_job_input_data)
```

This worked because `copy_back` contains a `list` of future objects. No operations are done on these future objects; rather, they are passed to `recover_ph_job`, which then knows to resolve the list of futures before starting that calculation. That is about the extent of things you can do with futures in a `@flow`.

src/quacc/utils/files.py
```diff
@@ -69,9 +69,57 @@ def copy_decompress_files(
...
+
+            if abs_f.is_symlink():
+                continue
+            if abs_f.is_file():
+                copy(abs_f, Path(destination, rel_f))
+                decompress_file(Path(destination, rel_f))
+            elif abs_f.is_dir():
+                copy_decompress_files_from_dir(abs_f, Path(destination, rel_f))
+            else:
+                warnings.warn(f"Cannot find file {abs_f}", UserWarning)
```
Andrew-S-Rosen, line 121: Would be good to try to avoid code duplication here with content from `copy_decompress_files`. Maybe there is a way to refactor accordingly?

src/quacc/utils/files.py
```diff
@@ -69,9 +69,57 @@ def copy_decompress_files(
         if f_path.is_file():
             copy(f_path, Path(destination, f_path.name))
             decompress_file(Path(destination, f_path.name))
+        elif f_path.is_dir():
+            copy_decompress_files_from_dir(f_path, destination)
```
Andrew-S-Rosen, line 73: I didn't comment on this previously, but this behavior (while essential for Espresso) will probably introduce confusion to some users/developers of other codes down the road. I imagine the default thought process for most people will be that if they provide a directory, that the directory will be copied (e.g. `cp -r` style) and then any files inside unzipped (`gunzip *`), whereas here what is really happening is a `cp *` on all the files _inside_ the directory. I will probably add a keyword argument at some point to more clearly address this.

src/quacc/runners/prep.py
```diff
@@ -64,6 +65,8 @@ def calc_setup(
         symlink.symlink_to(tmpdir, target_is_directory=True)
 
     # Copy files to tmpdir and decompress them if needed
```
Andrew-S-Rosen, line 67: With the change introduced in `copy_decompress_files` you can likely shorten this to

```python
if isinstance(copy_files, dict):
    copy_decompress_tree(copy_files, tmpdir)
if isinstance(copy_files, list):
    copy_decompress_files(copy_files, tmpdir)
elif isinstance(copy_files, (str, Path)):
    copy_decompress_files([copy_files], tmpdir)
```

if I'm understanding it correctly. No need to call `copy_decompress_files_from_dir` directly since that will be called from within `copy_decompress_files` if the list item `is_dir()`.

Andrew-S-Rosen (commented):
Hi @tomdemeyere: Thanks for the revisions! Some comments below, mostly about file I/O stuff and also one likely bug in the `@flow` as-constructed.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -45,21 +72,225 @@ def phonon_job(
...
+                    "verbosity": "high",
+                }
+            }
+        },
+        "ph_job": {
+            "input_data": {
+                "inputph": {
+                    "tr2_ph": 1e-12,
+                    "alpha_mix(1)": 0.1,
+                    "verbosity": "high",
```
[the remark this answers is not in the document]
tomdemeyere: Sorry I should have warned you: I am still actively working on this! Very sorry about that, I will try to name my commits better "not done" or something similar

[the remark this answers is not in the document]
Andrew-S-Rosen: Ah, sorry @tomdemeyere! I don't know why I assumed it was ready. That was on me not paying good enough attention. Anyway, now you get comments regardless :P

src/quacc/recipes/espresso/phonons.py
tomdemeyere: (I have seen your edit) I believed this would work because it worked with list... I thought they are both getter/setter under the hood and I would not make any difference.

Would some kind of deepcopy or something else would make sure that it always work?

Andrew-S-Rosen: The deepcopy won't be relevant here. Basically, the way things work in `@flow`s is that every return object from a `@job` is a `Future` (or equivalent), and you can't do any* operations on a `Future` other than pass it as the input to another `@job`. When it's passed to the `@job`, it will be auto-resolved inside that `@job`. The one exception is, generally, a `__getitem__` call where doing `Future["key"]` will basically defer the evaluation of `__getitem__(self, key)` until the `Future` is in fact resolved, e.g. something we often do where we pass `results["atoms"]` as the output of one `@job` to the input of another. It knows not to index it until later.

Before, you were appending `Future`s to a list (this is fine; the operation doesn't care what the value of `Future` will be) and passing that `list[Future]` to a `@job`. All okay. Here, you are effectively doing `dict[Future] = list[str]`. The key is not a `str` but rather a `Future`, so it's unclear to me in advance what will happen. I don't know off-hand how the auto-resolving of `Future`s happens across the various workflow engines.

This didn't work before because you were effectively doing `Future[str] = bool`, and you can't assign a property to a `Future`.

```python
input_data = grid_results[-1]["parameters"]["input_data"]
input_data["inputph"]["recover"] = True
```

In this commit, you're assigning a property to a dictionary whose key is a `Future`, and my brain hurts thinking about it so tl;dr --- there will only one way to find out...

tomdemeyere:
OK, so:

- All three problems are fixed. By default the function is set to put performance first, at the expense of (probably tons) of disk space. This balance can be fine tuned in many ways and this will be described in the doc.

- I had to change quite a lot of things, I will also open a ASE PR and ping you if it's ok with you.

I just added a test for the copy_decompress_tree

Andrew-S-Rosen (commented):
@tomdemeyere: Great, this is looking good. Still one issue with your `@flow` pattern. Let me know if you would like assistance with it. 

Will wait for the ASE MR and then hopefully this will pass and we can get to merging it when you're ready.

tomdemeyere:
Can/should the internal functions be located in another file? Users will not be interested by that and/or this might confuse them?

Andrew-S-Rosen:
> Can/should the internal functions be located in another file? Users will not be interested by that and/or this might confuse them?

Right now, I have all internal functions that are needed for recipes located inside their respective recipe, an example of which can be found [here](https://github.com/Quantum-Accelerators/quacc/blob/84effa261436a73d162dd9a45f3a6868478d8342/src/quacc/recipes/common/phonons.py). It would probably be wise to do the same with `_grid_phonon_subflow`.

Whether it, more generally would be valuable to move them to some other module is a good question. The pros are that the recipe becomes easier to read. The con is that the relationship between the internal function and its corresponding recipe becomes less clear. Personally, I would probably suggest moving `_grid_phonon_subflow` inside and then we can consider alternate ideas afterwards if there is a good proposal.

tomdemeyere:
>  Edit: Perhaps one could make a _phonons.py and move them there? What are your thoughts given these points?

For now, I moved them both inside as you suggested. I am not sure what is the best choice here.

This should be close to go! To my knowledge quacc is now the only package that allows running such flexible ph.x grid-parallelization. If file transfer between machines becomes routine this will allow to run very large phonon calculations

Andrew-S-Rosen:
FYI. All looks good on several tested workflow engines! I will try to merge this sometime tomorrow. I don't anticipate that anything else will be needed from you on this. [The codecov test will fail because of the coverage, but there has been substantial enough work on this that I feel okay with merging it after my final review.]

image
image

tomdemeyere:
> FYI. All looks good on several tested workflow engines! I will try to merge this sometime tomorrow. I don't anticipate that anything else will be needed from you on this. [The codecov test will fail because of the coverage, but there has been substantial enough work on this that I feel okay with merging it after any final edits.]
> 
> 
> 
> Sidenote: This only works on the newest (0.232.0) release of Covalent because your workflow wrecked their memory usage on the older versions. The newest version handles this better, but it does not install with `pip install quacc[covalent]` due to a [dependency conflict](https://github.com/AgnostiqHQ/covalent/issues/1871). I also tested this on Parsl and it seems to work fine there. I anticipate it will work fine on Dask, Redun, and Prefect (to be added soon).
> 
> 
> 
> image
> 
> image
> 
> 
> 
> 

Amazing! After the merge, I will run a first example using HPC. (Possibly using Covalent, except if the 'electron Dict' executor thing is not fixed. In which case I will fall back on another workflow engine with an HPC executor)

When that works, and after improving the documentation I will start to share my quacc work (Blog post and probably on the QE mailing list)

Andrew-S-Rosen:
Alright! I'm going to merge this now for the sake of both of us! 😅 Thank you again for your really nice work on this!! 

> Amazing! After the merge, I will run a first example using HPC. (Possibly using Covalent, except if the 'electron Dict' executor thing is not fixed. In which case I will fall back on another workflow engine with an HPC executor)

My personal suggestion is that Covalent will not be the most efficient way to run this type of workflow, but only you'll be able to judge that for sure given the many intricacies of HPC policies.

> When that works, and after improving the documentation I will start to share my quacc work (Blog post and probably on the QE mailing list)

Awesome!

Andrew-S-Rosen:
@tomdemeyere: Was `test_run` supposed to be `True` here? https://github.com/Quantum-Accelerators/quacc/blob/1606f2ce4147d4c3d2bf0dbc6d78ec3bf14000fa/src/quacc/recipes/espresso/phonons.py#L258-L260

tomdemeyere:
> @tomdemeyere: Was `test_run=True` supposed to be included` here?
> 
> https://github.com/Quantum-Accelerators/quacc/blob/1606f2ce4147d4c3d2bf0dbc6d78ec3bf14000fa/src/quacc/recipes/espresso/phonons.py#L258-L260

No, it is entirely replaced by this “only_init” keyword

PR #1410, opened 2023-12-25
Import `PH_KEYS` from ASE

Andrew-S-Rosen:


Andrew-S-Rosen:
@tomdemeyere: FYI, you'll need to update ASE if you haven't done so already.

```bash
pip install --force-reinstall --no-deps https://gitlab.com/ase/ase/-/archive/master/ase-master.zip
```

tomdemeyere:
About that, I started to write some reformatting for ASE. This will include keys for most QE binaries and be located in ase.io.espresso_keys.keys. I also created one read and write for arbitrary Fortran namelist. After that quacc should be espresso.io free. I will open the PR as soon as it is ready

Andrew-S-Rosen:
Awesome. Thank you so much!

PR #1427, opened 2023-12-27
Implement a minimal preset for Espresso

### LINKED ISSUE #1412: Espresso `static_job` doesn't work "as-is"

- **quacc version**: 0.4.6
- **Python version**: 3.9+

### What is the issue?

All quacc recipes should have an option that works "as-is" given the default settings. However, doing a default calculation with bulk Cu yields an error that no smearing is applied. A default preset should be selected, even if it is not perfect.

I would generally recommend Gaussian smearing with a sufficiently small `degauss`, which is safe for all kinds of systems.

```python
from ase.build import bulk
from quacc.recipes.espresso.core import static_job

atoms = bulk("Cu")
static_job(atoms)
```

### Suggestions

Maybe we have a `BasicSet.yaml` that looks something like:

```yaml
system:
  ecutwfc: 60
  ecutrho: 240
  occupations: smearing
  degaus: 0.001

kspacing: ???

parent_pseudopotentials: sssp_1.3.0_pbe_efficiency.yaml
```

And then have each recipe have `preset: str | None = "BasicSet"`

Andrew-S-Rosen:
Closes #1412. Also, I removed default parameters like `ecutwfc` and `ecutrho`, which should not be in the recipe defaults since they would be the same defaults for every recipe. The recipe defaults should have the minimal info needed to run a reliable calculation for that calculation type; we can put a default preset ("minimal.yaml") that will automatically set `ecutwfc` and `ecutrho` to the max value based on the pseudopotential YAML instead.

### DIFF
--- a/src/quacc/calculators/espresso/presets/basic.yaml
+++ b/src/quacc/calculators/espresso/presets/basic.yaml
@@ -1,0 +1,10 @@
+# Last modified: 12-27-2023
+
+input_data:
+  system:
+    occupations: smearing
+    degauss: 0.001
+
+kspacing: 0.03
+
+parent_pseudopotentials: sssp_1.3.0_pbe_efficiency.yaml
--- a/src/quacc/recipes/espresso/core.py
+++ b/src/quacc/recipes/espresso/core.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
 @job
 def static_job(
     atoms: Atoms,
-    preset: str | None = None,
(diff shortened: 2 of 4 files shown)


tomdemeyere:
I will probably open a small PR on that, as currently 'basic' will not work with the phonon_job (all qe binary have very different keywords) I think at the very least we will need to have a 'basic' for all implemented binaries (or None)

PR #1429, opened 2023-12-28
[WIP] Re-add support for flat input parameter format in Espresso

### LINKED ISSUE #1434: Re-add support for the "flat" input parameter format in Espresso recipes

See #1429

Andrew-S-Rosen:
Closes #1434.

@tomdemeyere: This is a draft PR to re-add flat input parameter support in Espresso. However, there are some stumbling blocks I don't quite know how to resolve.

Right now, the `input_data` keyword argument passed to the `Espresso()` calculator is converted into the nested form if needed. Additionally, any input parameters in the `input_data` section of a preset are converted into the nested form if needed (even though all the ones we ship in quacc are nested already). There are no issues there.

However, we also need to do it a third and fourth time [here](https://github.com/Quantum-Accelerators/quacc/blob/cc73104908cde675bb32ef15bed39e2166fe513b/src/quacc/recipes/espresso/_base.py#L67) on both `calc_defaults` and `calc_swaps` to ensure that the dictionary merge is always done appropriately. I suppose that's not a huge issue because I can get the binary name from the `template` and do the same thing as elsewhere.

The bigger issue is that users can _also_ pass in the flat form not just via `input_data` but also just as regular kwargs. For instance, see the example from the [ASE docs](https://wiki.fysik.dtu.dk/ase/ase/calculators/espresso.html#parameters):

```python
input_data = {
    'system': {
      'ecutwfc': 64,
      'ecutrho': 576
     },
    'disk_io': 'low'}  # automatically put into 'control'

calc = Espresso(profile=my_predefined_profile,
                pseudopotentials=pseudopotentials,
                # Optionally, if the ESPRESSO_PSEUDO env var isn't set
                pseudo_dir=path_to_pseudopotentials,
                tstress=True, tprnfor=True,  # kwargs added to parameters
                input_data=input_data)
```

It was at this point that I simply said "okay, this is silly, let's just require the nested form", but I can understand that this may not be ideal if you find that users like the flat form. Having three different ways of specifying input parameters is quite a mess and probably shouldn't have existed in the first place... 🤷‍♂️ 

TODO:
- Ensure tests pass
- Re-add explicit tests for flat in --> nested out.


tomdemeyere:
Few notes:

- If ASE is still ok with it, keys will all go there.
- All the kwargs should be catched at calc_swaps=calc_kwargs in the base_fn call, calling a construct_namelist in base_fn should automatically deal with all the mess (including additional flat kwargs).
- calc_defaults should always be nested, but it doesn't hurt to add an additional call for user defined flow/jobs...
- Due to how presets are read, this should go naturally inside construct_namelist if additional kwargs are added in a preset by a user.

I will have a look tomorrow

Andrew-S-Rosen:
Thanks! That's easy enough to implement. I can make sure that happens.

Andrew-S-Rosen:
> Few notes:
> 
> * If ASE is still ok with it, keys will all go there.
> * All the kwargs should be catched at calc_swaps=calc_kwargs in the base_fn call, calling a construct_namelist in base_fn should automatically deal with all the mess (including additional flat kwargs).
> * calc_defaults should always be nested, but it doesn't hurt to add an additional call for user defined flow/jobs...
> * Due to how presets are read, this should go naturally inside construct_namelist if additional kwargs are added in a preset by a user.
> 
> I will have a look tomorrow

Note: This will address most of the problems but not if the user does

```python
calc = Espresso(profile=my_predefined_profile,
                tstress=True, tprnfor=True,  # for some reason this is allowed outside of `input_data`.....?
                input_data=input_data)
```

Either way, I think that's fine.

tomdemeyere:
I have a branch (that is based on the current main) that addresses the various issues and use the new Namelist stuff from ASE should I open a new PR?

Andrew-S-Rosen:
@tomdemeyere That would be greatly appreciated!

PR #1446, opened 2023-12-31
Add a simple Espresso `relax_job`

Andrew-S-Rosen:
Will add more tests after a separate PR.


src/quacc/recipes/espresso/core.py
```diff
@@ -74,11 +74,73 @@ def static_job(
...
+    }
+
+    return base_fn(
+        atoms,
+        preset=preset,
+        template=EspressoTemplate("pw"),
+        calc_defaults=calc_defaults,
+        calc_swaps=calc_kwargs,
+        parallel_info=parallel_info,
+        additional_fields={"name": "pw.x Static"},
```
tomdemeyere, line 136: "pw.x Relax" ?

Andrew-S-Rosen: Patched in main, thanks!

PR #1447, opened 2023-12-31
Fix a few Espresso results parsing bugs

Andrew-S-Rosen:
### Checklist

- [X] I have read the [Contributing Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉
- [X] I have used `black`, `isort`, and `ruff` as described in the [style guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html#style).
- [X] I have added relevant, comprehensive [unit tests](https://quantum-accelerators.github.io/quacc/dev/contributing.html#unit-tests).

### Notes

## Summary of Changes

This PR addresses several existing issues with the Espresso calculator and recipes, as enumerated below in decreasing order of importance.

1. The Espresso recipes never updated the `Atoms` object in the returned output when a structure relaxation occurs. That is because a `geom_file` keyword argument needed to be passed to `run_calc()`. There is inconsistent behavior in ASE where some calculators update the `Atoms` object when an "internal" relaxation is done, whereas others don't (see ASE [issue 1368](https://gitlab.com/ase/ase/-/issues/1368)). In practice, what this means is that we must explicitly tell quacc which file is the geometry log file, which is then parsed at the end of a run and used to update the `Atoms` object in the returned schema. I have set this to be `pw.out` if `pw` is the binary, otherwise no geometry file is read (same behavior as before).

2. All quacc recipes dump out the `calc.results` dictionary in the output schema (see [here](https://github.com/Quantum-Accelerators/quacc/blob/606617717dc334219d8432046cd99a2addd99851/src/quacc/schemas/_aliases/ase.py#L14-L22)). This, by default, typically contains properties like "energy" and "forces" (i.e. the `implemented_properties` from the calculator). However, before this PR, this does not appear to be done for `pw.x` runs, as there was no "energy" key (for instance). This is now addressed by fixing the `read_results` method of the `Espresso` calculator. I then added tests to evaluate the energy from the static job runs.

3. If the preset has `kspacing` in it but the user specifies `kpts` explicitly when calling a recipe, an error is raised that both `kpts` and `kspacing` can't be set. This PR now checks to see if the user has specified `kpts`, and if so, `kspacing` will be dropped from the preset (and vice versa).

TODO:
- Speed up the cell relaxation test because it takes too long


Andrew-S-Rosen:
@tomdemeyere if you have a moment, it would be great to get your feedback on these bugfixes.

tests/core/recipes/espresso_recipes/test_espresso.py
```diff
@@ -207,10 +220,17 @@ def test_relax_job_cell(tmp_path, monkeypatch):
         relax_cell=True,
         input_data=input_data,
         pseudopotentials=pseudopotentials,
-        kspacing=0.5,
+        kpts=(1, 1, 1),
```
tomdemeyere, line 223: FYI: kpts = (1, 1, 1) leads to gamma with complex wf
kpts = None leads to gamma with real wf (~ 80% faster)

Most of the time your want the later except if you do some special ph.x calc later on or want non-collinear spin

src/quacc/recipes/espresso/_base.py
```diff
@@ -75,7 +75,9 @@ def base_fn(
         **calc_flags,
     )
 
-    final_atoms = run_calc(atoms, copy_files=copy_files)
+    geom_file = template.outputname if template.binary == "pw" else None
+
+    final_atoms = run_calc(atoms, geom_file=geom_file, copy_files=copy_files)
```
tomdemeyere, line 80: Had a look at run_calc(), maybe we even want to get all atoms using index = "::" so that the final result dict of the job contains all atoms object? This would be needed if the user run calculation = 'md' for example

Andrew-S-Rosen: Good point. This hasn't come up before because in other recipes where I'm running internal relaxations (e.g. Vasp, ORCA, Gaussian, Q-Chem), there are custom output file parsers being called (e.g. pymatgen, cclib) that fetch that data automatically. But here, we are relying on data obtained solely from ASE, and so that would be worth adding. Opened an issue report in #1451.

PR #1454, opened 2024-01-01
Espresso: the nested come back

tomdemeyere:
### Checklist

- [ ] I have read the [Contributing Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉

### Notes

## Summary of Changes

Ok, this is getting a little bit silly... The flat/nested form not being the problem, the most annoying is the possibility for any kwargs to be a keyword...

Without thinking about that (I never use the ASE calculator this way) I named one of the parameters in ase.io.espresso.write_espresso_ph, to the name of one of the ph.x parameters, which is conflicting now (you will see the test fail). I will need to open a small patch on the ASE main to change the name...

This also try to extend the recursive_dict_merge to any dict-like (UserDict, defaultdict, OrderedDict, Namelist) object and allow to conserve the class even if they are nested.

### DIFF
--- a/src/quacc/utils/dicts.py
+++ b/src/quacc/utils/dicts.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 from copy import deepcopy
 from typing import TYPE_CHECKING
+from collections.abc import MutableMapping
 
 if TYPE_CHECKING:
     from typing import Any
@@ -41,8 +42,8 @@ def recursive_dict_merge(
 
 
 def _recursive_dict_pair_merge(
-    dict1: dict[str, Any] | None, dict2: dict[str, Any] | None
-) -> dict[str, Any]:
+    dict1: MutableMapping[str, Any] | None, dict2: MutableMapping[str, Any] | None
+) -> MutableMapping[str, Any]:
     """
     Recursively merges two dictionaries. If one of the inputs is `None`, then it is
     treated as `{}`.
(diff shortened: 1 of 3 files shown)


buildbot-princeton:

src/quacc/utils/dicts.py
```diff
@@ -63,13 +64,16 @@ def _recursive_dict_pair_merge(
     dict
         Merged dictionary
     """
-    dict1 = dict1 or {}
-    dict2 = dict2 or {}
+
+    dict1 = dict1 or (dict1.__class__() if dict1 is not None else {})
+    dict2 = dict2 or (dict2.__class__() if dict2 is not None else {})
```
Andrew-S-Rosen, line 69: Could you walk me through this one? The only time the first expression in the `or` boolean would be `False` is if `dict1` is `None` or `False` or some other falsiness value (e.g. `[]`, `{}`, 0). How/when does it get to the second expression `(dict1.__class__() if dict1 is not None else {})`?

Andrew-S-Rosen (commented):
Thanks for this quick patch, @tomdemeyere! Yes, the "any kwargs being a keyword" thing was where I threw up my hands up in the air and gave up. 😅 But regardless, I'll be happy to get this merged once it's all ready to go with ASE upstream.

Just on question below so I can understand a bit better.

src/quacc/utils/dicts.py
tomdemeyere: The plan is to make sure if you send an empty kind of dict class it does not come back as a 'normal dict' which was the case before?

Andrew-S-Rosen: Oh, I only just learned that `{}` has a `.__class__()` attribute! I never even thought about the attributes of `{}`.

tomdemeyere: The biggest surprise for me was that None has a .__class__ attribute...

I guess it does not really matter, it is just to avoid any surprise to the user who would use this function and be surprised by the class change.

Andrew-S-Rosen:
@tomdemeyere: FYI, Just waiting on tests to pass for this before I can merge.

PR #1456, opened 2024-01-03
ORCA ASE relax job

Andrew-S-Rosen:
This PR adds support for an ASE-based ORCA relaxation with ORCA as the energy/force driver and ASE as the optimizer.


Andrew-S-Rosen:
@espottesmith: Once merged, to run with Sella, do the following.

```bash
pip install --force-reinstall --no-deps https://gitlab.com/ase/ase/-/archive/master/ase-master.zip
pip install git+https://github.com/quantum-accelerators/quacc.git
pip install sella
```

```python
from ase.io import read
from sella import Sella
from quacc.recipes.orca.core import ase_relax_job

atoms = read("MyFavoriteMol.xyz")
output = ase_relax_job(atoms, opt_params={"optimizer": Sella})
```

PR #1465, opened 2024-01-04
Avoid surprises deleting dense k-spacing from basic_pw preset

tomdemeyere:
Situation: A user is used to the ASE espresso calculator and wants to run a very large isolated system in a big box. The user doesn't set any kpts because they are used to Gamma being the default. Instead, quacc set a very high kspacing, which in this case will ruin the calculation.

This patch aims to avoid that.

## Summary of Changes

\>> Provide a description of your changes here, linking to any associated issue reports. <<

### Checklist

- [ ] I have read the [Contributing Guide](https://quantum-accelerators.github.io/quacc/dev/contributing.html). Don't lie! 😉

### Notes

### DIFF
--- a/src/quacc/calculators/espresso/presets/basic_pw.yaml
+++ b/src/quacc/calculators/espresso/presets/basic_pw.yaml
@@ -1,10 +1,10 @@
-# Last modified: 12-27-2023
+# Last modified: 04-01-2024
 
 input_data:
   system:
     occupations: smearing
-    degauss: 0.001
+    degauss: 0.005
 
-kspacing: 0.03
+kspacing: gamma
 
 parent_pseudopotentials: sssp_1.3.0_pbe_efficiency.yaml
--- a/src/quacc/calculators/espresso/presets/basic_pw.yaml
+++ b/src/quacc/calculators/espresso/presets/basic_pw.yaml
@@ -5,6 +5,6 @@ input_data:
     occupations: smearing
     degauss: 0.005
 
-kspacing: gamma
+kpts: gamma
 
 parent_pseudopotentials: sssp_1.3.0_pbe_efficiency.yaml


buildbot-princeton:

Andrew-S-Rosen:
@tomdemeyere: Thank you for this PR. I would like to propose an alternate mechanism.

We are on the same page that quacc recipes should be as user-friendly as possible. Importantly, this should also be true for users who are either new to `Espresso` or who are using `Espresso` in a high-throughput setting. For this reason, I think it's important to have a practical, default set of parameters (where practical) rather than simply a mirror of the default ASE settings.

I am proposing in #1467 to improve the default k-point density by using the standard applied in the Materials Project --- choosing a k-point grid that scales inversely with respect to the lattice parameters of the `Atoms` object. Then, if a user passes in a large (or isolated, with lots of vacuum) structure, they will have a small k-point grid and if they pass in a small structure, it will have many k-points. Please let me know your thoughts.

Andrew-S-Rosen:
Perhaps it makes the most sense to just make the default preset be `"sssp_1.3.0_pbe_efficiency"` in light of the prior discussion.

tomdemeyere:
For me this is what would make more sense. This would allow to have calculations which at least will run by having default pseudopotentials and cutoffs. This is where I would stop i.e. ensuring that the calculation launch. I would let the entire responsibility to the user to choose the best parameters for his calculation to converge let alone have sensible results.

Personal view: I am no UX expert but I think it also increases the affordance for the whole preset concept. I would feel more empowered discovering presets with one that links to a pseudopotentials table (don't ask why 😅). After seeing the other presets and reading the doc about it, I am just convinced, let's use them! If I see a basic preset, something feels forced, something feels wrong. But that's totally my view (probably over-thinking it). I also understand your position wanting to have presets which will help convergence for the general case and currently thinking about a better kspacing default.

Andrew-S-Rosen:
@tomdemeyere: Understood! Happy to defer to that approach instead. I will take care of that in #1472, which will in turn close this PR as obsolete.

PR #1503, opened 2024-01-10
[Not for merge] Branch of #1489

Andrew-S-Rosen:
This is a branch off of #1489 that makes the tests pass by moving the `TYPE_CHECKING` blocked imports. It is not made in #1489 itself because I have not 100% decided how I want to handle this, since there could be further updates in the upstream Prefect PR that I'm waiting to get merged.

Note: While the tests pass, there is one additional quality-of-life change that will eventually be needed. As noted in #1489, it would be nice to have the `@subflow` decorator automatically do a gather-type operation like in Dask.


Andrew-S-Rosen:
Closing in favor of #1489, which resolves the import problem via https://github.com/PrefectHQ/prefect/pull/11616.

PR #1515, opened 2024-01-11
Proposal: improve espresso settings in quacc.yaml

tomdemeyere:
### Checklist

### Notes

## Summary of Changes

As of now there is no way to change binary location neither using .quacc.yaml or the CLI? (Can't change the dictionary directly)

This additional keyword would allow it. If there is already a way then these changes are obsolete but this PR could then be used to explain it better in the doc

### DIFF
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -299,7 +299,7 @@ class Espresso(Espresso_):
 
         template = template or EspressoTemplate("pw")
         full_path = Path(
-            SETTINGS.ESPRESSO_BIN_DIR, SETTINGS.ESPRESSO_BIN_PATHS[template.binary]
+            SETTINGS.ESPRESSO_BIN_DIR, SETTINGS.ESPRESSO_BINARIES[template.binary]
         )
         self._bin_path = str(full_path)
         self._binary = template.binary
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -298,7 +298,8 @@ class Espresso(Espresso_):
         self._user_calc_params = {}
 
         template = template or EspressoTemplate("pw")
-        self._bin_path = str(SETTINGS.ESPRESSO_BIN_PATHS[template.binary])
+        full_path = Path(SETTINGS.ESPRESSO_BIN_DIR, SETTINGS.ESPRESSO_BIN_PATHS[template.binary])
+        self._bin_path = str(full_path)
         self._binary = template.binary
         self._cleanup_params()
         self._pseudo_path = (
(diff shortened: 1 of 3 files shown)


buildbot-princeton:

src/quacc/calculators/espresso/espresso.py
```diff
@@ -298,7 +298,10 @@ def __init__(
         self._user_calc_params = {}
 
         template = template or EspressoTemplate("pw")
-        self._bin_path = str(SETTINGS.ESPRESSO_BIN_PATHS[template.binary])
+        full_path = Path(
+            SETTINGS.ESPRESSO_BIN_DIR, SETTINGS.ESPRESSO_BIN_PATHS[template.binary]
+        )
```
Andrew-S-Rosen, line 303: This seems to conflict with the `ESPRESSO_BIN_PATHS` description, which implies that it's the full path to each executable. Perhaps some further clarification is in order.

src/quacc/settings.py
```diff
@@ -126,6 +126,9 @@ class QuaccSettings(BaseSettings):
     # ---------------------------
     # ESPRESSO Settings
     # ---------------------------
+    ESPRESSO_BIN_DIR: Path = Field(
+        Path(), description="Base path to the espresso binaries."
```
Andrew-S-Rosen, line 130: Please use `Path.cwd()` to get the absolute path.

src/quacc/calculators/espresso/espresso.py
tomdemeyere: ESPRESSO_BINS or ESPRESSO_BIN_NAMES ? I am open to other suggestions

src/quacc/settings.py
tomdemeyere: For now, it wouldn't work to put Path.cwd() for the default case we want it to be 'pw.x' (assuming it is in the path) not /some/random/path/cwd/pw.x ?

src/quacc/calculators/espresso/espresso.py
Andrew-S-Rosen: Depends on your intention. I think originally you wanted to have it be possible for every executable to have a custom path. That was the motivation for having a single `ESPRESSO_BIN_PATHS`. But if you think that use case is highly unlikely, then a `ESPRESSO_BINARIES` would probably be fine.

tomdemeyere: I would still like to be able to have a different path for each binary although I would like it to be compatible with the config file and CLI: quacc set ESPRESSO_BIN_PATHS['pw'] = ... but I guess that's out of reach

src/quacc/settings.py
Andrew-S-Rosen: Apologies. You are absolutely correct.

Maybe it would be useful to have the list of binaries be strings and not Paths to prevent confusion?

src/quacc/calculators/espresso/espresso.py
Andrew-S-Rosen: There's definitely a way to do that kind of thing with `typer` (albeit likely with different syntax) but yeah, not implemented currently.

PR #1530, opened 2024-01-13
Adjustment of how Espresso handles parallelization flags

Andrew-S-Rosen:
@tomdemeyere: I assume the reason that you have included a `parallel_info` in each `@job` is so that you have full control over the parallelization command for each `@job` in a workflow. Is that correct? Mostly curious what the thought process was for the `parallel_info` decision as opposed to something like that in this PR, and I imagine that must be the reason.

The `ESPRESSO_PARALLEL_CMD` in this PR can be modified on-the-fly [with an environment variable](https://quantum-accelerators.github.io/quacc/user/settings/settings.html#using-a-yaml-file), much like the historical ASE way. Most workflow engines give you the option to specify the environment variables in the decorator parameters. Of course though, it isn't the same as a keyword argument.

Note: This PR was to more easily express my question; not for merge. I haven't tried the whole `parallel_info` stuff in the `GenericFileIOCalculator`s yet, and I'm still trying to wrap my head around it......

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.1]
+
+### Fixed
+
+- Fixed autoamtic k-point schemes in VASP from return `kpts` as `list[float]` instead of `list[int]`
+
 ## [0.6.0]
 
 ### Added
--- a/docs/user/wflow_engine/executors2.md
+++ b/docs/user/wflow_engine/executors2.md
@@ -64,6 +64,12 @@ If you haven't done so already:
     quacc set WORKFLOW_ENGINE prefect && quacc set CREATE_UNIQUE_DIR True
     ```
 
+    Also make sure to connect to Prefect Cloud if you are not self-hosting:
+
+    ```bash
+    prefect cloud login
+    ```
+
 === "Jobflow"
 
     On both the local and remote machines:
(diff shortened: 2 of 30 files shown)


Andrew-S-Rosen:
Ah, I see now how the `parallel_info` stuff works. Not super intuitive from ASE's side, but... 🤷

tomdemeyere:
I completely agree with these changes btw. At least until ASE produce something convincing.

I would even go further by adding it as a parameter so that users can change it dynamically in a python script (to be able to change espresso's parallelisation parameters nk, nb, ... as it will very often be needed). But a setting is good enough

Andrew-S-Rosen:
@tomdemeyere Thanks for your input! These upstream changes to parallelization have been very frustrating and counter-intuitive. It looks like my attempt to use an `argv` here (as mentioned in the comment at the top of the ASE `Espresso` calculator) doesn't work as intended since there is no `argv` keyword argument (anymore?). I opened an upstream issue here: https://gitlab.com/ase/ase/-/issues/1415. I'll revisit this once those details are sorted out...

tomdemeyere:
It was changed to “binary” and “pseudo”_path”. Also binary has to be a string (without any space) not a list of string

I can take care of that if needed

Andrew-S-Rosen:
If it's just the binary and the pseudo path, then where does one put the parallelization flags? (In my discussions with Ask, it sounded like each calculator would have an argv to override this config stuff. He wants the binary to just be the binary --- no other commands.)

tomdemeyere:
In parallel_info, something like:

``` python

profile = EspressoProfile(binary='/some/path/pw.x', pseudo_path = '...')

parallel_info = {
     'binary' : 'srun',
     '-v' : True,
     '-n' : 40
}

Espresso(profile = profile, parallel_info = parallel_info)

Would give: 'srun -v -n 40 /some/path/pw.x'

Andrew-S-Rosen:
Closing this. Will re-open if/when we get some more sensible approach upstream in ASE......

PR #1531, opened 2024-01-13
Proposal/Attempt: supercharge external optimizers to allow running additional actions.

tomdemeyere:
### Checklist

### Notes

## Summary of Changes

This is far from working, it is just to show the global idea. Being able to perform custom actions during dynamics would allow to:

- Generate all kind of additional information when running NEB/MD which is very often necessary.
- With some small additional tuning that user can do in quacc.utils.opt: Perform (hyper)active learning when doing MD/MC and more generally stop/restart the dynamics based on different criteria.

This is the least-intrusive way I found to allow that, I am still very naive on the workflow side so It might not even be possible?

### DIFF
--- a/src/quacc/calculators/espresso/opt_patchs.py
+++ b/src/quacc/calculators/espresso/opt_patchs.py
@@ -20,9 +20,7 @@ class PostProcessingPatch:
         post_processing_job_params = self._calc_kwargs.get("post_processing_params", {})
         forced_params = {
             "input_data": {
-                "input_pp": {
-                    "fileout": f"pseudo_charge_density_{self.nsteps}.cube"
-                }
+                "plot": {"fileout": f"pseudo_charge_density_{self.nsteps}.cube"}
             }
         }
         self._calc_kwargs["post_processing_params"] = recursive_dict_merge(
--- a/src/quacc/calculators/espresso/opt_patchs.py
+++ b/src/quacc/calculators/espresso/opt_patchs.py
@@ -1,0 +1,43 @@
+from pathlib import Path
+
+from quacc.recipes.espresso.core import post_processing_job
+
+
+class PostProcessingPatch:
+    job_names = ["post_processing"]
+
+    def __init__(self, **calc_kwargs):
+        self._results = {job_name: {} for job_name in self.job_names}
+        self._calc_kwargs = calc_kwargs
+
+    def step(self, *args, **kwargs):
+        super().step(*args, **kwargs)
+        # Would this cause issues because everything is ran asynchrously?
+        # If the opt is fast compared to the post-processing, file could be
+        # overwritten while being copied or something like that?
+        # How to redecorate here?
+        self._results["post_processing"][self.nsteps] = post_processing_job(
+            prev_dir=Path().cwd(), **self._calc_kwargs.get("post_processing_params", {})
+        )
+
+
+class SchemaPatch:
+
+    """Base class to patch an ASE optimizer. By modifying methods below,
+    you can 'patch' an optimizer to perform custom actions at specific moment of
+    the simulation"""
+
+    job_names = []
+
+    def __init__(self, **calc_kwargs):
+        self._results = {job_name: {} for job_name in self.job_names}
+        self._calc_kwargs = calc_kwargs
+
+    def step(self, *args, **kwargs):
+        super().step(*args, **kwargs)
+
+    def run(self, *args, **kwargs):
+        super().run(*args, **kwargs)
+
+    def converged(self, *args, **kwargs):
+        super().converged(*args, **kwargs)
--- a/src/quacc/calculators/espresso/opt_patchs.py
+++ b/src/quacc/calculators/espresso/opt_patchs.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 
 from quacc.recipes.espresso.core import post_processing_job
+from quacc.utils.dicts import recursive_dict_merge
 
 
 class PostProcessingPatch:
@@ -16,6 +17,17 @@ class PostProcessingPatch:
         # If the opt is fast compared to the post-processing, file could be
         # overwritten while being copied or something like that?
         # How to redecorate here?
+        post_processing_job_params = self._calc_kwargs.get("post_processing_params", {})
+        forced_params = {
+            "input_data": {
+                "input_pp": {
+                    "fileout": f"pseudo_charge_density_{self.nsteps}.cube"
+                }
+            }
+        }
+        self._calc_kwargs["post_processing_params"] = recursive_dict_merge(
+            post_processing_job_params, forced_params
+        )
         self._results["post_processing"][self.nsteps] = post_processing_job(
             prev_dir=Path().cwd(), **self._calc_kwargs.get("post_processing_params", {})
         )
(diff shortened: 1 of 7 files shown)


buildbot-princeton:

Andrew-S-Rosen:
This is a very interesting proposal. Particularly your second point has me interested. It is similar in spirit to what we do on the Materials Project when running VASP calculations with [Custodian](https://github.com/materialsproject/custodian) but is instead here it is for ASE dynamics routines. 

I must admit that I think I'd need to see a minimal working example to understand a bit of the logistics here and to provide feedback, but the general concept seems okay. Regarding your question about the workflow stuff, I guess it depends on what you plan to happen. If you stop a run due to some criteria, so long as an error code is not raised to the parent process and it's treated like any other run that went to completion, there should be no issues. Similarly for restarts --- if that's happening all behind-the-scenes, that should be okay. 

You could potentially look at the `TrajectoryObserver` class in the `matcalc` repository [here](https://github.com/materialsvirtuallab/matcalc/blob/6338d7755466d707c7b071dab9f4c7b331bf4ca9/matcalc/relaxation.py#L24) in case there are any similarities with what you're trying to accomplish.

tomdemeyere:
Here you go! I produced a toy example: I patched a ASE optimizer (LBFGS, but it does not matter it should work with every class that is based on the "Dynamics" class) using what is in quacc.calculators.espresso.opt_patchs. The patched optimizer additionally computes the pseudo-charge density using the post_processing_job in the espresso recipes. What's the point? For example, people working with organics might be interested in computing the change in DORI during an MD/TS/OPT.

Eventually, this can be very powerful, any kind of workflow in the literature that involves tasks between steps will eventually be reproducible using this, with any calculator or even a mix of them. Without saying that it will be immediately deployable on large-scale HPC using the workflow engines.

If you decide to introduce this, this would need to be placed in the "advanced" part of the documentation since users would probably need to construct their path for their own needs. This requires knowledge about how classes work etc... Given enough examples for each calculator intermediate users might get a grasp of it.

src/quacc/calculators/espresso/opt_patchs.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, **calc_kwargs):
+        self._results = {job_name: {} for job_name in self.job_names}
+        self._calc_kwargs = calc_kwargs
+
+    def step(self, *args, **kwargs):
+        super().step(*args, **kwargs)
+        # Would this cause issues because everything is ran asynchrously?
+        # If the opt is fast compared to the post-processing, file could be
+        # overwritten while being copied or something like that?
```
tomdemeyere, line 18: If my understanding is correct here post_processing will be able to ran concurrently. If the simulation is fast enough, files could changed while being copied?

src/quacc/recipes/espresso/core.py
```diff
@@ -148,6 +152,91 @@ def relax_job(
     )
 
 
+@job
+def external_relax_job(
+    atoms: Atoms | UnitCellFilter,
```
Andrew-S-Rosen, line 157: No need to have this. The `run_opt()` method takes an optional parameter `relax_cell`, which automatically applies the `FrechetCellFilter` to enable volume relaxations (the `FrechetCellFilter` is the improved `ExpCellFilter` which itself was the improved `UnitCellFilter`).

What this means is you can mimic what was done in the [EMT relax_job](https://github.com/Quantum-Accelerators/quacc/blob/85bb036066ff217e013ae04f9948c0c3d37b3e4c/src/quacc/recipes/emt/core.py#L57) by adding a `relax_cell: bool = False` keyword argument to the job here. Then pass that `relax_cell` value to the `base_opt_fn()`.

src/quacc/recipes/espresso/core.py
```diff
@@ -148,6 +152,91 @@ def relax_job(
...
+    RunSchema
+        Dictionary of results from [quacc.schemas.ase.summarize_run][].
+        See the type-hint for the data structure.
+    """
+
+    calc_defaults = {
+        "input_data": {
+            "control": {
+                "calculation": "scf",
+                "tstress": True if isinstance(atoms, UnitCellFilter) else False,
```
Andrew-S-Rosen, line 212: With the prior comment, you'd then be able to do `True if relax_cell else False`.

src/quacc/recipes/espresso/core.py
```diff
@@ -148,6 +152,91 @@ def relax_job(
...
+                "tprnfor": True,
+            }
+        }
+    }
+
+    opt_defaults = {
+        "fmax": 0.01,
+        "max_steps": 500,
+        # LBFGS is very often the "best" optimizer for the general case.
+        "optimizer": LBFGS
```
Andrew-S-Rosen, line 222: Elsewhere I had been using max_steps of 1000 and FIRE for the optimizer. But I don't think it really matters much in practice.

src/quacc/recipes/espresso/core.py
```diff
@@ -148,6 +152,91 @@ def relax_job(
     )
 
 
+@job
+def external_relax_job(
```
Andrew-S-Rosen, line 156: Elsewhere, I had been calling this `ase_relax_job` but perhaps your naming is clearer. I'll let you decide.

src/quacc/runners/ase.py
```diff
@@ -116,6 +116,7 @@ def run_opt(
     fmax: float = 0.01,
     max_steps: int = 500,
     optimizer: Optimizer = FIRE,
+    optimizer_patch: Any | None = None,
```
Andrew-S-Rosen, line 119: Maybe we could make some sort of base class for this kind of patch.

src/quacc/runners/ase.py
```diff
@@ -194,6 +195,14 @@ def run_opt(
 
     # Run calculation
     with traj, optimizer(atoms, **optimizer_kwargs) as dyn:
+        if optimizer_patch is not None:
+
+            class PatchedOptimizer(optimizer_patch, dyn.__class__):
+                pass
+
+            dyn.__class__ = PatchedOptimizer
+            dyn.__init__(**optimizer_kwargs)
```
Andrew-S-Rosen, line 204: Hmm, why not simply define a custom optimizer in this case then? Why patch an existing optimizer in this way rather than have it be a new `optimizer` class being called above?

src/quacc/calculators/espresso/opt_patchs.py
Andrew-S-Rosen: Indeed that seems like it could be problematic if you are launching separate post processing jobs in parallel to the running optimization. Generally, dealing with file I/O and concurrent simulations is tricky (hence all the custom features you had to add in when making your Espresso recipes).

Andrew-S-Rosen (commented):
Hi @tomdemeyere: Thanks for your proposal! There are a bunch of moving parts here (no pun intended), so it's bit difficult for me to provide a thorough review on this at the moment.

What would be most helpful would be the following:
1. Open a PR for just the normal (non-patched) Espresso relaxation job with ASE optimizers. This is pretty standard and done elsewhere in quacc, as you have already seen.
2. Open a separate PR implementing a very simple, minimal example of the optimizer patching for something like EMT. In general, it is a lot easier to debug, test, and understand examples on a simplistic calculator like EMT or LJ or even something like tblite.

I don't anticipate any issues with the first item above, but it will make reviewing something like this easier. The second item will make it easier for me to follow what the proposal is and how we might want to implement it. Then we can consider adding this kind of feature to Espresso. 

Of course, if merging this into `main` isn't critical, you can go off and happily do your own thing. Since this behavior is advanced, it's important this is done in a way that is as simple as possible, which is a bit difficult for me to gauge with the current PR.

PR #1578, opened 2024-01-20
Cleaner tmpdir handling and setting `CREATE_UNIQUE_DIR` as `True` by default

### LINKED ISSUE #1477: Consider making `CREATE_UNIQUE_DIR` be `true` by default

Right now, `CREATE_UNIQUE_DIR` is set to `false` by default, such that all output files are written to the `RESULTS_DIR` (CWD by default). This is perhaps the most intuitive for users, but it causes file clobbering. It also prevents some workflows like the new Espresso ones from running correctly if they rely on knowledge about the filesystem.

Since most users will eventually need to enable `CREATE_UNIQUE_DIR`, we should consider making it enabled by default. There is, however, one important caveat. Some workflow engines, like Covalent and Jobflow/FireWorks, have their own internal mechanisms for task isolation. When this is the case, `CREATE_UNIQUE_DIR` should not be set to `true`. This is because there would then be unnecessary folder structures and because the UIs for the workflow engines will be pointing to the wrong directory.

Regardless, this would be a breaking change that would need to be carefully considered. I think the pros would outweigh the cons, but it needs to be done carefully with respect to end users.

Andrew-S-Rosen:
1. `CREATE_UNIQUE_DIR` is now set to `True` by default (closes #1477)
2. The tmpdir for actively running jobs will now have the same timestamp as the unique result folder where the data is stored. Before, they differed slightly, which did not make much sense.
2. The tmpdir is named `tmp-quacc-...` instead of `quacc-tmp...` for easier indexing/deleting. Similarly, the "symlink" now prepends the name instead of being placed at the end.


Andrew-S-Rosen:
CC @tomdemeyere.

PR #1602, opened 2024-01-23
Speed up Espresso tests

Andrew-S-Rosen:


Andrew-S-Rosen:
@tomdemeyere --- I moved the grid phonon flows to just the Dask test suite since it runs _so_ much faster that way on GitHub Actions. Of course, if the tests pass via Dask they are almost guaranteed to pass locally too. I guess that's a bonus of workflow engines --- faster tests!

PR #1618, opened 2024-01-25
Espresso: autorestarting in trivial cases

tomdemeyere:
## Summary of Changes

Currently ase_relax_job will ask QE to compute the wavefunctions/charge-density from scratch at each step, this is highly inefficient.

If a template called Profile.run more than once and the autorestart is activated (default False, but True when calling ase_relax_job) then we turn off QE restart keyword.

### Checklist

### Notes

### DIFF
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -112,7 +112,8 @@ class EspressoTemplate(EspressoTemplate_):
 
         if self.binary == "pw":
             if self.autorestart and self.nruns > 0:
-                parameters["input_data"]["control"]["restart_mode"] = "restart"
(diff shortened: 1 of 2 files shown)


tomdemeyere:
Note: this should maybe moved to clean_params instead of write_input?

Andrew-S-Rosen:
Thanks, @tomdemeyere. This makes a lot of sense, and your approach is quite smart.

Re: its placement, ultimately, I will leave that up to you. I personally like to keep the main calculator interface as minimal as possible, relying on calling of functions in other modules but whether that makes sense or not depends on personal opinion and how many attributes in `self` need to be passed. Just let me know.

tomdemeyere:
Actually, this can't be placed in _clean_params, this function is only called at the beginning, we need something at each write_input.

It's good to go for me

Issue #1614, opened 2024-01-24
Proposal: add xml reader for espresso


tomdemeyere:
### What new feature would you like to see?

This would be a very simple xml reader that does nothing except parse the xml. This is nice because:

1. This gives **all** information about a run, this would be stored in the results dict, as a dict.
2. This works with multiple binaries (pw.x, ph.x, neb.x).
3. This can be standardised if other codes use xml to store all the data about a run (which they should do)

One problem: This needs "import xml.etree.ElementTree" which is known for vulnerabilities: https://docs.python.org/3/library/xml.html#xml-vulnerabilities

This should not be very problematic since we are not reading arbitrary xml files, but I would rather have your opinion before starting to work on it.

Andrew-S-Rosen:
Thanks for pinging about this. Personally, I have no issues with it.

Nekkrad:
@Andrew-S-Rosen @tomdemeyere  What about this one?

https://github.com/QEF/qeschema

Andrew-S-Rosen:
That seems very reasonable to me. It's on the official QEF repo (big win), it's reasonably actively maintained, it's pip installable, and it's trivial to add as an optional dependency to quacc. I have no concerns.

tomdemeyere:
Wow, nice find, I didn't know about it!

PR #1643, opened 2024-01-31
added pdos job and flow

Nekkrad:
## Summary of Changes
Implemented a pdos job and a pdos flow (use of projwfc.x binary). 

There are multiple files created and in theory I could create custom functions to read all the relevant information and store them in the result dictionary. Another option is to read the xml file via either a big custom function or the use of xml_to_dict python package but that means we have to add a new dependency.

### Checklist

### DIFF
--- a/docs/user/recipes/recipes_list.md
+++ b/docs/user/recipes/recipes_list.md
@@ -199,6 +199,8 @@ The list of available quacc recipes is shown below. The "Req'd Extras" column sp
 | Espresso Non-SCF         | `#!Python @job`  | [quacc.recipes.espresso.core.non_scf_job][]         |              |
 | Espresso DOS             | `#!Python @job`  | [quacc.recipes.espresso.dos.dos_job][]              |              |
 | Espresso DOS Flow        | `#!Python @flow` | [quacc.recipes.espresso.dos.dos_flow][]             |              |
+| Espresso Projwfc         | `#!Python @job`  | [quacc.recipes.espresso.dos.projwfc_job][]          |              |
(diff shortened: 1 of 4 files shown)


buildbot-princeton:

Nekkrad:
1. It seems that projwfc.x has the same problem as dos.x which causes the test to fail. I have more recipes ready for the other binaries (bands.x and  fermi surface). Should I still open new PRs or should we try and see if we can get the binaries to work

2. Puzzled about the DeepSource problem. I understand what it says but it works fine. Should I ignore it?

Andrew-S-Rosen:
New dependencies are okay because we can make the optional (https://github.com/Quantum-Accelerators/quacc/issues/1614). That said, my preference would be for dependencies that are actively maintained (xml_to_dict hasn't had a commit in 10 months).

No worries about the deepsource issue. It's a bug I've reported to their team. I'll disable that error message over the repo later.

Regarding the binary, that's odd. I personally don't want to merge in code without passing tests, so we will probably need to either: figure out how to solve it, work around it, or (if we absolutely must) mock the Espresso call.

Nekkrad:
> Regarding the binary, that's odd. I personally don't want to merge in code without passing tests, so we will probably need to either: figure out how to solve it, work around it, or (if we absolutely must) mock the Espresso call.

So the error seems to be related to some missing library (as in dos.x ) . I will try to conda install qe instead of my compiled one to see what happens.

tomdemeyere:
Can you give here the result of `ldd $(which pw.x)` ?

Nekkrad:
> Can you give here the result of `ldd $(which d

PR #1672, opened 2024-02-07
Implement MD workflows

### LINKED ISSUE #1577: Add some general, ASE-based MD compute jobs

It would be nice to add in some MD workflows, ideally in [`quacc/recipes/common`](https://github.com/Quantum-Accelerators/quacc/tree/main/src/quacc/recipes) so the various codes could call it. This would be based around ASE. It isn't difficult to do, but I'm not an MD expert so don't know the ideal defaults and such to pick.

tomdemeyere:
## Summary of Changes

Closes #1577.

Before going further a little consultation:

MD users use workflows all the time, just for the NVE example in this commit there is pretty much a thousand ways you can run that; doing a first NVT to relax to the desired temperature, or a first NPT to relax to the desired density, etc...

1. For example, should we have one NVT workflow with a parameter "thermostat", or one workflow per thermostat?
2. If you want things to be in recipe.common, atoms will have to be sent in with a calculator already attached, is that what you want?
3. From what I understand you want to keep things simple (not something like the code here), users then build their workflows to get what they want.
4. In any case, for now doing these kinds of workflows is difficult, because it is generally hard to transfer results between ASE calculators, which will always happen between runs, this is critical since MD needs momenta and forces to ensure proper continuation. (This is a problem I would like fixed above since it also causes problems when restarting optimization etc)

### Checklist


buildbot-princeton:

Andrew-S-Rosen:
Happy to respond but before I do, I just wanted to check that this is the full commit you wanted to share --- just the args/kwargs, right?

tomdemeyere:
> @tomdemeyere: Happy to respond to each question but before I do, I just wanted to check that this is the full commit you wanted to share --- just the args/kwargs, right?
> 
> 
> 
> (In short, I really would love to have MD flows and am willing to work together to figure out how to get this done the best way possible)

Yes, before doing more I wanted to make sure we are on the same page.

I guess my main question here is: how to make these flows common to all codes? Should they call code specific MD jobs? Or, should a calculator be attached to the Atoms object being sent to the MD flow?

Once I know this we can work on details

Andrew-S-Rosen:
> I guess my main question here is: how to make these flows common to all codes? Should they call code specific MD jobs? Or, should a calculator be attached to the Atoms object being sent to the MD flow?

Good question.

Personally, I think what would likely make the most sense is to make a `run_md` function in [`quacc.runners.ase`](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/runners/ase.py) that would take an `Atoms` object with an attached calculator just like the other methods in there. You likely were aware of that already.

From there, we get to the recipes, and my answer is: "it depends." If it is just a simple `@job` (i.e. not a `@flow`), then this is no different than a relaxation job --- one would just be calling `run_md` instead of `run_calc` or `run_ase_opt`. Since there wouldn't be much logic beyond the simple function call, this is fine. In fact, I don't really intend most users to call the common recipes directly. They are there to reduce code duplication, but most users aren't comfortable using them directly. I would treat the individual MD jobs in the same way as a relaxation --- we have a runner, which is called in the various recipes where it is appropriate. For instance, we would make an `qucc.recipes.emt.md.py` and such. It does mean that there would be multiple MD modules, but we would need that anyway because every code will have different parameters (defaults) that are needed to properly run an MD calculation. There are other nuances for why this suggested approach is ideal related to (de)serialization that I won't bore you with (namely, the outermost `@job`/`@flow`, i.e. the one directly called by the user, must have arguments that are all pickle-able, and keyword arguments must be JSON serializable for the various workflow engines to function properly --- and the current code layout permits this).

If we are talking about a `@flow` where there are multiple `Job`s being stitched together, and we anticipate this pattern (i.e. the workflow DAG) to be calculator-agnostic, then we'll put that in `common`. This is the philosophy behind the existing `common` recipes.

To get started, I would suggest making `quacc.runners.ase.run_md` and then a simple demo for a cheap-to-run calculator, like EMT or LJ. That will likely be illustrative and help guide the design further.

> For example, should we have one NVT workflow with a parameter "thermostat", or one workflow per thermostat?

If the workflow logic is staying the same regardless of thermostat, then I would suggest one job/workflow where thermostat would be a keyword argument with a sensible default.

> If you want things to be in recipe.common, atoms will have to be sent in with a calculator already attached, is that what you want?

See comment above. If we are talking about a `@job`, then that would call `run_md` and be specific for the code since we need to set some sensible default parameters. If it's a `@flow` pattern, that can go in `common` but would not require passing in `Atoms` objects with attached calculators --- it would require passing in one or more `Job`s. 

> From what I understand you want to keep things simple (not something like the code here), users then build their workflows to get what they want.

Hard for me to tell from the commit you shared here, but I am not opposed to complexity. In fact, we should not rely on the users to build something complex. If the `@flow` pattern is "trivial", like a relax then an MD run, it is debatable if we should make that its own `@flow` or not (honestly, it can go either way). But if it's more complex, for sure we should. 

> In any case, for now doing these kinds of workflows is difficult, because it is generally hard to transfer results between ASE calculators, which will always happen between runs, this is critical since MD needs momenta and forces to ensure proper continuation. (This is a problem I would like fixed above since it also causes problems when restarting optimization etc)

This is a good point. However, I think it is solvable. Let's circle back to this when the time is right.

tomdemeyere:
@Andrew-S-Rosen Thanks for all these details. I will come back to it at some point

tomdemeyere:
@Andrew-S-Rosen I think this is now a good point to have your input

src/quacc/recipes/emt/core.py
Andrew-S-Rosen, line 1: Maybe we make a dedicated `md.py` module for such things.

Andrew-S-Rosen (commented):
Thank you for kicking this off, @tomdemeyere! Really appreciate the contribution!

This is looking good. I have some comments that are mostly related to some minor restructuring.

Starting with this isolated EMT example is very helpful for me in understanding the process and figuring out what should be refactored, so I thank you for doing that.

________________

Once the comments are taken care of, we can think about step 2, which is how to most readily get this `microcanonical_job` in other codes. Presumably we'd like to add support for an analogous `microcanonical_job` in several recipes, and there is a bit more to this one than a typical `run_calc` kind of thing. I would be okay with taking the bulk of this function and making a `microcanonical_fn` function in `recipes.common._base.py` that takes an `Atoms` object with attached `Calculator` as input like you suggested originally (or as separated arguments; it doesn't matter). We would still keep `microcanonical_job` in the EMT recipes, but it would only have one or two lines to it and would call `microcanonical_fn`. Note that `microcanonical_fn` is a pure function, not a `@job`. The `@job` decorator would still go on `microcanonical_job`. You wouldn't have to change your tests. The goal is to make it trivial to add `microcanical_job` for any other recipe later on. Thoughts?

I can take care of the above since it might be worth doing something analogous for the `relax_job`s and I want to make sure things are treated the same across calculators.

tomdemeyere:
> Thank you for kicking this off, @tomdemeyere! Really appreciate the contribution!
> 
> 
> 
> This is looking good. I have some comments that are mostly related to some minor restructuring.
> 
> 
> 
> Starting with this isolated EMT example is very helpful for me in understanding the process and figuring out what should be refactored, so I thank you for doing that.
> 
> 

I see that your comment is different than the one I received by email from git. But yes, originally my plan was to have these "_base" functions. Either per calculator or in common like you previously said. 

After discussing that along with other details I was then planning to introduce recipes for the other ensembles (NVT, NPT).

Andrew-S-Rosen:
> I see that your comment is different than the one I received by email from git. But yes, originally my plan was to have these "_base" functions. Either per calculator or in common like you previously said.

Apologies about the edits! I was trying to sort out where to place them and was going back-and-forth so decided to just save it for a future discussion to avoid confusion. But you got the right idea anyway. 

> After discussing that along with other details I was then planning to introduce recipes for the other ensembles (NVT, NPT).

That would be great!

> Didn't see your comments originally (github mobile) will work on it at some point

Thank you!

src/quacc/recipes/emt/core.py
```diff
@@ -89,3 +95,72 @@ def relax_job(
...
+
+    md_defaults = {"timestep": 1.0, "steps": 500, "dynamics": VelocityVerlet}
+
+    initial_temperature_defaults = {"temperature": None, "fixcm": True, "fixrot": True}
+
+    initial_temperature_params = recursive_dict_merge(
+        initial_temperature_defaults, initial_temperature_params
+    )
+
+    temperature = initial_temperature_params.pop("temperature", None)
```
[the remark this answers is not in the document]
tomdemeyere: In NVE temperature is not well defined, usually what is done is that initial velocities are initialised to a given temperature via a distribution. That is why I didn't put "temperature" directly in the parameter since it could be confusing: "temperature for a micro-canonical job, what is that???"

Same for the comment below, fix_rot or fix_cm are only for the initialisation phase, if the users wants proper COM fixing they should pass an Atoms object with a FixCom constraint.

That's why I put everything in an "initial..." dict, so that the users are clearly aware that this is limited to the initialisation phase. I am open to changing things, just some additional info here

src/quacc/recipes/emt/core.py
```diff
@@ -89,3 +95,72 @@ def relax_job(
...
+    fixcm = initial_temperature_params.pop("fixcm", False)
+    fixrot = initial_temperature_params.pop("fixrot", False)
+
+    if temperature:
```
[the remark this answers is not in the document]
tomdemeyere: If temperature = 0, everything works as expected since MaxwellBoltzmann should give no velocities, which is what happens here.

src/quacc/recipes/emt/core.py
```diff
@@ -89,3 +95,72 @@ def relax_job(
...
+    atoms
+        Atoms object
+    initial_temperature_params
+        Dictionary of custom kwargs for the initial temperature distribution. Set a value
+        to `quacc.Remove` to remove a pre-existing key entirely. For a list of available
+        keys, refer to [ase.md.velocitydistribution.MaxwellBoltzmannDistribution][].
+        Quacc has two additional parameters `fixcm` and `fixrot` to fix the center of mass
+        and rotation, respectively when setting the initial temperature. The default is
+        `{"temperature": None, "fixcm": True, "fixrot": True}`. If `temperature` is set to
+        `None`, the initial temperature will not be set. For an improved change of success when performing a microcanonical ensemble calculation, the initial temperature should be set to a value twice the
```
Andrew-S-Rosen, line 121: "change" --> "chance"

src/quacc/recipes/emt/core.py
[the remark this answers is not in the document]
tomdemeyere: >> Are these initial_temperature_params going to be included for all other ensembles as well

Well, for NVT or NPT, temperature plays a key a role, we will have a `temperature` keyword alone. But initial_temperature_params will have to be here as well. Usually the initial temperature should be equal to the one you aim, but everything is possible really.

I thought about adding these parameters individually as you suggest, but ultimately I found that having 4 different only for the initialisation phase is not ideal. I don't have a strong opinion here, both approach seems ok to me, with a slight preference to the dict form.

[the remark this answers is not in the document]
Andrew-S-Rosen: Let's stick with the `dict` form then :)

src/quacc/recipes/emt/core.py
```diff
@@ -89,3 +95,72 @@ def relax_job(
...
+
+    initial_temperature_defaults = {"temperature": None, "fixcm": True, "fixrot": True}
+
+    initial_temperature_params = recursive_dict_merge(
+        initial_temperature_defaults, initial_temperature_params
+    )
+
+    temperature = initial_temperature_params.pop("temperature", None)
+    fixcm = initial_temperature_params.pop("fixcm", False)
+    fixrot = initial_temperature_params.pop("fixrot", False)
```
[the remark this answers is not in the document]
tomdemeyere: BTW, these are not class, these are functions (ZeroRotation and Stationary) for some reason python nomenclature is not respected here. Might be worth checking on ASE side?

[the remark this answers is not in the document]
Andrew-S-Rosen: 🫠 Thanks for the headd's up.

tomdemeyere:
Few comments:

- Without run_kwargs there is no way to pass things to `dyn.run()`
- There is still a check_convergence, although it does not make much sense for MD. By definition you reached the max number of steps, **but** there is a case where it makes sense:

We are currently working on a `error-free` quacc as part of another project. We make sure that quacc always reaches the point of writing to the database: it writes the errors that occurred, along with the available results, if any (the errors might happen at some point during the MD, previous results should be returned, actually in some cases, errors might even be part of your journey (although this might not be what Quacc was originally made for)). Currently if one is running a bunch of simulations, if errors occur, there will be a bunch of directories with no really easy way to figure things out.

Because you said you might be interested in error handling I am leaving that there.

- I change how temperature/time etc... is handled, feel free to tell me what you think

Andrew-S-Rosen:
> Without run_kwargs there is no way to pass things to dyn.run()

The `fmax` and `steps` are already passed in `quacc.runners.ase.run_opt`, and `.run()` doesn't take any other keyword arguments, so that should be fine.

> There is still a check_convergence, although it does not make much sense for MD. By definition you reached the max number of steps, but there is a case where it makes sense:

I agree that a `check_convergence` doesn't make much sense for this job type. We should remove that logic entirely.

> We are currently working on a error-free quacc as part of another project. We make sure that quacc always reaches the point of writing to the database: it writes the errors that occurred, along with the available results,

There is a global setting [CHECK_CONVERGENCE](https://github.com/Quantum-Accelerators/quacc/blob/1d66b3d04c557f8c1d2a69e454f81304380747cf/src/quacc/settings.py#L101-L104). One could catch the typical `RuntimeError` errors when ASE is called and try to give the user some flexibility via the global settings about how to handle things with the `CHECK_CONVERGENCE` or some other global parameter. But of course, feel free to continue doing what you're doing!

> Currently if one is running a bunch of simulations, if errors occur, there will be a bunch of directories with no easy way to figure things out.

Indeed. This is a limitation of Parsl, which is really best thought of a task manager. Other true workflow tools like Prefect or Covalent have a separate database for the task metadata and would highlight when/where the job fails. But since Parsl does not have a task database, there is no mechanism to easily store that info. The recommended approach would be to launch your calculations in a `for` loop and log any errors from the Parsl `AppFuture`. But of course, your mechanism is fine as well.

> I changed how temperature/time etc... is written to results, feel free to tell me what you think

I like that a lot more!

Andrew-S-Rosen:
@tomdemeyere Thank you for addressing my comments! This looks good to me. How are you feeling about the state of this PR? 

I will likely go ahead and refactor `summarize_md_run` and `run_md` afterwards since there is some copy/paste between methods that we can likely reduce. But that's something I'm happy to sort out.

tomdemeyere:
I just changed the whole microcanonical thing to "md_job" because there will not be any major differences between these jobs except from Quacc point of view.

ASE has a lot of deprecated keywords, such as "temperature" or "pressure" which should be written as "temperature_K" or "pressure_au". Except... for some MD classes where they are perfectly fine 👌. Anyway, what I do here is: I change everything to what is normally expected, In Quacc if the user input "temperature" or "temperature_K" they will both work in Kelvin (A message will be printed in the first case). Feel free to change this behaviour if you want.

1. Momenta (results in general) has to be carried from job to job, this is a hard requirement, without this, flows do not make sense for MD jobs. For now, when setting a new calculator in the next job previous results will be erased.

2. fixcm and fixrot have been changed to fix_com and fix_rot, which I prefer as well, but fixcm and fixrot from ASE are still a thing in some ASE MD classes and might confuse people.

tests/core/recipes/emt_recipes/test_emt_recipes.py
```diff
@@ -1,10 +1,14 @@
+from __future__ import annotations
```
Andrew-S-Rosen, line 1: We should avoid `from __future__ import annotations` in the test suite.

src/quacc/runners/ase.py
```diff
@@ -240,6 +246,132 @@ def run_opt(
...
+            dynamics_kwargs.pop("compressibility") / GPa
+        )
+
+    traj_filename = tmpdir / "opt.traj"
+    traj = Trajectory(traj_filename, "w", atoms=atoms)
+    dynamics_kwargs["trajectory"] = traj
+
+    # Run calculation
+    with traj, dynamics(atoms, timestep=timestep * fs, **dynamics_kwargs) as dyn:
+        dyn.run(steps=steps)
```
Andrew-S-Rosen, line 364: Do you think it makes sense to add an option that would save the files from every step, e.g. for training ML models later? I did that recently for the optimization runner [here](https://github.com/Quantum-Accelerators/quacc/blob/a13538a7df5628117442c0745637bbe10e0c7aa5/src/quacc/runners/ase.py#L224-L238). I can add this to the MD run, but I am just curious about your opinion.

src/quacc/runners/ase.py
```diff
@@ -240,6 +246,132 @@ def run_opt(
...
+        raise ValueError(msg)
+
+    if "temperature" in dynamics_kwargs or "temp" in dynamics_kwargs:
+        LOGGER.warning(
+            r"The `temperature`\`temp` kwargs are ASE deprecated and will"
+            "be interpreted as `temperature_K` in Quacc."
+        )
+        dynamics_kwargs["temperature_K"] = dynamics_kwargs.pop(
+            "temperature", None
+        ) or dynamics_kwargs.pop("temp", None)
```
Andrew-S-Rosen, line 310: All this parameter checking stuff should probably be moved to its own internal function (e.g. `def _md_param_handler()`) so the main logic of the `run_md` function is easier to read.

src/quacc/runners/ase.py
```diff
@@ -240,6 +246,132 @@ def run_opt(
...
+
+    if "pressure" in dynamics_kwargs:
+        LOGGER.warning(
```
Andrew-S-Rosen, line 313: I am happy to do this, but in case you're curious, here is how you write a test for a LOGGER:


```python
LOGGER = logging.getLogger(__name__)
LOGGER.propagate = True

def my_test(caplog):
    with caplog.at_level(logging.WARNING):
        function_that_triggers_warning(arg1, arg2, ...)
```

Andrew-S-Rosen (commented):
Thank you for your thoughtful edits and comments, @tomdemeyere! I left a few comments of my own based on your updates. They are all minor. Of course, I can help address any remaining comments/concerns as desired (note: I have been traveling since Feb. 13th and will be back on Feb. 26th, at which point I can help out more directly if necessary).

> I just changed the whole microcanonical thing to "md_job" because there will not be any major differences between these jobs except from Quacc point of view.

I'm not 100% sure I follow. Is it simply because there would only be a change in defaults/input parameters? Maybe this will become clearer to me once we go through other ensembles later.

> ASE has a lot of deprecated keywords, such as "temperature" or "pressure" which should be written as "temperature_K" or "pressure_au". Except... for some MD classes where they are perfectly fine 👌. Anyway, what I do here is: I change everything to what is normally expected, In Quacc if the user input "temperature" or "temperature_K" they will both work in Kelvin (A message will be printed in the first case). Feel free to change this behaviour if you want.

🫠. Ah, legacy sections of ASE. Anyway, catching common user errors seems wise. Maybe this is something that we (i.e. the ASE community) should streamline sooner rather than later...

> Momenta (results in general) has to be carried from job to job, this is a hard requirement, without this, flows do not make sense for MD jobs. For now, when setting a new calculator in the next job previous results will be erased.

Right, this makes sense. I will have to think more about how to handle this in the best way possible. Presumably, if someone runs the job once, the moment will be stored in `calc.results` and therefore in the output schema. If they want to run the same job again as a continuation or a second job in a flow, they could do `Atoms.set_momenta()` but doing that manually is annoying.

I think would what make the most sense is to mimic what we do right now with magnetic moments. If you noticed, there is a `move_magmoms: bool = True` keyword argument in the various summarizers that will, by default, move the final converged magnetic moments to the initial magnetic moments of the returned `Atoms` object so that the `Atoms` object is automatically ready for the next run. We can add a `move_momenta: bool = True` keyword argument to `summarize_md_run` that does the same thing for momenta. Then there is no need for anyone to think about it at all. It should "just work".

> fixcm and fixrot have been changed to fix_com and fix_rot, which I prefer as well, but fixcm and fixrot from ASE are still a thing in some ASE MD classes and might confuse people.

🫠 again. If it might get confusing, perhaps it makes sense to stick with what's in ASE (if it's consistent). I wasn't aware this was coming from ASE land. Up to you in the end.

tests/core/recipes/emt_recipes/test_emt_recipes.py
tomdemeyere: Yes, for some reason ruff --fix put them here automatically, each time I try to be careful to delete them manually, this one escaped!

src/quacc/runners/ase.py
tomdemeyere: Yes, I wanted to do something like that at some point. I think it's a good idea, although you might want to prepare a big scratch space for this 😅

That's also what I had in mind when thinking about the "optimizer patching". I presented it as a very general concept but in practice this could be used to do something like that. A "FileSave" patch that you would send to your ase_md_job. This patch would then have parameters to allow more flexibility like "save_interval" etc... Usually it would allow to save files based on some other criteria (geometric descriptors, ML auto-evaluation...)

To be fair for this particular action I think your approach is better.

tests/core/recipes/emt_recipes/test_emt_recipes.py
Andrew-S-Rosen: Ah, that's odd. I thought I had set up `pyproject.toml` to avoid that scenario (certainly, I don't want contributors to have to bother manually removing stuff!). I can double check.

Andrew-S-Rosen:
Note to self: there are some interesting ideas in https://github.com/materialsproject/atomate2/pull/722/files#diff-ae55fbbe68bef6a3e4bd052ea86c473adb44a91b440d2abb63a89ff1be198fe7 by Materials Project colleagues.

tomdemeyere:
> Note to self: there are some interesting ideas in https://github.com/materialsproject/atomate2/pull/722/files#diff-ae55fbbe68bef6a3e4bd052ea86c473adb44a91b440d2abb63a89ff1be198fe7 by Materials Project colleagues.

FYI: I didn't give up on this P:, I have some other things to finish first. Also, in the meantime I want to think about what would be the best way to proceed with this

Andrew-S-Rosen:
All good! I also have this on the backburner to think about some more.

tomdemeyere:
Some progress! I think I would like to take advantage of this merge request to fix a few things on ASE's side. Currently there are multiple problems:

- Some Molecular Dynamics classes simply can't restart correctly (NPT)
- During a restart, one configuration will always be calculated, twice.
- No units convention.

This makes ASE simulations not (or hardly) **reproducible**, which is a big deal, in my opinion. Something that would greatly increase the reproducibility of ASE simulation's, and fix the problems above is to allow a verbose mode, where the state of each object involved is written to the disk in a json format (Calculator, Profile, Template, Dynamics) if applicable.

This json format would allow to instantly recreate the objects as well. What do you think?

Andrew-S-Rosen:
Thanks for the update! I think improvements on the ASE side are always welcome. I will give this (and your other PRs, e.g. in ASE and here) a more detailed look later this week. Just got back from travel so am a bit behind.

tomdemeyere:
This should be close to go, there is still something essential before being able to run MD correctly: restarts. They are still difficult to do (impossible in some cases), but that's an ASE problem. I suspect this will take ages to fix; for now I deleted everything related to restart while we wait to see what solutions will get accepted upstream (if any).

I looked at automate2, there are indeed some interesting things, like the possibility to specify a temperature/pressure gradient. I will charge back with my idea of "patches" to accomplish that, this would look like:

``` python

steps = np.arange(1000, 10000, 1)
temperatures = function_of_steps(steps)

my_temp_gradient = TemperatureGradient(steps, temperatures)
my_file_copier = FileCopier(interval = 10)
my_press_gradient = PressureGradient(...)

md_job(... patches = [my_temp_gradient, my_file_copier, my_pressure_gradient])
```

This is achievable by tuning the generator as you did previously (although it might get packed), for some reason I love this "patching" concept 😅. People could fine-tune their MD, picking what triggers when, Quacc could even have a section in the documentation explaining how to do complex things as we already discussed. Anyway, daydreaming here 🙃

Andrew-S-Rosen:
Thanks for your persistence with this! I will try to look at this within the next few days (please ping me if I forget). I'm definitely interested to chat more about the patches idea here; perhaps within the context of MD, it could be an interesting proof-of-concept to think about. I'll revisit this when I have the time to give this a proper review!

tomdemeyere:
Some news on this? Just to know if it is waiting on https://gitlab.com/ase/ase/-/merge_requests/3310

src/quacc/recipes/emt/md.py
```diff
@@ -0,0 +1,123 @@
...
+
+        - Time: femtoseconds (fs)
+        - Pressure: GPa
+        - Temperature: Kelvin (K)
+        - Compressibility: 1/GPa
+
+    !!! Note "Keywords"
+
+        Additionally, Quacc uses the keywords `fix_com` and `fix_rot` instead of
+        `fixcm` and `fixrot` to fix the center of mass and rotation, respectively.
```
Andrew-S-Rosen, line 53: If this is just a matter of renaming and nothing else, let's just stick with the ASE ones. If you approve, let me know and I can revert the change here.

src/quacc/recipes/emt/md.py
```diff
@@ -0,0 +1,123 @@
...
+        As of March 2024, ASE still accept deprecated keywords such as:
+
+        - temperature instead of temperature_K
+        - pressure instead of pressure_au
+        - compressibility instead of compressibility_au
+        - dt instead of timestep
+
+        All of these keywords will be accepted by Quacc, but it is recommended to use the
+        new keywords to avoid confusion. Everything will always be converted to the Quacc
+        units mentioned above.
```
Andrew-S-Rosen, line 64: We probably don't need to mention this since ASE itself will raise a deprecation warning if they are used. Ultimately, it's on the user to use the current ASE names.

src/quacc/runners/ase.py
```diff
@@ -362,3 +418,80 @@ def _set_sella_kwargs(self, optimizer_kwargs: dict[str, Any]) -> None:
...
+                "temperature", None
+            ) or dynamics_kwargs.pop("temp", None)
+
+        if "pressure" in dynamics_kwargs:
+            LOGGER.warning(
+                "In quacc `pressure` and `pressure_au` are equivalent and will be"
+                " interpreted in GPA."
+            )
+            dynamics_kwargs["pressure_au"] = dynamics_kwargs.pop("pressure")
+
```
Andrew-S-Rosen, line 456: Do we need all these checks, or can we just rely on the fact that the user should be using the up-to-date names and if they don't, ASE will warn them or a crash will happen.

Andrew-S-Rosen (commented):
@tomdemeyere: Thank you for your enormous amount of patience with this PR. I didn't see your comment from 3 weeks ago, so my apologies about not responding and for keeping you in the dark about this.

The reason I had been reluctant to merge this was that there was a fair bit of refactoring I wanted to do to remove code duplication in the `run_md` method and also in the schema. This was dependent on some other PRs I had in the pipeline. I was finally able to devote time to do so properly, and now it is much cleaner. 👍 

At this point, I am quite pleased with this. Whenever you have a moment, however, I have a few remaining questions outlined below. I am happy to take over of any (small) changes that result from my questions --- but I wanted to get your professional input before attempting to proceed since I am not an MD expert. Finally, before I merge, if you have a moment to look at a similar attempt recently implemented in [matcalc](https://github.com/materialsvirtuallab/matgl/blob/da3dab9affa09feb391351ab2e9fa38862f2c31d/src/matgl/ext/ase.py#L362), it'd be nice to know if we are missing anything particularly useful that's in there. For instance, there is something about [upper triangular cells](https://github.com/materialsvirtuallab/matgl/blob/da3dab9affa09feb391351ab2e9fa38862f2c31d/src/matgl/ext/ase.py#L576-L577) that is entirely foreign to me.

src/quacc/runners/ase.py
tomdemeyere: The thing is that, parameters are a very **big** mess currently, for example for NVT Berendsen we have:

```
temperature: float
The desired temperature, in Kelvin.

temperature_K: float
Alias for temperature
```
And for :

```
temperature: float (deprecated)
The desired temperature in eV.

temperature_K: float
The desired temperature in K.
```

Sometimes `temperature` is in eV, sometimes in K, you don't know, surprise!

I know it's a lot of time to ask from you but your support on https://gitlab.com/ase/ase/-/merge_requests/3391 (probably not everything in this PR is good to take) would be more than welcome.

If/when this goes, I would rewrite the entire MD doc and we can finally move on with X-years old deprecated keywords and have user-friendliness and consistency.

Andrew-S-Rosen: Ooof, yeah that is a mess! Thanks for the reminder.

Let's see how the discussion plays out and then I am happy to proceed how you suggest. I am happy to review on ASE's side if/when we get a blessing from Ask on certain components.

tomdemeyere:
> @tomdemeyere: Thank you for your enormous amount of patience with this PR. I didn't see your comment from 3 weeks ago, so my apologies about not responding and for keeping you in the dark about this.
> 
> The reason I had been reluctant to merge this was that there was a fair bit of refactoring I wanted to do to remove code duplication in the `run_md` method and also in the schema. This was dependent on some other PRs I had in the pipeline. I was finally able to devote time to do so properly, and now it is much cleaner. 👍 I didn't change any functionality (to the best of my knowledge).
> 
> At this point, I am quite pleased with this. Whenever you have a moment, however, I have a few remaining questions outlined below. I am happy to take over of any (small) changes that result from my questions --- but I wanted to get your professional input before attempting to proceed since I am not an MD expert. Finally, before I merge, if you have a moment to look at a similar attempt recently implemented in [matcalc](https://github.com/materialsvirtuallab/matgl/blob/da3dab9affa09feb391351ab2e9fa38862f2c31d/src/matgl/ext/ase.py#L362), it'd be nice to know if we are missing anything particularly useful that's in there. For instance, there is something about [upper triangular cells](https://github.com/materialsvirtuallab/matgl/blob/da3dab9affa09feb391351ab2e9fa38862f2c31d/src/matgl/ext/ase.py#L576-L577) that is entirely foreign to me.

I had a look at the code you sent me and I am afraid I have never had issues with the upper vs lower triangular matrix (mainly because I don't often run NPT) but from what I understand this is taken care by a PR in ASE (https://gitlab.com/ase/ase/-/merge_requests/3277) by just modifying the NPT class without the need of modifying the atoms object like in matcalc.

For more context: left is my slab before, right is the slab after being passed in matcalc.

image

image

To refresh the current challenges in this PR:

- The lack of proper restart, which breaks the purpose of "MD workflows"
- Choose a convention for the units being used and make sure that it is clear and user-friendly

For the first point, I am afraid this is not for today :/. For the later I tried to make things as clean as I could but of course feel free to discard these suggestions and do as you like.

src/quacc/recipes/emt/md.py
tomdemeyere: Yes sure!

src/quacc/recipes/emt/md.py
tomdemeyere: This can probably be deleted yes

Andrew-S-Rosen:
Thanks for the reply, @tomdemeyere! Regarding the triangular stuff, thanks for linking me to the MR. If this is just an ASE bug/inconvenience, we will take care of that upstream. I'll keep tabs on the MR.

As for this PR, not to worry about restarts. I don't think it is a dealbreaker, and I disagree that it defeats the purpose of a workflow. After all, one might want to do a non-MD job afterwards (e.g. a DFT relaxation) which would be fully supported. It's just that MD-->MD might be tricky. I wouldn't let that hold up this PR. Let me know if/when your upstream ASE MR though needs any attention from me.

The second point is really the only one I feel is necessary to make sure we get right. I trust your opinion on this and am content with leaving it as-is.

So, I think the plan (unless you disagree) should be:
- I'll monitor the conversation at https://gitlab.com/ase/ase/-/merge_requests/3391 to see if there's anything productive that might turn up in the short term.
- Assuming the above answer is no, I'll go ahead and merge this PR pretty much as-is sometime next week.

Thoughts?

Andrew-S-Rosen:
Oh, but also I still had the question about `fixcm` and `fixrot`. Did we need to rename them for some reason?

tomdemeyere:
> Oh, but also I still had the question about fixcm and fixrot. Did we need to rename them for some reason? Was that also an internal consistency thing?

Most of the time you will find `fixcm` but occasionally you might find `fix_com` (mostly internally) I prefer `fix_com` so I used that (it is less obscure in my opinion) but I don't feel strongly about it, change it as you want.

src/quacc/runners/ase.py
Andrew-S-Rosen: Going to postpone changes here for the time being.

src/quacc/runners/ase.py
```diff
@@ -309,6 +326,81 @@ def run_vib(self, vib_kwargs: VibKwargs | None = None) -> Vibrations:
...
 
+    def run_md(
+        self,
+        timestep: float = 1.0,
+        steps: int = 1000,
+        dynamics: MolecularDynamics = VelocityVerlet,
+        dynamics_kwargs: dict[str, Any] | None = None,
+        initial_temperature: float | None = None,
+        fix_com: bool = True,
+        fix_rot: bool = True,
```
Andrew-S-Rosen, line 343: These were set to `True` by default in your commits. Is this standard? If someone sets `initial_temperature`, it's expected that COM and rotation are also fixed? My naive assumption would be that this would be unexpected, but I'm not an MD guy so wanted to check before touching this.

src/quacc/recipes/emt/md.py
```diff
@@ -0,0 +1,69 @@
...
+
+    !!! Important "Units"
+
+        Quacc does not follow ASE standards for Molecular Dynamics units and instead
+        uses the following units:
+
+        - Time: femtoseconds (fs)
+        - Pressure: GPa
+        - Temperature: Kelvin (K)
+        - Compressibility: 1/GPa
```
Andrew-S-Rosen, line 41: I do wonder if this is going to introduce additional confusion simply because if people don't read the docstring (common), it'll just lead to a confusing mess. At the same time, the ASE units themselves are disgusting. I'm a bit torn.

src/quacc/recipes/emt/md.py
```diff
@@ -0,0 +1,69 @@
...
+    DynSchema
+        Dictionary of results, specified in [quacc.schemas.ase.summarize_md_run][].
+        See the type-hint for the data structure.
+    """
+    md_params = md_params or {}
+
+    calc = EMT(**calc_kwargs)
+    dyn = Runner(atoms, calc, copy_files=copy_files).run_md(**md_params)
+
+    return summarize_md_run(dyn, additional_fields={"name": "EMT MD"})
```
Andrew-S-Rosen, line 68: FYI: I moved a lot of the logic that was originally here into the `run_md` method. The recipes are meant to be minimal to make them more easily composable. Of course, if there are certain keyword arguments in `run_md` that are deemed worthwhile to make more user-facing, we can still include them as keyword arguments in the recipe too. For instance, right now, the user is passing in keyword arguments via `md_params`. But if we wanted to pull out something like `initial_temperature`, we could do that just like we pull out `relax_cell` in the optimizations. This is purely subjective.

Andrew-S-Rosen (commented):
@tomdemeyere: I am finally (virtually...) done cleaning things up. It took a fair bit longer than I expected...

Should you have a moment, would you be willing to address my few remaining questions? Again, no further coding is needed from you here.

src/quacc/runners/ase.py
```diff
@@ -364,3 +457,74 @@ def _set_sella_kwargs(self, optimizer_kwargs: dict[str, Any]) -> None:
...
+                "The `dt` kwarg is ASE deprecated and will"
+                " be interpreted as `timestep` in Quacc."
+            )
+            dynamics_kwargs["timestep"] = dynamics_kwargs.pop("dt")
+
+        if "fixrot" in dynamics_kwargs:
+            LOGGER.warning("`fixrot` is interpreted as `fix_rot` in Quacc.")
+
+        if "fix_rot" in dynamics_kwargs:
+            dynamics_kwargs["fixrot"] = dynamics_kwargs.pop("fix_rot")
```
Andrew-S-Rosen, line 528: For `fix_rot`, there is a logger warning (and corresponding swap) of ``fixrot` is interpreted as `fix_rot` in Quacc.`. However, I don't see `fixrot` in the ASE codebase anywhere. Is this actually needed?

src/quacc/runners/ase.py
```diff
@@ -309,6 +332,76 @@ def run_vib(self, vib_kwargs: VibKwargs | None = None) -> Vibrations:
...
+        if initial_temperature is not None:
+            MaxwellBoltzmannDistribution(
+                self.atoms,
+                temperature_K=initial_temperature,
+                rng=np.random.default_rng(seed=rng_seed) if rng_seed else None,
+            )
+            if fix_initial_com:
+                Stationary(self.atoms)
+            if fix_initial_rot:
+                ZeroRotation(self.atoms)
```
Andrew-S-Rosen, line 396: These `Stationary` and `ZeroRotation` classes are _only_ supposed to be used if a `MaxwellBoltzmannDistribution()` is applied, right? That's because the MD classes have their own `fixcm` keyword argument? [I did not see a similar `fixrot` keyword argument, for what it's worth].

src/quacc/runners/ase.py
```diff
@@ -309,6 +341,71 @@ def run_vib(self, vib_kwargs: VibKwargs | None = None) -> Vibrations:
...
 
+    def run_md(
+        self,
+        timestep: float = 1.0,
+        steps: int = 1000,
+        dynamics: MolecularDynamics = VelocityVerlet,
+        dynamics_kwargs: dict[str, Any] | None = None,
+        maxwell_boltzmann_kwargs: MaxwellBoltzmanDistributionKwargs | None = None,
+        fix_initial_com: bool = True,
+        fix_initial_rot: bool = True,
```
Andrew-S-Rosen, line 352: Having these default to `True` is the expected behavior for most MD users?

src/quacc/recipes/emt/md.py
```diff
@@ -0,0 +1,89 @@
...
+@job
+def md_job(
+    atoms: Atoms,
+    dynamics: MolecularDynamics = VelocityVerlet,
+    dynamics_kwargs: dict[str, Any] | None = None,
+    initial_temperature: float | None = None,
+    md_params: MDParams | None = None,
+    copy_files: SourceDirectory | dict[SourceDirectory, Filenames] | None = None,
+    **calc_kwargs,
+) -> DynSchema:
```
Andrew-S-Rosen, line 37: I have included some sensible default keyword arguments here to make it more accessible to users. What we include/not include is subject to discussion, of course.

tomdemeyere:
FYI: I have seen all your comments, but I am moving right now. I will make sure to come back to it ASAP 😀 !

Andrew-S-Rosen:
No worries, @tomdemeyere! I have been kind of going back and forth on a bunch of things, so I'm sorry about all the spam here. 😅 I should be done for now. Hopefully sometime later you can take a look and tell me what's a good or bad idea from here. Thanks for understanding.

I'm also moving next week!! Good luck with your move!

tomdemeyere:
> 1. When you have a moment, could you please re-review the changed files and see if this is to your liking?

I think it's nice (I skipped the part where the runners have been changed to a class!) you changed `fix_rot` and `fix_com` to `set_zero_rotation` and `set_stationary` my only concern here is that it is not immediately clear what `set_stationary` is, but with a little bit research this should be ok on the user side (`zero_com_momentum`, `zero_angular_momentum` might be better? I don't know)

> 2. In the quacc.recipes.emt.md.md_job recipe, I thought it would be nice to have temperature, pressure, and timestep as dedicated keyword arguments like what is done [here](https://github.com/materialsvirtuallab/matgl/blob/da3dab9affa09feb391351ab2e9fa38862f2c31d/src/matgl/ext/ase.py#L362). This is because the parameters are so critical to running MD and because it will potentially help clarify that users should be playing by "our rules" in terms of units. I have made that change accordingly. What do you think?

When I initially tackled the problem I had a few points in mind:

- Convention/units choice has to be uniform and well-thought: Kelvin for temperature, femtosecond for time, GPa for pressure... All these units are widely used, and they are all powers of SI units, so Quacc somehow follow international conventions.
- Users must be able to understand quickly what's going on, despite ASE being a mess on this side, this is especially challenging since some ASE's MD use different units with similar keywords.

With that in mind I was planning to add Quacc specific keywords such as "temperature_K", "pressure_gpa", and "timestep_fs", so that it would be clear that Quacc follow a different convention than ASE. ASE base keywords would still be allowed inside the `kwargs`. The problem with this approach is that it is not uniform; additional keywords (compressiblity, taut, etc) which are class specific are not integrated and will not follow the same convention, except if Quacc keeps track of every keywords, but that's a lot of burden on the dev side.

Instead, I went all-in by forcing every ASE keywords (deprecated or not...) to follow Quacc's new convention.

> I have made that change accordingly. What do you think

Pressure might be a little bit tricky, because `NPTBerendsen` expects a scalar while NPT (Nose-Hoover) can take both an array and scalar (`externalstress` keyword), and is represented as stress (so minus the pressure), if you are fine with the keyword `pressure` being used for multiple meaning, then it is fine. 

> 3.  Why do we have logs and handling for deprecated keyword arguments like pressure and temperature? Would it not be better to simply not accept them? Then there won't be any confusion about units. That was more-or-less my question from earlier. Instead of trying to be "smart", we could simply just not accept those old keywords at all.

Yes, but again, you have to be careful because these keywords are not deprecated in some classes, for example `temperature` and `temperature_K` both seem to work for `NVTBerendsen`.

Not sure if I am very helpful here, I think the current approach including your suggestions and changes is good. Although I must say I am still thinking about Quacc specific keywords, as mentioned above, as it would make it independent from what ASE is doing, at least for now.

Andrew-S-Rosen:
@tomdemeyere: Thank you for your review of this! Apologies this has dragged on for so long. I recognize your time is valuable and so appreciate your comments here. I will address your comments below later this week. I think they are all logical. I will ping you (one last time? 😅) when that's done to keep you updated.

> I think it's nice (I skipped the part where the runners have been changed to a class!) you changed fix_rot and fix_com to set_zero_rotation and set_stationary my only concern here is that it is not immediately clear what set_stationary is, but with a little bit research this should be ok on the user side (zero_com_momentum, zero_angular_momentum might be better? I don't know)

These are good suggestions. I found the original keywords a bit confusing because they are not actually "fixing" it per se rather they are setting it. But I agree my change is not clearer for other reasons. I will change to the syntax to something clearer based on your suggestion.

> Instead, I went all-in by forcing every ASE keywords (deprecated or not...) to follow Quacc's new convention.

This is more-or-less my concern --- that an ASE expert will come in and expect that if they are requesting an ASE keyword that it will behave as one would expect from ASE. 

I like your suggestion of quacc-specific keywords. The suggestions you have are particularly nice because there is no way that somebody is going to misinterpret the units on something like `time_fs` unless they are really clueless. It is true that something like `pressure_gpa` is lengthier, but I don't think a few characters will bother people. I am inclined to make this change and thank you for the suggestion.

If we go ahead with the custom keyword arguments, then we really aren't doing anything unusual with the units here. We don't need to have the behind-the-scenes transformations and logging. If the user passes in a custom ASE keyword argument, then that's on them to deal with. Additional keyword arguments like those related to compressibility would be missing (and require the user to supply an ASE keyword with ASE units), but there is also nothing stopping us from adding in one or more keyword arguments to the recipe if it is convenient for the end-user.

> Pressure might be a little bit tricky, because NPTBerendsen expects a scalar while NPT (Nose-Hoover) can take both an array and scalar (externalstress keyword), and is represented as stress (so minus the pressure), if you are fine with the keyword pressure being used for multiple meaning, then it is fine.

It looks like you may have meant `NPT`? If so and I'm understanding the documentation correctly, it does not look like this should be a major problem because `NPT` accepts `externalstress` but has no `pressure` keyword argument. So, this is a scenario where `pressure_gpa` (or whatever we name it) would not apply. `NVTBerendesen` looks to have just a `pressure_au` but not `externalstress`, unless I am missing something.

> EDIT: Also, the problem is that now you have pressure_au that takes a pressure in GPa 🙃

Thanks! Will fix!

Andrew-S-Rosen:
Hey @tomdemeyere, while reviewing your PR, I'd suggest the following code changes:

👉 [**Code Suggestion for #1672**](https://gitkraken.dev/drafts/8a91c80a-c14d-4883-9a9a-8050c90c9d85?source=PRComment)
>https://github.com/Quantum-Accelerators/quacc/pull/1672

You can also review and apply these suggestions [locally on your machine](https://gitkraken.dev/link/drafts/8a91c80a-c14d-4883-9a9a-8050c90c9d85?type=suggested_pr_change&prEntityId=WyJnaXRodWIiLCJwciIsIjEiLCIiLCJQUl9rd0RPR2pWLTBjNW1VS3FSIl0=&source=PRComment).

<details><summary>Learn more about GitKraken Code Suggest</summary>

>[Code Suggest](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbS9zb2x1dGlvbnMvY29kZS1zdWdnZXN0P3NvdXJjZT1naXRodWIuY29tJm1lZGl1bT1wcl9jb21tZW50?source=PRSuggest&event=redirectToCodeSuggestPage&provider=github) liberates your code reviews from GitHub's restrictive, comment-only feedback style. As simple as suggesting changes in a Google-doc, provide real code suggestions from where you code, e.g. your IDE, and on anything in your project — not just on the lines of code changed in the PR.

_Join your team on [GitKraken](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbT9zb3VyY2U9Z2l0aHViLmNvbSZtZWRpdW09cHJfY29tbWVudA==?source=PRSuggest&event=redirectToIndexPage&provider=github) to speed up PR review._
</details>

Andrew-S-Rosen:
@tomdemeyere: How does the proposal in https://github.com/Quantum-Accelerators/quacc/pull/2308 sound?

Andrew-S-Rosen:
Thanks again, @tomdemeyere for your help and patience! I'm going to go ahead and merge this in. Certainly, there remain areas for improvement, but I think this is a good place to be at. I will mark it as for "beta testing" in the changelog just for the time being.

PR #1682, opened 2024-02-10
Ensure all results are DB-compatible

Andrew-S-Rosen:
Requires:
- https://github.com/materialsvirtuallab/matgl/pull/225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - PDOS jobs and flows for Espresso
 
+### Changed
+
+- The `PRIMARY_STORE` setting has changed to `STORE` and is now formatted more intuitively
+
+### Fixed
+
+- Fixed usage of `MontyStore` as the data store option
+
 ### Removed
 
 - Removed the deprecated "q-chem legacy" recipes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - The `PRIMARY_STORE` setting has changed to `STORE` and is now formatted more intuitively
+- Uses 2024 stable release of `black`
 
 ### Fixed
 
--- a/docs/user/db/basics.md
+++ b/docs/user/db/basics.md
@@ -4,26 +4,34 @@ Here, we describe how to set up quacc with a database of your choosing.
 
 === "General Purpose"
 
-    For a given recipe, you can have quacc automatically store the final output summaries in your desired database by defining a [Maggma data store](https://materialsproject.github.io/maggma/reference/stores/) in the `PRIMARY_STORE` [quacc setting](../settings/settings.md).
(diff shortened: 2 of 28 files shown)


Andrew-S-Rosen:
@tomdemeyere: 

I am adding a test in this PR that ensures all results can safely be deposited in a database. There are two minor instances in the Espresso recipes that fail this test, as you can see [here](https://github.com/Quantum-Accelerators/quacc/actions/runs/7853478814/job/21433140349?pr=1682#step:7:545). Basically, it boils down to the fact that one can't have a "." in the key of a dictionary for most database formats, as this messes with querying.

The first instance of this is, as you can see in the aforementioned report, has to do with the lines [here](https://github.com/Quantum-Accelerators/quacc/blob/70dee9824febd49fec27db57a5064207899d9ab9/src/quacc/calculators/espresso/espresso.py#L210-L222). The `pwscf.dos` and `pwscf.pdos_tot` are stored as keys since `fildos.name` and `filpdos.name` are keys to the results dictionary. Any chance we can rename this for data storage purposes? I imagine this one might be okay if we do something like replace the "." with a more DB-friendly delimiter like a "_"? Let me know what you'd prefer, and I can make the change here.

The second instance is a bit trickier. Namely, when running the phonon workflow, the k-point info is stored like `(0.0, 0.0, 0.0)` as a key. This causes some trouble. Any idea on how we might get around this one?

For a minimal test, see below:

```python
from maggma.stores import MemoryStore

results = {"uuid": 12345, "pwscf.dos": "test"}
with MemoryStore() as store:
    store.update(results, key="uuid") # should not error out
```

tomdemeyere:
Feel free to change the naming stuff to an underscore

For phonons: https://gitlab.com/ase/ase/-/merge_requests/3252

tomdemeyere:
For the database there are a few things that came up in the office while working with Quacc:

- Would there be a way to add a SETTINGS that control the verbosity of results? from 1 to 3, default would be 3, 1 would only return plain results (atoms, energy, forces...) without any additional information.

- Would it be possible to send a label in static_job (maybe in additional_fields?) that would get carried down to the results?

Of course I explained that this can all be done manually, just giving some feedback here.

Andrew-S-Rosen:
> Feel free to change the naming stuff to an underscore

Thanks. Done!

> For phonons: https://gitlab.com/ase/ase/-/merge_requests/3252

Thanks. Merged! I just need to patch the new keys in `quacc` now.

> Would there be a way to add a SETTINGS that control the verbosity of results? from 1 to 3, default would be 3, 1 would only return plain results (atoms, energy, forces...) without any additional information.

I am hesitant to enable this because it will increase the fragility in writing new recipes. By having a constant schema, it's always known what keys are available to be used when designing a new `Flow`. If we have a dynamic schema, however, it means developers of new recipes will have to check what verbosity the user is running with and then, potentially, write the `Flow` accordingly, which is messy. It also means that the type hints would all be wrong (in the code and in the documentation). The principle of least surprise here wins out.

> Would it be possible to send a label in static_job (maybe in additional_fields?) that would get carried down to the results?

Ah, this was planned at some point but I had forgotten to do so. Yes, this should be done and is why `additional_fields` exists.

Andrew-S-Rosen:
Merging this in. Feel free to let me know if there are any problems!

PR #1701, opened 2024-02-13
Espresso Bands job and flow

Nekkrad:
## Summary of Changes

- Added job and flow to do a band calculation using quantum espresso
- Updated the documentation

### Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new Espresso bands-related jobs and a Fermi surface job to enhance material calculations.
	- Updated Quantum ESPRESSO functionality for better handling of unsupported binaries.
	- Expanded the `quacc` library with recipes for conducting bands and Fermi surface calculations.
- **Documentation**
	- Updated the list of available `quacc` recipes to include new Espresso bands-related and Fermi surface jobs.
- **Refactor**
	- Modified Espresso-related modules for improved clarity and efficiency in calculations.
- **Tests**
	- Added test cases for validating the new bands and Fermi surface calculations.


buildbot-princeton:

Nekkrad:
Do not merge this yet. I have to tweak few things

Nekkrad:
@Andrew-S-Rosen  This is ready to be merged from my side, but I think we should discuss if there is an easy way to generate a good k-path from the system space group. This PR so far does not specify any path so the results are crap unless the user specify a proper k-path

Andrew-S-Rosen:
> @Andrew-S-Rosen This is ready to be merged from my side, but I think we should discuss if there is an easy way to generate a good k-path from the system space group. This PR so far does not specify any path so the results are crap unless the user specify a proper k-path

Great! I will review this shortly. I'll be traveling today and tomorrow so will get to it by the 15th.

In the meantime, your question about k-path is a good one. There is a general-purpose [`quacc.utils.kpts`](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/utils/kpts.py) module containing various utilities for k-point/k-path handling. It builds off of Pymatgen's utilities for this. Refer to https://pymatgen.org/pymatgen.symmetry.html#module-pymatgen.symmetry.bandstructure, specifically the `HighSymmKpath` function. I would recommend using that function with the `latminer_munro` method described in this [article](https://www.nature.com/articles/s41524-020-00383-7). There is already some basic logic for this in [`quacc.utils.kpts.convert_pmg_kpts`](https://github.com/Quantum-Accelerators/quacc/blob/2020cd6fb890df6fe2900ee4c9e6207e2091d9e7/src/quacc/utils/kpts.py#L74-L82), but I haven't used that part for anything so you may need to tweak it/refactor it/test it. That's where I would recommend starting though!

src/quacc/recipes/espresso/dos.py
```diff
@@ -125,6 +129,49 @@ def projwfc_job(
...
+@job
+def bands_pp_job(
+    prev_dir: str | Path,
+    parallel_info: dict[str] | None = None,
+    test_run: bool = False,
+    **calc_kwargs,
+) -> RunSchema:
+    """
+    Function to carry out a basic bands.x calculation.
+    It re-orders bands, computes band-related properties and more. Fore more details please see
```
Andrew-S-Rosen, line 141: small typo: "fore"

src/quacc/recipes/espresso/core.py
```diff
@@ -350,3 +350,60 @@ def non_scf_job(
...
+def bands_job(
+    atoms: Atoms,
+    prev_dir: str | Path,
+    preset: str | None = "sssp_1.3.0_pbe_efficiency",
+    parallel_info: dict[str] | None = None,
+    test_run: bool = False,
+    **calc_kwargs,
+) -> RunSchema:
+    """
+    Function to carry out a basic bands calculation with pw.x.
```
Andrew-S-Rosen, line 365: Might be helpful to the user to mention that this is a non-self consistent calculation.

src/quacc/recipes/espresso/dos.py
```diff
@@ -125,6 +129,49 @@ def projwfc_job(
     )
 
 
+@job
+def bands_pp_job(
```
Andrew-S-Rosen, line 133: It would be helpful to clarify a bit more about the distinction between `bands_job` and `bands_pp_job`. Presumably the latter is only doing post-processing (i.e. no NSCF calculation) and relies on a prior `bands_job`? If a job relies on some other job, it's worth mentioning that in the docstring, including in the `prev_dir` description. We want to make it as easy as possible for users who may not be familiar with Espresso.

Andrew-S-Rosen, line 133: Is there any reason to have a separate post-processing job? Why not put the post-processing step within `band_job`? I assume that almost all users will want to do post-processing.

My recommendation would be to do both steps in `band_job`, removing the need for this separate job.

src/quacc/recipes/espresso/dos.py
```diff
@@ -311,3 +358,88 @@ def projwfc_flow(
...
+    job_params: dict[str, Any] | None = None,
+) -> BandsSchema:
+    """
+    This function performs a bands calculation.
+
+    Consists of following jobs that can be modified:
+
+    1. pw.x static
+        - name: "static_job"
+        - job: [quacc.recipes.espresso.core.static_job][]
```
Andrew-S-Rosen, line 376: I would suggest making this first step optional (e.g. `run_static: bool = True`) in case the user has already done a relax+static elsewhere and doesn't want to re-do that step. You can see an example of an optional step [here](https://github.com/Quantum-Accelerators/quacc/blob/2020cd6fb890df6fe2900ee4c9e6207e2091d9e7/src/quacc/recipes/vasp/slabs.py#L127).

That said, if my understanding is correct and it makes sense to merge `bands_job` and `bands_pp_job` into one step, then it does beg the question about whether a distinct flow is needed for this pattern. I would likely lean towards no if it's just static-->`bands_job`.

src/quacc/recipes/espresso/dos.py
```diff
@@ -311,3 +358,88 @@ def projwfc_flow(
...
+        the keys are the names of the jobs and the values are decorators.
+
+    Returns
+    -------
+    BandsSchema
+        Dictionary of results from [quacc.schemas.ase.summarize_run][].
+        See the type-hint for the data structure.
+    """
+
+    static_job_defaults = {"kspacing": 0.2}
```
Andrew-S-Rosen, line 404: Any reason for this default?

src/quacc/recipes/espresso/dos.py
```diff
@@ -311,3 +358,88 @@ def projwfc_flow(
...
+    bands_job_defaults = recursive_dict_merge(
+        job_params.get("static_job", {}),
+        {"input_data": {"control": {"calculation": "bands", "verbosity": "high"}}},
```
Andrew-S-Rosen, line 407: Maybe it makes sense to set `verbosity` to high in the `bands_job` function itself as a default?

Andrew-S-Rosen (commented):
Thank you very much, @Nekkrad!!! Overall, this is in good shape. I just have a few logistical questions for you that might help with cleaning things up.

src/quacc/recipes/espresso/core.py
```diff
@@ -350,3 +350,60 @@ def non_scf_job(
...
+
+    Parameters
+    ----------
+    atoms
+        The Atoms object.
+    prev_dir
+        Outdir of the previously ran pw.x calculation. This is used to copy
+        the entire tree structure of that directory to the working directory
+        of this calculation.
```
Andrew-S-Rosen, line 374: Could be helpful to mention what needs to be present in the previous directory. It's not just any `pw.x` calculation --- presumably it must be one with the charge density, right?

Nekkrad:
@Andrew-S-Rosen  Thanks!

Regarding your questions:

1. The only reason  behind having a separate post-processing job is because I was thinking of implementing a fermi surface job (fs.x in QE)  which requires a bands pw.x calculation, but not the post-processing. For all practical applications that I can come up with, one could merge bands_pp and the future fermi_job inside the bands_job. These are both very computationally cheap post-processing jobs. Another option is to merge only bands_pp  and have a different fermi_surface job

2. If we merge bands_pp job inside bands_job then there is no real need for the flow. You can directly to a static/relaxation --> bands_job.

I am willing to merge the bands_pp_job into bands_job, add k-path defaults with pymatgen. I can also remove the flow.

tomdemeyere:
@Andrew-S-Rosen 

QE has a lot of binaries... and I think you are right here to say that they probably do not all deserve a job. Initially I wanted each binary to have its own job because I believe it is very nice in a user experience point of view. But, considering *other* addons such as https://anharmonic.github.io/thermal2/ that contains more than 10 binaries alone... It might be well advised to do a "all_kind" job which would be able to run any kind of software using Espresso style input.

Few constraints that might be annoying:

1. This "all_kind" job will have to take a "binary" argument that would give the full path to the binary since we cannot list them in Quacc.
2. Similarly, it should have a "read_output_function" parameter that takes a function. The function will be written by the user to read the output(s)

Also it might be fair enough to say "let the user create their own job to run these binaries".

Andrew-S-Rosen:
Thanks for the input! I'm not opposed to having a bunch of individual jobs if you feel like people will use them frequently enough. In this case, my comment was mostly about the post-processing step, which I felt like could safely be done as part of the parent DOS job. 

Re: the "all" style of job, it's an interesting idea. Certainly, I can see value for it. One challenge that comes to mind that wasn't mentioned is that different binaries might have different sensible defaults as parameters, in which case one might as well separate them into individual jobs. Also, requiring users to provide a parsing function is quite messy. Indeed, under those scenarios, we might as well just assume that quacc is providing the infrastructure for people to do things on their own if they so choose.

@Nekkrad: This is all separate from the current PR, which I think looks great and just requires some minor adjustments so that users aren't running a dedicated post-processing step unnecessarily.

Nekkrad:
@Andrew-S-Rosen  Sorry been pretty busy.

Everything is good but I have a couple of questions

1. The k-path with pymatgen is done on the primitive cell, while this is generally a good choice some users may want to have a different k-path for whatever reason (e.g. conventional cell calculation). If this is the case then we may want to specify this in the documentation/function. Or we can come up with a smart way to address this, e.g. matrix algebra to map primitive path into the conventional path (this raises some issues tho, and it is not very extendable to other possible situations)

2. Looking at the slab flow you pointed me to, it still requires to define the jobs somewhere which I thought it is something we want to avoid. Another solution is to define a job, call multiple base_jobs : one bands, one pp and one optional fermi_surface and return a list of RunSchemas. Have you got in mind any other cleaner solution?

Andrew-S-Rosen:
> The k-path with pymatgen is done on the primitive cell, while this is generally a good choice some users may want to have a different k-path for whatever reason (e.g. conventional cell calculation). If this is the case then we may want to specify this in the documentation/function. Or we can come up with a smart way to address this, e.g. matrix algebra to map primitive path into the conventional path (this raises some issues tho, and it is not very extendable to other possible situations)

This is a fair point. Perhaps it would make sense to have the default route be to: 1) [make a primitive cell](https://pymatgen.org/pymatgen.symmetry.html#pymatgen.symmetry.analyzer.SpacegroupAnalyzer.get_primitive_standard_structure); 2) generate the pymatgen k-path. We could mention in the docs that this is what happens. If the user does wants to explicitly provide a k-path, then no transformation would be done on the structure. I'd avoid trying to map the primitive path to the conventional path --- seems messy.

> Looking at the slab flow you pointed me to, it still requires to define the jobs somewhere which I thought it is something we want to avoid. Another solution is to define a job, call multiple base_jobs : one bands, one pp and one optional fermi_surface and return a list of RunSchemas. Have you got in mind any other cleaner solution?

Apologies --- I'm trying to find my comment about the slab flow you mention but am having trouble seeing it. In any case, there are basically two options:

1. Define three separate jobs, one for bands, one for pp, one for fermi surface. In the bands job, you could have keyword options to call the pp and/or fermi surface, removing the `@job` decorator from the fast post-processing jobs via the [strip_decorator](https://quantum-accelerators.github.io/quacc/reference/quacc/wflow_tools/customizers.html#quacc.wflow_tools.customizers.strip_decorator) function. See a similar example [here](https://github.com/Quantum-Accelerators/quacc/blob/f71de24922fc311989f82fbed5019133a57c7bc1/src/quacc/recipes/vasp/core.py#L160-L177).

2. Simply define one job for bands that does as you suggested --- runs all three steps (optionally) and returns a _dictionary_ of schemas. You would return `{"bands": {blah}, "pp": {blah}, "fermi_surface": {}}` that way if the user does not want (say) the fermi surface, that key is just not present. Lists are a bit harder to deal with because you have to remember the specific order from the source code. A similar example is [here](https://github.com/Quantum-Accelerators/quacc/blob/f71de24922fc311989f82fbed5019133a57c7bc1/src/quacc/recipes/vasp/qmof.py#L29).

The first option only makes sense if there are scenarios where people would want to run a standalone fermi surface and/or post-processing job outside of the DOS calculation. If that's not the case, then option 2 makes more sense. But both are doing the same thing really.

Nekkrad:
@Andrew-S-Rosen 
This is a first commit for the new job, there are still some issues to address but let me know if it looks good.

The issues are

- There is no "fs" in the [espresso_key.py](https://gitlab.com/ase/ase/-/blob/master/ase/io/espresso_namelist/keys.py?ref_type=heads) in ASE.  I will open a PR later on today. This prevents us from running any fermi_surface calculations
- The function "_cleanup_params" [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/calculators/espresso/espresso.py) check whether we are using both kspacing and kpts. The problem occurs if .get("kpts") is a list cause it will raise Value error.

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+        prev_dir = bands_result["dir_name"]
+        bands_pp_results = _bands_pp(
+            atoms, prev_dir, parallel_info, test_run, **bands_pp_kwargs
+        )
+
+    if run_fermi_surface:
+        fermi_kwargs = job_params.get("fermi_surface", {})
+        prev_dir = bands_result["dir_name"]
+        fermi_results = _fermi_surface(
+            atoms, prev_dir, parallel_info, test_run, **fermi_kwargs
```
Andrew-S-Rosen, line 117: Some of these are kwargs and should be done as, e.g. `test_run=test_run`.

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+        )
+
+    results["bands"] = bands_result
```
Andrew-S-Rosen, line 120: Move this to right below `bands_result`

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+    results["bands_pp"] = bands_pp_results if run_bands_pp else None
```
Andrew-S-Rosen, line 121: Move this into the corresponding `if` statement that way we don't have the entry if the job isn't run.

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+    results["fermi_surface"] = fermi_results if run_fermi_surface else None
```
Andrew-S-Rosen, line 122: Move this into the corresponding `if` statement that way we don't have the entry if the job isn't run.

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+from quacc.utils.kpts import convert_pmg_kpts
+
+if TYPE_CHECKING:
+    from typing import Any, TypedDict
+
+    from ase.atoms import Atoms
+
+    from quacc.schemas._aliases.ase import RunSchema
+
+    class BandsSchema(TypedDict):
```
Andrew-S-Rosen, line 28: Do `BandsSchema(TypedDict, total=False)` to indicate that not all three entries may necessarily be present.

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+        template=EspressoTemplate("bands", test_run=test_run),
+        calc_defaults=calc_defaults,
+        calc_swaps=calc_kwargs,
+        parallel_info=parallel_info,
+        additional_fields={"name": "bands.x post-processing"},
+        copy_files=prev_dir,
+    )
+
+
+def _fermi_surface(
```
Andrew-S-Rosen, line 237: For this and the other tasks, perhaps what would be best would be to make them all `@job`s (and remove the `_` in front). Then, when they are called within `bands_job`, you can do

```python
from quacc.wflow_tools.customizers import strip_decorator

strip_decorator(fermi_surface)
```

Then one _could_ run the task as a standalone job if they wish. If you do this, feel free to add the jobs to the recipe list too.

src/quacc/recipes/espresso/bands.py
```diff
@@ -0,0 +1,284 @@
...
+    parallel_info: dict[str] | None = None,
+    test_run: bool = False,
+    job_params: dict[str, Any] | None = None,
+) -> BandsSchema:
+    """
+    Function to compute bands structure and fermi surface using pw.x, bands.x and fs.x.
+    This is all done in a single job as a multi-step process.
+
+    1. A pw.x non-self consistent calculation
+        - name: "bands"
```
Andrew-S-Rosen, line 50: Please also include the hyperlinks for these, e.g. `[quacc.recipes.espresso.bands.bands_pp][]`

tests/core/recipes/espresso_recipes/test_bands.py
Andrew-S-Rosen, line 1: The Fermi surface task isn't covered, so we'll want to address this.

Andrew-S-Rosen (commented):
Thanks, @Nekkrad! This looks very nice. I have left some comments.

> There is no "fs" in the [espresso_key.py](https://gitlab.com/ase/ase/-/blob/master/ase/io/espresso_namelist/keys.py?ref_type=heads) in ASE. I will open a PR later on today. This prevents us from running any fermi_surface calculations

I merged the MR in ASE mentioned by @tomdemeyere. Next time you make a commit to this PR, the updated ASE will be installed.

> The function "_cleanup_params" [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/calculators/espresso/espresso.py) check whether we are using both kspacing and kpts. The problem occurs if .get("kpts") is a list cause it will raise Value error.

Apologies. The issue here isn't immediately clear to me yet. Feel free to make an edit or perhaps if you wouldn't mind explaining a bit more, I can help.

tomdemeyere:
@Andrew-S-Rosen 

Thanks for keeping up with this whole espresso binary thing, I added a conditions that should help dealing with binaries that are not compatible with ASE.

Usually I think we (ASE) will have to choose between two paths:

1. We get rid of that binary thing and just accept input_data only in nested form, which admittedly would avoid plenty of headaches... (for the developers)
2. We setup some kind of automatic retrieval from espresso documentations to make sure that we are always up to date and not missing any binaries. Which is still not possible considering that binaries exists outside of the official espresso documentation...

Andrew-S-Rosen:
> Usually I think we (ASE) will have to choose between two paths:

> We get rid of that binary thing and just accept input_data only in nested form, which admittedly would avoid plenty of headaches... (for the developers)

I am generally of the opinion that there shouldn't be multiple ways to do something, but I also understand that this would be a breaking change and also likely inconvenient (per our prior conversations).

> We setup some kind of automatic retrieval from espresso documentations to make sure that we are always up to date and not missing any binaries. Which is still not possible considering that binaries exists outside of the official espresso documentation...

Yeah, this one seems like a lost cause IMO.

This one may be worth discussing in further detail with Ask.

Nekkrad:
@Andrew-S-Rosen 

Thanks to Tom's fixes we can now run fs.x. I have made few changes to the PR. The most important one is the following:

- Fermi surface (fs.x) calculations require a uniform unshifted k-points grid bands calculations. Usually you want to compute bands along a path which not uniform.  I have added another test to check the bands_job with fs.x using a uniform grid. Technically I can move everything into one test if you reckon adding an extra test is redundant.

docs/user/recipes/recipes_list.md
```diff
@@ -201,6 +201,12 @@ The list of available quacc recipes is shown below. The "Req'd Extras" column sp
 | Espresso DOS Flow        | `#!Python @flow` | [quacc.recipes.espresso.dos.dos_flow][]             |              |
 | Espresso Projwfc         | `#!Python @job`  | [quacc.recipes.espresso.dos.projwfc_job][]          |              |
 | Espresso Projwfc Flow    | `#!Python @flow` | [quacc.recipes.espresso.dos.projwfc_flow][]         |              |
+| Espresso Bands Job       | `#!Python @job`  | [quacc.recipes.espresso.bands.bands_job][]          |              |
+| Espresso Bands           | `#!Python @job`  | [quacc.recipes.espresso.bands.bands][]              |              |
+| Espresso Bands PP        | `#!Python @job`  | [quacc.recipes.espresso.bands.bands_pp][]           |              |
+| Espresso Fermi Surface   | `#!Python @job`  | [quacc.recipes.espresso.bands.fermi_surface][]      |              |
```
Andrew-S-Rosen, line 207: The naming convention in `quacc` is that all jobs end in `_job` (and all flows with `_flow`). Here, can you work on renaming some of these accordingly? You will have to be a bit creative with the two different bands jobs. 

The naming helps users because they don't have to look at the source code to know what type of function it is.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -349,7 +353,19 @@ def __init__(
         )
         self._bin_path = str(full_path)
         self._binary = template.binary
-        self._cleanup_params()
+
+        if self._binary in ALL_KEYS:
+            self._cleanup_params()
+        else:
+            LOGGER.warning(
+                f"the binary you requested `{self._binary}` is not supported by ASE, this means that presets and usual checks will not be carried, your input_data must be sent in nested format."
```
Andrew-S-Rosen, line 361: Slight cleanup requested:
```python
f"The binary you requested, `{self._binary}`, is not supported by ASE. This means that presets and usual checks will not be carried out, and your `input_data` must be provided in nested format.
```

src/quacc/recipes/espresso/_base.py
```diff
@@ -22,6 +24,9 @@
     from quacc.schemas._aliases.ase import RunSchema
 
 
+LOGGER = logging.getLogger(__name__)
```
Andrew-S-Rosen, line 27: This is not used in the function and should be removed.

Andrew-S-Rosen (commented):
Thank you very much, @Nekkrad! This looks great.

My only real remaining comments are about the function naming. Once that's cleaned up, I am happy to proceed with the merge.

Nekkrad:
@Andrew-S-Rosen 

Name changed, tell me if you like them. Not very inspired but probably ok ahah

Andrew-S-Rosen:
@Nekkrad Thanks for your edits! The names and such look good to me.

I think we can consider this one ready. Thank you for your modifications and input. I will proceed with the merge after I merge in #1759 shortly.

Andrew-S-Rosen:
@Nekkrad --- Thank you again for your contribution, and apologies for how long this has taken to get merged! I refactored this a bit to reduce some unnecessary code. I also changed the `bands_job` to a `bands_flow`. I know I was the one saying we should pack it all in a `@job`, but I thought about it some more and perhaps it is more intuitive to have it as a `@flow`. In any case, you can certainly feel free to open a follow-up PR and change it however you like (if you desire). I think you'll find the current code to be nice and clean :)

I am done with the cleanup here. The only challenge is that the tests run for a very long time now. Is there anything you can do in the parameters to make things less expensive? If not, I can look into splitting up the tests some more. You can see the longest running tests [here](https://github.com/Quantum-Accelerators/quacc/actions/runs/8101840145/job/22142985702?pr=1701#step:7:1)

Nekkrad:
@Andrew-S-Rosen No worries! I will take a look at it tomorrow. I can probably make them run very quickly, will let you know once it's done

Andrew-S-Rosen:
> @Andrew-S-Rosen No worries! I will take a look at it tomorrow. I can probably make them run very quickly, will let you know once it's done

Thanks!! That would be a huge help :) Will merge ASAP afterwards!

Nekkrad:
@Andrew-S-Rosen I added few keywords to the bands_pw_job and flow. These allow the users to have a little more flexibility if they decide to use make_bandpath. 

I also changed the parameters for the tests, hopefully they should run quicker. 

Let me know if they are still very slow and/or if there are other ONETEP or QE recipes that run too slowly

Andrew-S-Rosen:
@Nekkrad: Thanks a ton! The time was dropped for the most expensive run by nearly 2x, which  was a big help.

Here is the current timing:

```
============================= slowest 10 durations =============================
185.70s call     tests/core/recipes/espresso_recipes/test_bands.py::test_bands_flow
72.76s call     tests/core/recipes/espresso_recipes/test_dos.py::test_dos_flow
68.89s call     tests/core/recipes/espresso_recipes/test_dos.py::test_projwfc_flow
31.81s call     tests/core/recipes/espresso_recipes/test_phonons.py::test_phonon_job_list_to_do
14.21s call     tests/core/recipes/espresso_recipes/test_core.py::test_ase_relax_cell_job
11.72s call     tests/core/recipes/espresso_recipes/test_phonons.py::test_phonon_job
11.52s call     tests/core/recipes/espresso_recipes/test_core.py::test_relax_job_cell
6.47s call     tests/core/recipes/espresso_recipes/test_core.py::test_ase_relax_job
6.11s call     tests/core/recipes/espresso_recipes/test_core.py::test_relax_job
5.02s call     tests/core/recipes/espresso_recipes/test_core.py::test_static_job_v2
========== 49 passed, 8 skipped, 215 deselected in 457.10s (0:07:[37](https://github.com/Quantum-Accelerators/quacc/actions/runs/8110364645/job/22167485028#step:7:38)) ===========
```

I'm going to go ahead and merge this, but certainly if you can think of other words to reduce the timing, I would be greatly appreciative. At some point though, I will probably need to just split the Espresso tests into two separate parts either way. As for ONETEP, right now the test suite is only running the mocked tests because the "real" ONETEP execution would happen on the Princeton HPC machines, and I haven't yet successfully compiled ONETEP. Eventually...

Setting this to auto-merge!

PR #1712, opened 2024-02-15
Use a consistent default ASE optimizer

Andrew-S-Rosen:
## Summary of Changes

Sets a standardized, default ASE optimizers for ASE relaxations. This is a WIP.

### DIFF
--- a/src/quacc/recipes/emt/core.py
+++ b/src/quacc/recipes/emt/core.py
@@ -81,7 +81,7 @@ def relax_job(
         Dictionary of results, specified in [quacc.schemas.ase.summarize_opt_run][].
         See the type-hint for the data structure.
     """
-    opt_defaults = {"fmax": 0.01, "max_steps": 1000, "optimizer": FIRE}
+    opt_defaults = {"fmax": 0.01, "max_steps": 1000}
     opt_flags = recursive_dict_merge(opt_defaults, opt_params)
 
     atoms.calc = EMT(**calc_kwargs)
--- a/src/quacc/recipes/espresso/core.py
+++ b/src/quacc/recipes/espresso/core.py
@@ -220,7 +220,7 @@ def ase_relax_job(
         }
     }
 
-    opt_defaults = {"fmax": 0.01, "max_steps": 1000, "optimizer": LBFGS}
+    opt_defaults = {"fmax": 0.01, "max_steps": 1000}
 
     return base_opt_fn(
         atoms,
--- a/src/quacc/recipes/lj/core.py
+++ b/src/quacc/recipes/lj/core.py
@@ -79,7 +79,7 @@ def relax_job(
         Dictionary of results, specified in [quacc.schemas.ase.summarize_run][].
         See the type-hint for the data structure.
     """
-    opt_defaults = {"fmax": 0.01, "max_steps": 1000, "optimizer": FIRE}
+    opt_defaults = {"fmax": 0.01, "max_steps": 1000}
     opt_flags = recursive_dict_merge(opt_defaults, opt_params)
 
     atoms.calc = LennardJones(**calc_kwargs)
--- a/src/quacc/recipes/mlp/core.py
+++ b/src/quacc/recipes/mlp/core.py
@@ -91,7 +91,7 @@ def relax_job(
         See the type-hint for the data structure.
     """
 
-    opt_defaults = {"fmax": 0.05, "max_steps": 1000, "optimizer": FIRE}
+    opt_defaults = {"fmax": 0.05, "max_steps": 1000}
     opt_flags = recursive_dict_merge(opt_defaults, opt_params)
 
     atoms.calc = pick_calculator(method, **calc_kwargs)
--- a/src/quacc/recipes/onetep/core.py
+++ b/src/quacc/recipes/onetep/core.py
@@ -109,7 +109,7 @@ def ase_relax_job(
         {"keywords": {"write_forces": True, "forces_output_detail": "verbose"}},
     )
 
-    opt_defaults = {"fmax": 0.01, "max_steps": 1000, "optimizer": LBFGS}
+    opt_defaults = {"fmax": 0.01, "max_steps": 1000}
 
     return base_opt_fn(
         atoms,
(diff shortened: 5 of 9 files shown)


tomdemeyere:
Not that I know about no, our choice is purely motivated by empirical testing. Our group works with fairly different systems (Organics, bulk, slabs, clusters...) For the general case we found that the best optimiser is the "SciPyMinBFGS" with LBFGS being second best.

Since the SciPy is a little bit "exotic" I recommended to use LBFGS in Quacc.

Andrew-S-Rosen:
Good to know, thanks! I'll use LBFGS here.

> Since the SciPy is a little bit "exotic" I recommended to use LBFGS in Quacc.

It also does not support the `.irun()` feature in ASE, which I needed in #1711. :(

Andrew-S-Rosen:
(That said, BFGS _should_ perform virtually same or marginally better than LBFGS in most cases one would think! But perhaps reality says otherwise.)

Andrew-S-Rosen:
After giving this some more thought, I am going to make the default `BFGS`. That said, I will keep the LBFGS for Espresso and ONETEP given your preference. 

My motivation is as follows:
- It's nice to have some default
- It's clear from the [simple ASE benchmarks](https://wiki.fysik.dtu.dk/gpaw/devel/ase_optimize/ase_optimize.html) that FIRE is not that great. overall I have had issues close to the minimum of the PES as well, where FIRE said the structure is converged but all other methods found a solution with slightly lower energy at the same `fmax`.
- LBFGS and BFGS _should_ perform similarly (also shown in above benchmark) in terms of number of steps, but the former can be faster on a per-step basis for large systems. However, it also is not guaranteed to be as stable as BFGS.

An example of the last case can be found in this fun toy example:

```python
from ase.build import molecule
from ase.calculators.lj import LennardJones
from ase.optimize import LBFGS, BFGS

lj = LennardJones()
atoms = molecule('H2O')
atoms.calc = lj
LBFGS(atoms).run(fmax=1e-4)
```

```
       Step     Time          Energy          fmax
LBFGS:    0 17:44:11        1.772069       52.568263
LBFGS:    1 17:44:11       -1.601559        3.380930
LBFGS:    2 17:44:11       -1.694259        3.385208
LBFGS:    3 17:44:11       -0.572349        1.437886
LBFGS:    4 17:44:11       -0.273321        0.654332
LBFGS:    5 17:44:11       -0.152022        0.349696
LBFGS:    6 17:44:11       -0.078236        0.177432
LBFGS:    7 17:44:11       -0.038100        0.091912
LBFGS:    8 17:44:11       -0.014876        0.047141
LBFGS:    9 17:44:11       -0.002932        0.024202
LBFGS:   10 17:44:11        0.000000        0.000000
```

```python
from ase.build import molecule
from ase.calculators.lj import LennardJones
from ase.optimize import LBFGS, BFGS

lj = LennardJones()
atoms = molecule('H2O')
atoms.calc = lj
BFGS(atoms).run(fmax=1e-4)
```

```
      Step     Time          Energy          fmax
BFGS:    0 17:43:08        1.772069       52.568263
BFGS:    1 17:43:08       -1.601559        3.380930
BFGS:    2 17:43:08       -1.694259        3.385208
BFGS:    3 17:43:08        4.046608       76.780547
BFGS:    4 17:43:08       -1.760256        3.348774
BFGS:    5 17:43:08       -1.822601        3.276793
BFGS:    6 17:43:08       11.639889      156.865026
BFGS:    7 17:43:08       -1.856251        3.222236
BFGS:    8 17:43:08       -1.888426        3.156676
BFGS:    9 17:43:08        9.450609      148.745354
BFGS:   10 17:43:08       -1.926180        3.089243
BFGS:   11 17:43:08       -1.962342        3.007756
BFGS:   12 17:43:08       -2.325191       10.421584
BFGS:   13 17:43:08       -2.487222        1.619617
BFGS:   14 17:43:08       -2.889303        2.740923
BFGS:   15 17:43:08       51.852321      948.257913
BFGS:   16 17:43:08       -2.924469        1.772625
BFGS:   17 17:43:08       -2.926515        1.826648
BFGS:   18 17:43:08       -2.931914        1.929549
BFGS:   19 17:43:08       -2.939076        1.946583
BFGS:   20 17:43:08       -2.954127        1.700094
BFGS:   21 17:43:08       -2.967986        1.396977
BFGS:   22 17:43:08       -2.979426        0.793505
BFGS:   23 17:43:08       -2.983498        0.097157
BFGS:   24 17:43:08       -2.983561        0.007908
BFGS:   25 17:43:08       -2.983562        0.000059
```

Very far from the minimum, LBFGS can exhibit odd behavior where it doesn't find the desirable local minimum in the PES. BFGS, while taking more steps overall, finds the desired structure. FIRE also finds the same answer as BFGS but does so in 85 steps instead of 25.

tomdemeyere:
> ```python
> from ase.build import molecule
> from ase.calculators.lj import LennardJones
> from ase.optimize import LBFGS, BFGS
> 
> lj = LennardJones()
> atoms = molecule('H2O')
> atoms.calc = lj
> BFGS(atoms).run(fmax=1e-4)
> ```

It seems that the implementation of LBFGS and BFGS are not completely similar, LBFGS Hessian is updated at each step.

To be fair, all the tests we did were performed with a software where the noise on the force is large due to the basis set, hence our benchmark was more "which optimiser is going down the fastest?" and not "which optimiser is going to desired accuracy the fastest?" desired accuracy in our case was often unreachable.

Going further with your example here is some more tests I did with EMT. Each bar is an average over 1000 optimisations, all rattled randomly. (Very bad graph sorry, I did that very quickly, there should be a space before each blue bar to separate all the optimisers correctly).

I plot the number of actual force call, not steps so all line search are included. Order seems to be:

BFGSLineSearch < LBFGSLineSearch < SciPyMinBFGS < LBFGS ~ BFGS

image

Andrew-S-Rosen:
Interesting! Thanks for the follow-up analysis, @tomdemeyere! I'm not too surprised in hindsight about the performance of the LineSearch methods --- they tend to do quite well when you start with high forces in my experience (and it seems like from your experience the SciPyFminBFGS is generally better than BFGS). In any case, I think BFGS is at least better than FIRE as a default so am comfortable with that. But certainly if you have a strong opinion about changing things, don't hesitate to let me know.

I left Espresso and ONETEP as they are (i.e. you should see no change).

PR #1729, opened 2024-02-19
Write out results schema to filesystem in summarize functions

tomdemeyere:
## Summary of Changes

This has two purposes:

- Make sure the results are written safely somewhere while the file database problem is not sorted

- For calculators where nothing is written to the disk (MACE, EMT, ...) this results file gives sense to the currently empty directories

What do you think?

EDIT: Probably not going to be as easy as I was expecting

### Checklist

### Notes


buildbot-princeton:

Andrew-S-Rosen (commented):
Thanks! 

> EDIT: Probably not going to be as easy as I was expecting

How come? Because of the pickle error? See below.

As a general comment, it's unclear to me if pickle is the best approach here because the results are not easy to inspect outside of Python, and pickle is not guaranteed to work across Python versions. It's also not exactly how the results are stored in the database since it is the _serialized_ results that are stored. All schemas are all (de)serializable. As such, I would recommend the following:

```python
from monty.serialization import dumpfn

dumpfn(task_doc, "results.json") # it will get auto-gzizpped if `GZIP_FILES: bool = True`
```

You can read it back in a lossless way via

```python
from monty.serialization import loadfn

loadfn("results.json")
```

> Make sure the results are written safely somewhere while the file database problem is not sorted

Sure, that's fair. Most users of `quacc` are also MongoDB users and so we haven't had such issues, but I have long been aware that (seemingly convenient) filesystem-based stores are a large challenge for concurrent workflows in general. This is because databases that involve concurrent connections often need a client/server model, which is what most people using a filesystem-based data store are trying to avoid in the first place. This is a non-trivial issue but one I hope can be solved by others on the Materials Project team.

> For calculators where nothing is written to the disk (MACE, EMT, ...) this results file gives sense to the currently empty directories

They will only be empty for static calculations, as relaxations will dump out the log and restart files. In any case, it is perhaps worthwhile for us to deal with this directly if it is a nuisance. One option would be to see if `CREATE_UNIQUE_DIR` is `True` and if `directory` is blank. If so, the `directory` can be removed. Additionally, we would replace the value of `dir_name` in the output with a value of `None` to reflect the fact that it doesn't exist on the filesystem anymore.

This is not something you need to implement. I am just suggesting it as a viable option for later.

Andrew-S-Rosen:
I'll take a look later to figure out what's going on here.

tomdemeyere:
@Andrew-S-Rosen Thanks, I thought the same. It seems that a string is finding its way in the Monty code while it should be a dict.

Andrew-S-Rosen:
I am sorting this out. Seems to be related to https://github.com/materialsproject/emmet/issues/914.

Andrew-S-Rosen:
I just fixed what I believe is the upstream issue. Just waiting on a new release of [`emmet-core`](https://pypi.org/project/emmet-core/).

Andrew-S-Rosen:
Note to self for later: 

We can circumvent these issues by [`jsanitize`-ing](https://github.com/materialsvirtuallab/monty/blob/4d60fd4745f840354e7b3f48346eaf3cd68f2b35/monty/json.py#L635) the data beforehand.

Andrew-S-Rosen:
Alright, this is close... but not all the way there. This is definitely 100% do-able, but there are a few spots where `directory` is not defined appropriately (hence the errors) plus one serialization issue in the defect tests that is easily solvable.

Andrew-S-Rosen:
The light at the end of the tunnel: the checks have emerged!! Now we just need some tests 😄

tomdemeyere:
@Andrew-S-Rosen 

Thanks for taking care of that, I think that was a little bit beyond my python abilities, tests will come this weeks probably

Andrew-S-Rosen:
> EDIT: it seems that one particular line takes a lot of time (flat 30 seconds each time I run a calculation) I will turn off get_metadata for now

Is this a result of the change in the present PR, or are you just commenting more generally?

tomdemeyere:
> > EDIT: it seems that one particular line takes a lot of time (flat 30 seconds each time I run a calculation) I will turn off get_metadata for now
> 
> Is this a result of the change in the present PR, or are you just commenting more generally? Could you also provide some insight about the `Atoms` object? I am curious if there is a lot of data in `.info`. If this is a new thing, perhaps it is related to #1829. I doubt it is related to present PR specifically.

Ah yes! There is something inside the atoms.info, although it's not a lot just a simple short string. I assume this was related to this commit but it might not be you are right

tomdemeyere:
Here are the tests! Did not create new ones but made sure to test all summarise functions

Andrew-S-Rosen:
@tomdemeyere: Fantastic, thank you very much! I will proceed with the merge. 

Later on, if you find needing JSON is critical, let me know because I can likely make it work if needed.

tomdemeyere:
Just thought about it:

Would you mind adding a reserved key name in "additional_fields" which value would be used as a name for the pickle file?

{"pickle_label": "water_dimer_5.5"}

This is a nice niche feature that helps visual inspection?

Andrew-S-Rosen:
Personally, I'm a bit hesitant to add in "hidden" features where we overload a given method with multiple purposes (as you likely have observed from my comments about disliking the `.pop` business on user-supplied dictionaries).

PR #1737, opened 2024-02-21
Refactor subprocess calls in VASP

### LINKED ISSUE #1733: Direct call to run_custodian rather than subprocess call.

The current quacc.calculators.vasp.Vasp calculator can leverage custodian to handle errors, and does so by making the VASP command a subprocess call to a custom custodian script. This works but is less clean than just calling the custodian script in python directly. Working with bash/etc can also lead to small errors with environments/etc if users aren't careful.

Overloading the `_run()` method on the VASP calculator would be cleaner. Untested pseudocode below:

```
class DirectCustodianVasp(quacc.calculators.vasp.Vasp):

    def _run(self, command=None, out=None, directory=None):
        """Method to explicitly execute VASP"""
        if command is None:
            command = self.command
        if directory is None:
            directory = self.directory

        if not self.use_custodian:
            errorcode = subprocess.call(command,
                                    shell=True,
                                    stdout=out,
                                    cwd=directory)
        else:
            run_custodian()

        return

```

Andrew-S-Rosen:
Closes #1733.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ requires-python = ">=3.9, <3.13"
 dependencies = [
     "ase>3.22.1", # for Atoms object and calculators
     "cclib>=1.8", # for I/O parsing of molecular DFT codes
-    "custodian>=2023.6.5", # for automated error corrections
+    "custodian>=2024.2.15", # for automated error corrections
     "emmet-core>=0.69.6", # for pre-made schemas
     "maggma>=0.57.0", # for database handling
     "monty>=2024.2.2", # miscellaneous Python utilities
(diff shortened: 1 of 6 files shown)


Andrew-S-Rosen:
@zulissimeta: If you have a moment, could I get your review on this?

zulissimeta:
This looks reasonable to me! 

> @zulissimeta: If you have a moment, could I get your review on this?
> 
> Note: The lines are not shown as covered by codecov, but they are run on an HPC machine at Princeton via GitHub Actions, and the tests pass. The codecov report does not include the coverage from the HPC machine.

Andrew-S-Rosen:
Thanks, @zulissimeta! Merging now. Please let me know if any issues arise.

PR #1759, opened 2024-02-24
Refactor and streamline `copy_files` handling

### LINKED ISSUE #1742: Copy file passing breaks in dispatched workflows

- **quacc version**: latest
- **Python version**: 3.11

### What is the issue?

Any function that operates on inputs of a previous function should be a job so that the dependency is clear and the results are non-blocking. This is mostly what happens in quacc, including when looking up parameters like results["key"].

However, when passing file locations between tasks, file paths are specified as functions of the previous inputs using Path:
https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/vasp/core.py#L173

This causes a problem when running in prefect (and possibly others), because here `Path(summary[...]) / 'WAVECAR'` is actually a function operating on the previous summary, and Path() cannot be resolved until sometime in the future when the summary is actually available, similar to how __getitem__ is monkey patched in prefect with quacc. 

One possible fix for this is to use a helper function job for these cases everywhere copy_files is specified in quacc:
```
@job
def file_locations(summary, file_names):
    return [Path(summary["dir_name"]) / file_name for file_name in file_names]
```

When using prefect as the backend:
```python
from quacc.recipes.vasp.core import double_relax_job
double_relax(atoms)
```

Andrew-S-Rosen:
Closes #1742.

Note: 
- https://github.com/Parsl/parsl/issues/3108

## Motivation

In short, the `copy_files` handling was in desperate need of a behind-the-scenes overhaul. There are several reasons for this. 

1. As noted in #1742, you can't do `Path()` on a `Future`, which means any recipes doing `Path()` (or string concatenation of paths) in a `Flow` are going to crash when dispatched over a workflow engine.
2. There are several file handling functions that do redundant things or are handled in existing dependencies: `copy_decompress_files`, `copy_decompress_tree`, `copy_decompress_files_from_dir`.
3. Because of Point 2, not all the functions handle things in the same way. For instance, only some have glob support right now, which is confusing.
4. There are many instances where the type hints are simply inaccurate. For instance, in the Espresso recipes, the `prev_dir: str | Path` positional argument is often getting `dict` entries thrown into, as a result of `pw_copy_files` and such (see the test suite). 

Because of this mess, things were in desperate need of a revamp, which this PR addresses.

## User-Facing Changes

In most cases, there are no breaking changes to users and is mostly an internal refactor. Some positional arguments of `prev_dir` have been renamed `copy_files` for consistency with the type hint, although no user changes are expected if used as positional. Some internal functions have slightly modified behavior, such as `copy_decompress_files`. Refer to the function signature for more.


Andrew-S-Rosen:
@Nekkrad, @tomdemeyere: Thank you very much for your patience with the PR reviews while I was out of town and then trying to handle this absolute nightmare of a refactor.

I am tagging you in this PR because, as things are implemented right now, there is a slight but notable breaking change to some of the Espresso recipes. The easiest way for to summarize this would be to check out the "Files changed" tab and go to the tests for the Espresso recipes. For example, [here](https://github.com/Quantum-Accelerators/quacc/pull/1759/files#diff-9c4eaf29e43983160b1ca5709dfc2d85023084db1f1622a1138c50744e88a15dL100) you will notice that

```python
pp_results = post_processing_job(prev_dir=results["dir_name"])
```

was changed to

```python
pp_results = post_processing_job({results["dir_name"]: "*"})
```

Why the change? The long answer is see my description in the PR summary. The short answer is as follows:

To streamline things, when a user passes in a directory to the `copy_files` keyword argument used in the various recipes, the directory will be copied as-is. This is in contrast to the old behavior where everything got unpacked. If you want all the files in a directory but _not_ the directory itself the way to do that would be to use the `*` glob pattern. The following three methods do exactly the same thing:

```python
pp_results = post_processing_job({results["dir_name"]: "*"})
pp_results = post_processing_job(f"{results['dir_name']}/*")
pp_results = post_processing_job(Path(results["dir_name"], "*"))
```

In general, however, I am recommending that users take the dictionary approach because if `results` is a `Future` (e.g. when dispatched via a workflow engine), you can get a crash as noted in #1742. 

I recognize that this would be a breaking change, and changes are annoying. I generally try to avoid them whenever possible. However, I have been unable to come up with a solution that would be clear and retain old functionality in this scenario.

I also understand that, given the huge size of this PR, my intentions may not be clear. I am happy to hop on a brief call if that would be helpful. In any case, while this PR is critical, I would like to open this up for discussion before I go ahead and merge.

Nekkrad:
@Andrew-S-Rosen I will be looking at this probably later today or tomorrow. Feel free to merge and if there are breaking changes me or Tom will fix them

Andrew-S-Rosen:
Thanks, @Nekkrad! Much appreciated. I'll likely go ahead and merge this in the evening here so that it can unblock a few other PRs (including one of yours). I'll make any changes to your PR to ensure everything passes after I do --- no need for you to do anything there. In any case, I wanted to communicate this minor but notable change in terms of how files are handled.

Andrew-S-Rosen:
@tomdemeyere No worries. Try not to open GitHub too much on your time off. 😉 

Much of this refactoring were due to nice feature additions you made, so thank you! I will go ahead and merge this shortly, but when you return, don't hesitate to reach out if there are any issues/concerns.

PR #1789, opened 2024-02-29
Espresso: q2r and matdyn jobs 

tomdemeyere:
## Summary of Changes

End of the implementation of the phonon jobs, q2r.x is used to transform the force constant matrix from reciprocal space to real space, matdyn.x can then interpolate it to any q-point.

This is needed to perform phonon-dos with Espresso.

I also added a flow that does all of that (Note to self: test is missing for this flow)

EDIT: I think moving the sanity checks from espresso.utils into the template will be needed at some point, there are a lot of self attributes that this function should access.

### Checklist

### Notes


buildbot-princeton:

src/quacc/recipes/espresso/phonons.py
```diff
@@ -84,12 +84,201 @@ def phonon_job(
...
+    Returns
+    -------
+    RunSchema
+        Dictionary of results from [quacc.schemas.ase.summarize_run][].
+        See the type-hint for the data structure.
+    """
+
+    calc_defaults = {"input_data": {"input": {"fildyn": fildyn, "flfrc": "q2r.fc"}}}
+
+    prev_dir = {prev_dir: [f"{fildyn}*"]}
```
Andrew-S-Rosen, line 126: This is smart. The reason that I like this is that you aren't relying on the user to know what files to provide. You're just asking them for the directory, and then the code takes care of the rest. Everyone knows their prior directory, so it's easy.

I wonder if this kind of logic can/should be implemented in other Espresso `@job`s that are inherently a follow-up. It's probably worth thinking about. Right now, they are just assuming the user provides the right thing.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -84,12 +84,201 @@ def phonon_job(
...
+        Custom parameters to pass to each Job in the Flow. This is a dictionary where the keys are the names of the jobs and the values are dictionaries of parameters.
+    job_decorators
+        Custom decorators to apply to each Job in the Flow. This is a dictionary where the keys are the names of the jobs and the values are decorators.
+
+    Returns
+    -------
+    RunSchema
+        Dictionary of results from [quacc.schemas.ase.summarize_run][].
+        See the type-hint for the data structure.
+    """
```
Andrew-S-Rosen, line 229: For `@flow`s, I find it a bit nicer to return a dictionary containing _all_ of the results. So that would be something like

```python
{"pw": pw_job_results, "ph": ph_job_results, "fc": fc_job_results, "dos": dos_job_results}
```

with a type hint to clarify this to the user.

Andrew-S-Rosen (commented):
Thank you for the contribution, @tomdemeyere!! This is already in great shape! I know you're still working on it, but I wanted to leave some preliminary comments.

I also updated the `copy_decompresss_files` calls in the test suite to match the new function I introduced recently. Hopefully I didn't break anything else in this PR of yours in the process of that change.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -84,12 +84,201 @@ def phonon_job(
...
+
+    calc_defaults = {"input_data": {"input": {"fildyn": fildyn, "flfrc": "q2r.fc"}}}
+
+    copy_files = {prev_dir: [f"{fildyn}*"]}
```
Andrew-S-Rosen, line 126: This is smart. The reason that I like this is that you aren't relying on the user to know what files to provide. You're just asking them for the directory, and then the code takes care of the rest. Everyone knows their prior directory, so it's easy.

I wonder if this kind of logic can/should be implemented in other Espresso `@job`s that are inherently a follow-up. It's probably worth thinking about. Right now, they are just assuming the user provides the right thing.

tests/core/recipes/espresso_recipes/test_phonons.py
```diff
@@ -188,3 +193,27 @@ def test_matdyn_job(tmp_path, monkeypatch):
...
+
+
+def test_phonon_dos_flow(tmp_path, monkeypatch):
+    monkeypatch.chdir(tmp_path)
+
+    atoms = bulk("Li")
+
+    copy_decompress_files([DATA_DIR / "Li.upf.gz"], tmp_path)
+
+    SETTINGS.ESPRESSO_PSEUDO = tmp_path
```
Andrew-S-Rosen, line 205: We want to make sure this is returned to its default value after the test is done.

So, do `DEFAULT_SETTINGS = SETTINGS.model_copy()` at the top of the module. And then at the bottom of this function do `SETTINGS.ESPRESSO_PSEUDO = DEFAULT_SETTINGS.ESPRESSO_PSEUDO`. 

[It'd be nice](https://github.com/Quantum-Accelerators/quacc/issues/1791) of there were a way to cleanly do this with `monkeypatch.setattr`, but I've struggled to do so.

tomdemeyere:
@Andrew-S-Rosen I have seen some of your messages about tests taking time, can you try to set "export OMP_NUM_THREADS=1" or maybe 2?

At some point I have seen something like 8 in the test outputs, which I suspect might slow things down on this particular setup?

Andrew-S-Rosen:
> @Andrew-S-Rosen I have seen some of your messages about tests taking time, can you try to set "export OMP_NUM_THREADS=1" or maybe 2?
> 
> At some point I have seen something like 8 in the test outputs, which I suspect might slow things down on this particular setup?

Thanks! Good suggestion. I noticed that on my local machine which has a very large amount of threads, the tests take a lot longer than on CI for that reason. I think the proper way to do this would be `monkeypatch.setenv("OMP_NUM_THREADS", 2)` in the relevant tests. I can try that separately from this PR to see if/where it helps.

tomdemeyere:
> > @Andrew-S-Rosen I have seen some of your messages about tests taking time, can you try to set "export OMP_NUM_THREADS=1" or maybe 2?
> > At some point I have seen something like 8 in the test outputs, which I suspect might slow things down on this particular setup?
> 
> Thanks! Good suggestion. I noticed that on my local machine which has a very large amount of threads, the tests take a lot longer than on CI for that reason. I think the proper way to do this would be `monkeypatch.setenv("OMP_NUM_THREADS", 2)` in the relevant tests. I can try that separately from this PR to see if/where it helps.

I've just tried here with your suggestion, also adding parallel info to have 2 processes, let's see

tomdemeyere:
Before: image

After:
image

Not bad! Only phonons and bands test have been setup here, I will fix it back and open a separate PR, when applied on every tests we should be able to get that to 3 min or so

Andrew-S-Rosen:
> @Andrew-S-Rosen
> 
> Before: image
> 
> After: image
> 
> Not bad! Only phonons and bands test have been setup here, I will fix it back and open a separate PR, when applied on every tests we should be able to get that to 3 min or so

Thank you very much, @tomdemeyere! That is great!

src/quacc/recipes/espresso/phonons.py
```diff
@@ -84,6 +90,201 @@ def phonon_job(
...
+
+    This job requires the results of a previous q2r.x calculation, you might
+    want to create your own flow to run both jobs in sequence.
+
+    Parameters
+    ----------
+    prev_dir
+        Outdir of the previously ran q2r.x calculation. This is used to copy
+        the the force constant file.
+    flfrc
```
Andrew-S-Rosen, line 163: Is this a standard name in Espresso? Because it's super cryptic to me 😅

src/quacc/recipes/espresso/phonons.py
```diff
@@ -84,6 +90,201 @@ def phonon_job(
...
+
+    This job requires the results of a previous ph.x calculation, you might
+    want to create your own flow to run both jobs in sequence.
+
+    Parameters
+    ----------
+    prev_dir
+        Outdir of the previously ran ph.x calculation. This is used to copy
+        the the dynamical matrix files.
+    fildyn
```
Andrew-S-Rosen, line 112: I'd prefer something a bit more descriptive like `matdyn_file` (unless `fildyn` is widely acknowledged by  the Espresso community).

Andrew-S-Rosen (commented):
This looks great, @tomdemeyere! Just a few minor comments from me on the names of keyword arguments, but otherwise I am happy to merge this once everything passes and you give me the okay.

src/quacc/recipes/espresso/phonons.py
Andrew-S-Rosen: Sure, either approach is fine with me. The new approach looks nice.

tomdemeyere:
This should be ok to merge!

Andrew-S-Rosen:
Thank you very much, @tomdemeyere!

Issue #1774, opened 2024-02-27
Add a non-SCF job for VASP


Andrew-S-Rosen:
### What new feature would you like to see?

Ideally, we should take some inspiration from the typical MP non-SCF workflow in [atomate2](https://github.com/materialsproject/atomate2/blob/e9fe5c27a4a9d24a90d964964d5fe809aa78d5f0/src/atomate2/vasp/sets/core.py#L168) as well in terms of settings.

yw-fang:
Since I inquired about the feature of non-SCF job for VASP several days ago and no one others are working on it, I implemented a little bit about it 1b4ab8e217ae85127bff6f25fbe6dee66f58bd09
I still need some time to further meet the guidelines of code contributions if @Andrew-S-Rosen thinks this implementation looks not so bad

Andrew-S-Rosen:
Thanks for the comment and your effort, @yw-fang! It is definitely a solid start. Should it be of interest to you, I would be happy to provide initial comments and feedback on it if you decide to open a PR. Most of the content is there; it just requires some cleanup of things.

Issue #1708, opened 2024-02-15
Ensure all `with pytest.raises()` checks in the test suite have an associated `match`


Andrew-S-Rosen:
### What would you like to report?

We want to make sure we catch a specific exception. When done, the following lines in `pyproject.toml` should be removed:

https://github.com/Quantum-Accelerators/quacc/blob/d95ca63dab1e2ca33930b4254a0bd8416c88bad8/pyproject.toml#L161-L163

PR #1840, opened 2024-03-05
[WIP] nscf_job for vasp calculations

yw-fang:
## Summary of Changes

\>> It aims to implement the non-SCF job for VASP. https://github.com/Quantum-Accelerators/quacc/issues/1774  <<

### Checklist

### Notes

### DIFF
--- a/src/quacc/recipes/vasp/core.py
+++ b/src/quacc/recipes/vasp/core.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 
 from quacc import flow, job
 from quacc.recipes.vasp._base import base_fn
+import numpy as np
 
 if TYPE_CHECKING:
     from typing import Any
@@ -14,6 +15,110 @@ if TYPE_CHECKING:
 
     from quacc.schemas._aliases.vasp import DoubleRelaxSchema, VaspSchema
     from quacc.utils.files import Filenames, SourceDirectory
+    from pymatgen.io.vasp import Vasprun
+
+@job
+def nscf_job(
+    atoms: Atoms,
+    preset: str | None = "BulkSet",
+    copy_files: SourceDirectory | dict[SourceDirectory, Filenames] | None = None,
+    kpoints_mode: str = "line",
+    calculate_optics: bool = False,
+    bandgap: float = None,
+    vasprun: Vasprun = None,
+    **calc_kwargs,
+) -> VaspSchema:
+    """
+    Carry out a non-self-consistent field (NSCF) calculation.
+
+    Parameters
+    ----------
+    atoms
+        Atoms object.
+    preset
+        Preset to use from `quacc.calculators.vasp.presets`.
+    copy_files
+        Files to copy (and decompress) from source to the runtime directory.
+    kpoints_mode
+        Type of k-points mode. Options are "uniform" or "line". "boltztrap" may
+        be implemented in future.
+    calculate_optics
+        Whether to calculate optical properties.
+    bandgap
+        The band gap of the material.
+    **calc_kwargs
+        Custom kwargs for the Vasp calculator. Set a value to
+        `None` to remove a pre-existing key entirely. For a list of available
+        keys, refer to [quacc.calculators.vasp.vasp.Vasp][].
+
+    Returns
+    -------
+    VaspSchema
+        Dictionary of results from [quacc.schemas.vasp.vasp_summarize_run][].
+        See the type-hint for the data structure.
+    """
+
+    # Normalize kpoints_mode
+    kpoints_mode = kpoints_mode.lower()
+
+    # Validate kpoints_mode
+    supported_modes = ("line", "uniform")
+    if kpoints_mode not in supported_modes:
+        raise ValueError(f"Supported kpoints modes are: {', '.join(supported_modes)}")
+
+    calc_defaults = {
+        "lorbit": 11,
+        "lwave": False,
+        "lcharg": False,
+        "nsw": 0,
+        "isym": 0,
+        "icharg": 11,
+        "kspacing": None,
+        "nedos": 5001,
+    }
+
+    updates: dict[str, Any] = {}
+
+    if vasprun is not None:
+        nbands_factor = 1.2
+        nbands = int(np.ceil(vasprun.parameters["NBANDS"] * nbands_factor))
+        updates["nbands"] = nbands
+
+    def _get_nedos(vasprun: Vasprun | None, dedos: float) -> int:
+        """Automatic setting of nedos using the energy range and the energy step."""
+        if vasprun is None:
+            return 5001
+        emax = max(eigs.max() for eigs in vasprun.eigenvalues.values())
+        emin = min(eigs.min() for eigs in vasprun.eigenvalues.values())
+        return int((emax - emin) / dedos)
+
+    if kpoints_mode == "uniform":
+        # Automatic setting of NEDOS using the energy range and the energy step
+        n_edos = _get_nedos(vasprun, 0.005)  # Example value, adjust as needed
+        # Use tetrahedron method for DOS and optics calculations
+        updates.update({"ismear": -5, "isym": 2, "nedos": n_edos})
+    elif kpoints_mode == "line":
+        sigma = 0.2 if bandgap == 0 else 0.01
+        updates.update({"ismear": 0, "sigma": sigma})
+
+    if calculate_optics:
+        # LREAL not supported with LOPTICS = True; automatic NEDOS usually
+        # underestimates, so set it explicitly
+        n_edos = _get_nedos(vasprun, 0.0025)  # Example value, adjust as needed
+        updates.update({"loptics": True, "lreal": False, "cshift": 1e-5, "nedos": n_edos})
+
+    # integrate updates to calc_kwargs
+    calc_kwargs.update(updates)
+
+
+    return base_fn(
+        atoms,
+        preset=preset,
+        calc_defaults=calc_defaults,
+        calc_swaps=calc_kwargs,
+        additional_fields={"name": "VASP NSCF"},
+        copy_files=copy_files,
+    )
 
 
 @job


buildbot-princeton:

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+    }
+
+    updates: dict[str, Any] = {}
+
+    if vasprun is not None:
+        nbands_factor = 1.2
+        nbands = int(np.ceil(vasprun.parameters["NBANDS"] * nbands_factor))
+        updates["nbands"] = nbands
+
+    def _get_nedos(vasprun: Vasprun | None, dedos: float) -> int:
```
Andrew-S-Rosen, line 89: Since we may use this function in other places, we should move it out of the recipe definition. A good location might be [`quacc.calculators.vasp.params`](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/calculators/vasp/params.py). I would also suggest having it be `vasprun: Vasprun` and simply assume that it is being provided.

The function would also need a proper docstring if moved.

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+        """Automatic setting of nedos using the energy range and the energy step."""
+        if vasprun is None:
+            return 5001
```
Andrew-S-Rosen, line 92: We can assume a `vasprun` is provided and remove this logic.

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+        emax = max(eigs.max() for eigs in vasprun.eigenvalues.values())
+        emin = min(eigs.min() for eigs in vasprun.eigenvalues.values())
+        return int((emax - emin) / dedos)
+
+    if kpoints_mode == "uniform":
+        # Automatic setting of NEDOS using the energy range and the energy step
+        n_edos = _get_nedos(vasprun, 0.005)  # Example value, adjust as needed
```
Andrew-S-Rosen, line 99: We should have the `dedos` here be an optional keyword argument that the user can supply. Let's also use the `0.02` default (i.e. `dedos: float = 0.02`) unless you suggest otherwise.

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+        # Use tetrahedron method for DOS and optics calculations
+        updates.update({"ismear": -5, "isym": 2, "nedos": n_edos})
+    elif kpoints_mode == "line":
+        sigma = 0.2 if bandgap == 0 else 0.01
```
Andrew-S-Rosen, line 103: Since the `bandgap` produced by VASP may not be exactly 0 yet still be metallic, we should have this be something more like `if bandgap < 1e-4 else 0.01` instead.

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+        updates.update({"ismear": 0, "sigma": sigma})
+
+    if calculate_optics:
+        # LREAL not supported with LOPTICS = True; automatic NEDOS usually
+        # underestimates, so set it explicitly
+        n_edos = _get_nedos(vasprun, 0.0025)  # Example value, adjust as needed
+        updates.update(
+            {"loptics": True, "lreal": False, "cshift": 1e-5, "nedos": n_edos}
+        )
```
Andrew-S-Rosen, line 112: I'd personally suggest we remove this block for now. If we want to allow optical calculations, we can have a dedicated `optics_job` that essentially calls the `nscf_job` with some updated parameters to add. This will help keep the `nscf_job` easier to read.

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+    # integrate updates to calc_kwargs
+    calc_kwargs.update(updates)
+
+    return base_fn(
+        atoms,
+        preset=preset,
+        calc_defaults=calc_defaults,
+        calc_swaps=calc_kwargs,
+        additional_fields={"name": "VASP NSCF"},
+        copy_files=copy_files,
```
Andrew-S-Rosen, line 123: Since the user is providing a `prev_dir`, we need to define `copy_files` here. We don't need to copy every file in the directory, so we can safely do in the line before this function call:

```python
copy_files = {prev_dir: ["CHGCAR*", "WAVECAR*"]}
```

Andrew-S-Rosen (commented):
@yw-fang: Thank you very much for your contribution! This is off to a great start. I have provided some comments here that will hopefully help streamline this function. Please let me know if you have any questions.

Once that is completed, tests will need to be added. They can be found in [`quacc/tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes`](https://github.com/Quantum-Accelerators/quacc/blob/main/tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py). You can follow the examples in there to make a new test case for the `nscf_job`, evaluating the parameters to make sure they are correct given a set of inputs. The VASP calculator is "mocked" in this test suite (see `conftest.py` for details), so it is not running the VASP executable but rather `EMT` in the background.

Additionally, you will want to add this recipe to the [list of recipes](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/user/recipes/recipes_list.md) in the documentation.

Finally, you may (if you wish) add yourself to the [contributors section](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/about/contributors.md) in the documentation.

src/quacc/recipes/vasp/core.py
```diff
@@ -4,18 +4,126 @@
...
+        See the type-hint for the data structure.
+    """
+
+    # Normalize kpoints_mode
+    kpoints_mode = kpoints_mode.lower()
+
+    # Validate kpoints_mode
+    supported_modes = ("line", "uniform")
+    if kpoints_mode not in supported_modes:
+        raise ValueError(f"Supported kpoints modes are: {', '.join(supported_modes)}")
```
Andrew-S-Rosen, line 69: You can remove this section and instead have the check be done below.

```python
elif kpoints_mode == "line":
    ...
else:
   raise ValueError(...)
```

src/quacc/recipes/vasp/core.py
```diff
@@ -16,6 +21,122 @@
...
+        vasprun_path = Path(prev_dir, "vasprun.xml")
+        if (vasprun_path_gz := Path(str(vasprun_path) + ".gz")).exists():
+            vasprun_path = zpath(
+                vasprun_path_gz
+            )  # if vasprun.xml.gz, zpath will decompress it
+            vasprun = Vasprun(vasprun_path)
+        else:
+            print(
+                "Warning: vasprun.xml* file does not exist in the specified directory."
+            )
```
Andrew-S-Rosen, line 99: Logging statements should be used instead of print statements. See [here](https://github.com/Quantum-Accelerators/quacc/blob/32e3316536d7e19a7367f97e1d5a03791557dfd3/src/quacc/recipes/mlp/_base.py#L42) for an example.

src/quacc/recipes/vasp/core.py
```diff
@@ -16,6 +21,122 @@
...
+    for file_name in basics_to_copy[1:]:  # actually it checks the existence of WAVECAR*
+        matching_files = [
+            file for file in os.listdir(prev_dir) if file.startswith(file_name[:-1])
+        ]
+        if not matching_files:
+            print(
+                f"Warning: {file_pattern} file does not exist in the specified directory."
+            )
```
Andrew-S-Rosen, line 107: [We already log a warning](https://github.com/Quantum-Accelerators/quacc/blob/32e3316536d7e19a7367f97e1d5a03791557dfd3/src/quacc/utils/files.py#L143) if specified files aren't found. We should aim to keep the recipe as clean and minimal as possible. Since this check is already done internally, we should remove this block. I also don't know that we need to have the other `FileNotFoundError`s above. The warning will be raised, and VASP will crash. That is enough. We shouldn't try to overengineer things too much.

src/quacc/recipes/vasp/core.py
Andrew-S-Rosen, line 99: This whole block can simply be replaced by

```python
vasprun = Vasprun(zpath(Path(prev_dir, "vasprun.xml")))
```

That's all that's needed.

Andrew-S-Rosen:
@yw-fang I left some comments as you're working on this. In general, resist the need to overengineer things :) the earlier version of the code was largely good and only needed the changes suggested.

src/quacc/recipes/vasp/core.py
```diff
@@ -16,6 +21,122 @@
...
+        "lcharg": False,
+        "nsw": 0,
+        "isym": 0,
+        "icharg": 11,
+        "kspacing": None,
+        "nedos": 5001,
+    }
+
+    # check the expected files in prev_dir
+    basics_to_copy = ["CHGCAR*", "WAVECAR*", "vasprun.xml*"]
```
Andrew-S-Rosen, line 81: We do not need (and should not) copy the `vasprun.xml` file to the new VASP working directory.

yw-fang:
Hi, @Andrew-S-Rosen 

I am wondering how can we let pytest read the vasprun.xml.gz included in `quacc/tests/core/recipes/vasp_recipes/mocked/mocked_vasp_run`?
I couldn't pass one unit test because it seems to use Vasprun from pymatgen to parse a real vasprun.xml (the test script is 5673552222c859e890fb86a503c98fd1f8bd8ccf).

The specific errors from pytest was
```
        # test nbands when vasprun.xml exists
        (tmp_path / "vasprun.xml").touch()
>       output = nscf_job(
            atoms,
            prev_dir=tmp_path,
            nbands_factor=1.2,
            kpoints_mode="uniform",
            **calc_kwargs,
        )

/home/ywfang/git/quacc/tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py:71: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/ywfang/git/quacc/src/quacc/recipes/vasp/core.py:91: in nscf_job
    vasprun = Vasprun(zpath(vasprun_path))
/home/ywfang/.conda/envs/py39_develop/lib/python3.9/site-packages/pymatgen/io/vasp/outputs.py:287: in __init__
    self._parse(
/home/ywfang/.conda/envs/py39_develop/lib/python3.9/site-packages/pymatgen/io/vasp/outputs.py:410: in _parse
    raise exc
/home/ywfang/.conda/envs/py39_develop/lib/python3.9/site-packages/pymatgen/io/vasp/outputs.py:315: in _parse
    for _, elem in ET.iterparse(stream):
/home/ywfang/.conda/envs/py39_develop/lib/python3.9/xml/etree/ElementTree.py:1259: in iterator
    root = pullparser._close_and_return_root()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <xml.etree.ElementTree.XMLPullParser object at 0x7f08bb51d940>

    def _close_and_return_root(self):
        # iterparse needs this to set its root attribute properly :(
>       root = self._parser.close()
E       xml.etree.ElementTree.ParseError: no element found: line 1, column 0

/home/ywfang/.conda/envs/py39_develop/lib/python3.9/xml/etree/ElementTree.py:1302: ParseError

```

Thanks very much!

Andrew-S-Rosen:
@yw-fang: Thanks for your question and further work on this.

You are correct in what's happening. Since VASP isn't installed on GitHub Actions, I have "mocked" the call to VASP and instead of relying on actual generated output files, it reads in a "mock" set of files in [this line](https://github.com/Quantum-Accelerators/quacc/blob/a86cb3cecf67eb395782d42aeadbe2bf5a884fb3/tests/core/recipes/vasp_recipes/mocked/conftest.py#L12). The mocked VASP files are found in [tests/core/recipes/vasp_recipes/mocked/mocked_vasp_run](https://github.com/Quantum-Accelerators/quacc/tree/main/tests/core/recipes/vasp_recipes/mocked/mocked_vasp_run). You can see that there is a `vasprun.xml.gz` file there.

What I would suggest you do is in your test do the following. At the top of the file have

```python
from pathlib import Path

FILE_DIR = Path(__file__).parent
MOCKED_DIR = FILE_DIR / "mocked_vasp_run"
```

and then for your specific test you can do

```python
from shutil import copy

copy(MOCKED_DIR / "vasprun.xml.gz", tmp_path / "vasprun.xml.gz")
```

and now you have a `vasprun.xml.gz` file in your current directory. Alternatively, you could upload a `vasprun.xml.gz` of your own to the test suite and do the same thing, but the former should be sufficient.

yw-fang:
> @yw-fang: Thanks for your question and further work on this.
> 
> You are correct in what's happening. Since VASP isn't installed on GitHub Actions, I have "mocked" the call to VASP and instead of relying on actual generated output files, it reads in a "mock" set of files in [this line](https://github.com/Quantum-Accelerators/quacc/blob/a86cb3cecf67eb395782d42aeadbe2bf5a884fb3/tests/core/recipes/vasp_recipes/mocked/conftest.py#L12). The mocked VASP files are found in [tests/core/recipes/vasp_recipes/mocked/mocked_vasp_run](https://github.com/Quantum-Accelerators/quacc/tree/main/tests/core/recipes/vasp_recipes/mocked/mocked_vasp_run). You can see that there is a `vasprun.xml.gz` file there.
> 
> What I would suggest you do is in your test do the following. At the top of the file have
> 
> ```python
> from pathlib import Path
> 
> FILE_DIR = Path(__file__).parent
> MOCKED_DIR = FILE_DIR / "mocked_vasp_run"
> ```
> 
> and then in the specific function for your test do
> 
> ```python
> from shutil import copy
> 
> copy(MOCKED_DIR / "vasprun.xml.gz", tmp_path / "vasprun.xml.gz")
> ```
> 
> and now you have a `vasprun.xml.gz` file in your directory. Alternatively, you could upload a `vasprun.xml.gz` of your own to the test suite and do the same thing, but the former should be sufficient.

Thanks very much for your insights! At present, I have done all the unit tests for nscf_job in the test_vasp_recipes.py (I guess you'd like to carry out all unit tests relevant to VASP in this file). Thus, the modules you mentioned have been included in this file.

src/quacc/recipes/vasp/core.py
```diff
@@ -15,6 +22,104 @@
...
+    Parameters
+    ----------
+    atoms
+        Atoms object.
+    prev_dir
+        Directory (str | path) of the static_job
+    bandgap
+        The band gap of the material.
+    nbands_factor
+        A factor used to adjust NBANDS provided vasprun.xml(.gz) exists in
```
Andrew-S-Rosen, line 51: A few suggestions:
1. Let's emphasize that it's a multiplicative factor
2. Let's set the default to 1.2
3. Let's make the type `float` instead of `float | None`. If someone doesn't want to increase it, they can set it to 1.0.

src/quacc/recipes/vasp/core.py
```diff
@@ -15,6 +22,104 @@
...
+        file.startswith("vasprun.xml") for file in os.listdir(prev_dir)
+    )
+    if vasprun_exists:
+        vasprun_path = Path(prev_dir, "vasprun.xml")
+        vasprun = Vasprun(zpath(vasprun_path))
+        if nbands_factor is not None:
+            nbands = int(np.ceil(vasprun.parameters["NBANDS"] * nbands_factor))
+            updates["nbands"] = nbands
+    else:
+        logger.warning("vasprun.xml* file does not exist in the specified directory.")
```
Andrew-S-Rosen, line 96: Since the `vasprun.xml` file is always written out by VASP, perhaps it makes sense to assume it's always there. Would you agree?

src/quacc/recipes/vasp/core.py
```diff
@@ -15,6 +22,104 @@
...
+        sigma = 0.2 if bandgap < 1e-4 else 0.01
+        updates.update({"ismear": 0, "sigma": sigma})
+    else:
+        raise ValueError("Supported kpoint modes are 'uniform' and 'line' at present")
+
+    if calculate_optics:
+        updates.update({"loptics": True, "lreal": False, "cshift": 1e-5})
+
+    # integrate updates to calc_kwargs
+    calc_kwargs.update(updates)
```
Andrew-S-Rosen, line 111: These updates should be added to the `calc_defaults` instead. The reason for this is we want to always respect the user. If the user chose to do `NBANDS = 10000`, then we don't want to override their decision. In `base_fn()`, `calc_swaps` is merged with `calc_defaults`, with the former taking precedent.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
Andrew-S-Rosen, line 1: I find it helpful to also have one test where the default arguments are used, just to make sure it's not broken.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
```diff
@@ -19,6 +24,91 @@
...
+    assert output["parameters"]["icharg"] == 11
+    assert output["parameters"].get("kspacing") is None
+
+    # test nbands and log when vasprun.xml doesn't exist
+    vasprun_exists = False
+    if vasprun_exists:
+        assert "nbands" in output["parameters"]
+    else:
+        assert output["parameters"].get("nbands") is None
+        assert "vasprun.xml* file does not exist in the specified directory." in caplog.text
```
Andrew-S-Rosen, line 71: Since you are defining `vasprun_exists` as `False`, we don't need an `if`/`else` block. It is also best practice to avoid including such logic statements in tests to help with debugging.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
```diff
@@ -19,6 +24,91 @@
...
+
+
+    # test nbands when vasprun.xml exists
+    from shutil import copy
```
Andrew-S-Rosen, line 75: Import can be added to the top of the file.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
```diff
@@ -19,6 +24,91 @@
...
+        prev_dir=tmp_path,
+        nbands_factor=1.2,
+        kpoints_mode="uniform",
+        **calc_kwargs,
+    )
+    assert "nbands" in output["parameters"]
+
+    # test kpoints_mode line mode
+    kpoints_mode = "line"
+    output = nscf_job(
```
Andrew-S-Rosen, line 88: Note that at this point, the `vasprun.xml.gz` copied from before is still present in the working directory. Since there are some I/O operations being done beforehand, it might be best to split this up into 3 individual test functions (e.g. `test_nscf_job1`, `test_nscf_job2`, etc) to prevent unexpected behavior.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
```diff
@@ -19,6 +24,91 @@
...
+        kpoints_mode=kpoints_mode,
+        calculate_optics=calculate_optics,
+        **calc_kwargs,
+    )
+    assert output["parameters"]["sigma"] == 0.2
+    assert output["parameters"]["ismear"] == 0
+
+    # test invalid kpoints_mode
+    kpoints_mode = "dummy"
+    with pytest.raises(ValueError):
```
Andrew-S-Rosen, line 103: This is something I'm [still improving upon](https://github.com/Quantum-Accelerators/quacc/issues/1708), but could you include the `reason` keyword argument here so we make sure we're catching the right `ValueError`?

src/quacc/recipes/vasp/core.py
```diff
@@ -15,6 +22,104 @@
     from quacc.schemas._aliases.vasp import DoubleRelaxSchema, VaspSchema
     from quacc.utils.files import Filenames, SourceDirectory
 
+logger = logging.getLogger(__name__)
+
+
+@job
+def nscf_job(
```
Andrew-S-Rosen, line 29: Thanks! This look good. The only logic we didn't end up including the NEDOS handling, but I think we can handle that later on if there's still interest.

Andrew-S-Rosen (commented):
@yw-fang: Thank you for taking the time to address my comments. This is looking to be in great shape. I have left a few remaining (minor) comments. Once these are addressed, I will be happy to merge in the PR.

Two other points from my original review:
- Additionally, you will want to add this recipe to the [list of recipes](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/user/recipes/recipes_list.md) in the documentation.
- Finally, you may (if you wish) add yourself to the [contributors section](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/about/contributors.md) in the documentation.

yw-fang:
> @yw-fang: Thank you for taking the time to address my comments. This is looking to be in great shape. I have left a few remaining (minor) comments. Once these are addressed, I will be happy to merge in the PR.
> 
> Two other points from my original review:
> 
> * Additionally, you will want to add this recipe to the [list of recipes](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/user/recipes/recipes_list.md) in the documentation.
> * Finally, you may (if you wish) add yourself to the [contributors section](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/about/contributors.md) in the documentation.

Before going to revise the documentation, would you like us to use a more similar between different codes?
I noted that in the case of quantum espresso, former contributors used `Espresso Non-SCF `  to name the job and the function name is `non_scf_job `?
| Espresso Non-SCF         | `#!Python @job`  | [quacc.recipes.espresso.core.non_scf_job][]  

But in the present implementation of nscf_job for VASP, the name of the job is 'VASP NSCF', and the function name is 'nscf_job'? Would you like to change them into "VASP Non-SCF" and "non_scf_job" respectively?

Andrew-S-Rosen:
Thanks for bringing this up, @yw-fang! I admit that I haven't given it much thought. I suppose non_scf_job is a bit clearer than nscf, but I'll ultimately leave the decision up to you and your best judgment.

yw-fang:
> Thanks for bringing this up, @yw-fang! I admit that I haven't given it much thought. I suppose non_scf_job is a bit clearer than nscf, but I'll ultimately leave the decision up to you and your best judgment.

Thanks for the quick response. I can change the names a little bit to be closer to the implementation of recipes of Quantum Espresso. Although I prefer 'nscf' because Quantum Espresso also used it in the source code, I think keeping the same (or at least similar) style should be better if one user wants to use several different codes to compare the calculation results between them. I don't want to bother them to spend additional time changing the names of the recipes, and they can just focus their attention on the specific parameters depending on the codes. Let me know what you think!

Andrew-S-Rosen:
I'll trust your judgment on this one. 👍

yw-fang:
> I'll trust your judgment on this one. 👍

Thanks for your trust! The names are changed db3819f63f7b455abe56dbfb36f174591c34fe43 and aa4d702f6069390196f521844bc8b95b0f7d3bc8

yw-fang:
The documentation about the recipe and contribution was updated 17314b16dd8f0d13dc3a19c518a294195d15a61b

Andrew-S-Rosen:
@yw-fang: Thanks for your contribution! I have made some modifications, mainly to add some additional features. Please take a look when you get a moment and let me know what you think. If you're happy with it, then I will merge it in.

yw-fang:
> @yw-fang: Thanks for your contribution! I have made some modifications, mainly to add some additional features. Please take a look when you get a moment and let me know what you think. If you're happy with it, then I will merge it in.

Hello @Andrew-S-Rosen thanks for enhancing this feature. Almost all of them look more professional than I did. Especially I like the `test_non_scf_job3` which both considered the metal and insulator cases.

One thing I'd like to resume is the `nedos` of 5001 in the `non_scf_job`. While I agree with you that `int(np.ceil(prior_nbands * nbands_factor))` is a robust way to determine the `nedos`, it's prudent to explicitly set `nedos` to a large value (e.g. 5001) in the `calc_defaults` dictionary. This is because in our `static_jobs`, we've adjusted `nedos` to 3001. Without setting a larger default value, the parameter `nedos` derived from int(np.ceil(prior_nbands * nbands_factor)) might end up smaller than 3001 (actually I tested several cases including silicon and several oxides, and found the `nedos` were all smaller than 3001), which is not expected by us.

yw-fang:
@Andrew-S-Rosen  I noticed one of the tests by GitHub actions didn't pass. It seems that the CODECOV_TOKEN expires.

Andrew-S-Rosen:
@yw-fang good point. I was thinking about that as well. I will revert the change and have NEDOS be a sufficiently large value.

Also, not to worry about codecov. Sometimes, it just misbehaves and we have to try running it again...

yw-fang:
> @yw-fang good point. I was thinking about that as well. I will revert the change and have NEDOS be a sufficiently large value.
> 
> Also, not to worry about codecov. Sometimes, it just misbehaves and we have to try running it again...

Great! Thanks for your trust! I don't have any further comments now. Quacc and you have motivated  me to go back to read some pages about python in the latest week. Thanks!

Andrew-S-Rosen:
Merging this. Thank you again!

Issue #1852, opened 2024-03-06
Modifying the decorator of `ph_recover_job` in Espresso's `grid_phonon_flow` does not work


tomdemeyere:
### What would you like to report?

I find myself in the impossibility to run the `quacc.recipes.espresso.phonons.grid_phonon_flow` with custom decorators by doing:

``` python

grid_phonon_decorators = {
    "relax_job": job(executors=["surface"]),
    "ph_job": job(executors=["phonon"]),
}

grid_phonon_results = grid_phonon_flow(
     ...
     job_decorators=grid_phonon_decorators,
)
```
Parsl will throw: `parsl.errors.NoDataFlowKernelError: Must first load config`, other basic jobs that I redecorate manually with `redecorate` work fine, any thoughts here?

Andrew-S-Rosen:
Thank you for the report, @tomdemeyere! I haven't seen this behavior before, and it's not immediately obvious to me yet what the root cause is.

I tried a minimal example as a sanity check:

```python
import parsl
from ase.build import bulk
from quacc.recipes.emt.slabs import bulk_to_slabs_flow
from quacc import job

parsl.load()

atoms = bulk("Cu")
future = bulk_to_slabs_flow(atoms,job_decorators={"relax_job": job(), "static_job": job()})
future.result()
```

This, thankfully, works as expected. I can probably help debug further with a minimal (local) example, but at the moment the cause of the issue evades me.

tomdemeyere:
Ah! your example works on my machine as well.

Although, changing `bulk_to_slabs_flow` to the `grid_phonon_flow` still fails for me (same error as before), can you try to reproduce it?

Andrew-S-Rosen:
If you provide a minimal reproducible example, I'd be happy to run it and try to sort out the root cause. 👍

tomdemeyere:
``` python

import parsl
from ase.build import bulk
from quacc import job
from quacc.recipes.espresso.phonons import grid_phonon_flow

parsl.load()

atoms = bulk("Cu")
future = grid_phonon_flow(
    atoms,
    job_params={
        "relax_job": {"occupations": "smearing", "degauss": 0.01, "smearing": "cold"}
    },
    job_decorators={"relax_job": job(), "ph_init_job": job()},
)
future.result()

```

Should throw something like:  `TypeError: 'AppFuture' object is not callable`

(Also it seems that there is a wrong type hint in _grid_phonon_subflow?)

``` python

    @subflow
[...]

PR #1882, opened 2024-03-15
Issues with the grid_phonon_flow and custom decorators when using Parsl

tomdemeyere:
## Summary of Changes

It seems that the "job within a job" concept is not well appreciated by Parsl when using custom executors (redecorated jobs). Leading to problems as mentioned in #1852

Changing this `job` (_ph_recover_job) to a `subflow` and removing strip_decorators does make things work. We have now to understand how this can possibly affect other workflow engines.

### Checklist

### Notes

### DIFF
--- a/src/quacc/recipes/espresso/phonons.py
+++ b/src/quacc/recipes/espresso/phonons.py
@@ -20,7 +20,7 @@ from quacc.calculators.espresso.utils import grid_copy_files, grid_prepare_repr
 from quacc.recipes.espresso._base import base_fn
 from quacc.recipes.espresso.core import relax_job
 from quacc.utils.dicts import recursive_dict_merge
-from quacc.wflow_tools.customizers import customize_funcs, strip_decorator
+from quacc.wflow_tools.customizers import customize_funcs
 
 if TYPE_CHECKING:
     from typing import Any, Callable, TypedDict
@@ -360,8 +360,8 @@ def grid_phonon_flow(
         See the type-hint for the data structure.
     """
 
-    @job
-    def _ph_recover_job(grid_results: list[RunSchema]) -> RunSchema:
+    @subflow
+    def _ph_recover_subflow(grid_results: list[RunSchema]) -> RunSchema:
         prev_dirs = {}
         for result in grid_results:
             prev_dirs[result["dir_name"]] = [
@@ -371,7 +371,7 @@ def grid_phonon_flow(
                 Path("**", "wfc*.*"),
                 Path("**", "paw.txt.*"),
             ]
-        return strip_decorator(ph_recover_job)(prev_dirs)
+        return ph_recover_job(prev_dirs)
 
     @subflow
     def _grid_phonon_subflow(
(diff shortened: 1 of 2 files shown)


buildbot-princeton:

Andrew-S-Rosen:
I understand the bug now. You are updating the decorator to the `quacc.recipes.espresso.phonons.phonon_job.ph_recover_job`, but that function does not have a decorator associated with it. It is a pure function. What really needs to happen is that the decorator to `_ph_recover_job` needs to be updated.

tomdemeyere:
>> What really needs to happen is that the [decorator](https://github.com/Quantum-Accelerators/quacc/blob/89e1b9dda91fc6835a2de8f7e1dc83612b698a3b/src/quacc/recipes/espresso/phonons.py#L363) to _ph_recover_job needs to be updated.

I tried to do this, but somehow it was still throwing the same error?

``` python

    pw_job, ph_init_job, ph_job, ph_recover_job = customize_funcs(
        ["relax_job", "ph_init_job", "ph_job", "ph_recover_job"],
        [relax_job, phonon_job, phonon_job, phonon_job],
        parameters=job_params,
        decorators=job_decorators,
    )

    _ph_recover_custom = redecorate(_ph_recover_job, job_decorators.get("ph_recover_job", job()))

```

Andrew-S-Rosen:
Your example doesn't update the `ph_recover_job`.

```python
    pw_job, ph_init_job, ph_job, ph_recover_job = customize_funcs(
        ["relax_job", "ph_init_job", "ph_job", "ph_recover_job"],
        [relax_job, phonon_job, phonon_job, phonon_job],
        parameters=job_params,
        decorators=job_decorators,
    )
```

Here, you are updating the `phonon_job` function and assigning it to the variable `ph_recover_job`. You need to pass the function to decorate as the second `list`. The name in the first list is simply for user-facing querying purposes.

tomdemeyere:
It seems like the last job (recover job) is actually re-doing everything instead of reading files. This is probably due to file not being copied

I modified one test to make sure that Quacc catch this. I will fix it when I got the time

Andrew-S-Rosen:
> It seems like the last job (recover job) is actually re-doing everything instead of reading files. This is probably due to file not being copied
> 
> I modified one test to make sure that Quacc catch this. I will fix it when I got the time

Good idea for the test. Makes sense as a potential culprit.

The `@subflow` vs. `@job` issue is still on my radar. I'm hoping to do a deeper dive on your open issues around this weekend or so. 👍 Currently drowning in a sea of grant proposals, some of which would hopefully secure funds for a software engineer to help out with quacc if successful!

Andrew-S-Rosen:
@tomdemeyere: If you keep the original `@job` (i.e. don't switch to `@subflow`) and simply remove the `strip_decorator()` call, does that work? My understanding of the code suggests this would resolve your problems because you would no longer be stripping the decorator that you pre-customized. If that works, let's go with that approach.

Andrew-S-Rosen:
Actually, scratch my comment. That would require the use of the `@subflow` decorator as you implemented.

I'll go ahead and merge this then, although I may return to it in the future.

PR #1903, opened 2024-03-19
Clearer annotation of default settings

### LINKED ISSUE #1897: Pass additional kwargs to summarize_opt_run

`summarize_opt_run` in most recipes is called using the ASE dynamics dyn object and additional fields to store. Example call in the EMT recipe:
https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/emt/core.py#L97

There's no way to pass through additional options through to the summary object, for example `check_convergence` or `move_magmoms`
https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/schemas/ase.py#L135

Adding a `summarize_kwargs` parameter to the parameters of jobs everywhere would be very helpful!

Andrew-S-Rosen:
Closes #1897.


Andrew-S-Rosen:
Unclear if this will work when dispatching the function to a remote machine though...

PR #1938, opened 2024-03-22
Fix Espresso DOS flows from crashing with workflow engines

### LINKED ISSUE #1787: Espresso DOS flows likely crash when using a workflow engine

- **quacc version**: 0.6.7
- **Python version**: 3.9+

### What is the issue?

In `quacc.recipes.espresso.dos.projwfc_flow` and `quacc.recipes.espresso.dos_flow`, both perform operations on a `Future`. For instance, see below:

https://github.com/Quantum-Accelerators/quacc/blob/bca1e5a6fba604c906784bb0efac7fd4f0a7a93d/src/quacc/recipes/espresso/dos.py#L199-L204

Here, `pw_copy_files()` does some dictionary based operations on a yet-to-be-resolved `Future`. That's almost certainly going to cause a problem when executing the workflow with a workflow engine.

Ideally, if we implement #1743, we would have caught this.

I'll need to come back to this to confirm --- this is just from my reading of the code. CC @tomdemeyere.

Andrew-S-Rosen:
Closes #1787. Before, `pw_copy_files()` was being called on `Future` objects, which is not permitted. Here, I have just naively assumed we'll copy everything, and we'll let Espresso overwrite things as needed. Is this suitable or a no-go, @Nekkrad / @tomdemeyere?

### DIFF
--- a/src/quacc/recipes/espresso/_base.py
+++ b/src/quacc/recipes/espresso/_base.py
@@ -195,6 +195,8 @@ def _prepare_atoms(
         Atoms object with attached Espresso calculator.
     """
     atoms = Atoms() if atoms is None else atoms
+    calc_defaults = calc_defaults or {}
+    calc_swaps = calc_swaps or {}
 
     calc_defaults["input_data"] = Namelist(calc_defaults.get("input_data"))
     calc_swaps["input_data"] = Namelist(calc_swaps.get("input_data"))
--- a/src/quacc/recipes/espresso/bands.py
+++ b/src/quacc/recipes/espresso/bands.py
@@ -269,7 +269,7 @@ def bands_flow(
         Dictionary of results from [quacc.schemas.ase.summarize_run][].
         See the type-hint for the data structure.
     """
-    (bands_pw_job_, bands_pp_job_, fermi_surface_job_) = customize_funcs(
+    bands_pw_job_, bands_pp_job_, fermi_surface_job_ = customize_funcs(
         ["bands_pw_job", "bands_pp_job", "fermi_surface_job"],
         [bands_pw_job, bands_pp_job, fermi_surface_job],
         parameters=job_params,
--- a/src/quacc/recipes/espresso/dos.py
+++ b/src/quacc/recipes/espresso/dos.py
@@ -68,7 +68,7 @@ def dos_job(
     """
     return run_and_summarize(
         template=EspressoTemplate("dos", test_run=test_run),
-        calc_defaults={},
+        calc_defaults=None,
         calc_swaps=calc_kwargs,
         parallel_info=parallel_info,
         additional_fields={"name": "dos.x Density-of-States"},
@@ -109,7 +109,7 @@ def projwfc_job(
     """
     return run_and_summarize(
         template=EspressoTemplate("projwfc", test_run=test_run),
-        calc_defaults={},
+        calc_defaults=None,
         calc_swaps=calc_kwargs,
         parallel_info=parallel_info,
         additional_fields={"name": "projwfc.x Projects-wavefunctions"},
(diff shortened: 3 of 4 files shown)


tomdemeyere:
Copying everything should not be the way to go unfortunately, wavefunction files are not needed and might be very large in some cases.

Andrew-S-Rosen:
That's understandable. I guess I have never really been concerned about storage space on our machines.

In any case, it seems like this will be resolved shortly anyway, so it's not a big deal.

In most recipes in quacc, the currently "recommended" way that this file I/O stuff is handled in quacc is to use something like the following:

https://github.com/Quantum-Accelerators/quacc/blob/728b3765181a3838f5fa2d0d1c2746f1682a98d8/src/quacc/recipes/vasp/mp.py#L379

Here, the dictionary key is technically a `Future`, but _most_ workflow engines will resolve this when the `@job` input parameters are resolved, The exception has historically been for Parsl, sadly, although that should hopefully change imminently: https://github.com/Parsl/parsl/pull/3289. 

Taking a look at Espresso's `pw_copy_files` function, it seems like it too will function properly in such scenarios once the upstream PR is merged in Parsl. There are no operations being done on the Future: it is just being used as a key, and that key will need to be resolved when it is used in the subsequent `@job` (just like the example above). 

So, I'll try to get the Parsl PR merged and then we won't have any of these concerns. Although maybe one day I'll come up with a solution to this I/O stuff that doesn't involve passing around a dictionary with Futures in the keys...

tomdemeyere:
@Andrew-S-Rosen 

This is very much what prevents me to use Quacc for production purposes, I have access to three clusters, each of them have:

250 GB, 500 GB and 2TB of scratch space (weirdly this is inversely proportional to the cluster computational power), the first being UK's most powerful HPC. I suspect this is somewhat similar for the rest of Europe. One of my colleague have access to two HPCs in the US and I quickly understood that scratch space is not an issue over there.

I didn't raise that point, because I understand that Quacc is made for high-throughput scenario where scratch space is not an issue at all. Although I must say I recently found myself advertising Quacc during the UK CCP9 conference. PhD students and Post-Doc are definitely the most interested, not for high-throughput purposes but to make their life easier (that's why I made my first blog post without any workflow engine). In this case the disk usage is something that I know will cause them problems. Because PhDs/post-docs might be an important catalyst for Quacc's popularity, it might be advisable to give better tools to deal with the data usage. The first thing that came to my mind would be convenience functions located in quacc.calculators.espresso.utils such as `clean_dir`, which users might place in their workflow to free some space. A less espresso focused solution would be an option `MOVE_ONLY` where the data is not copied but only moved to each jobs.

I often find myself having to run calculations that are close to the scratch space total capacity, having to copy files around in this case is impossible, I can only move...

Andrew-S-Rosen:
Thank you very much for the helpful feedback! This is all useful. I also agree that the main users of quacc are/will be people who just want to streamline calculations, especially if they are approaching a code or topic they are not yet familiar with. I think the high-throughput cases will predominantly be for expert users, and that's totally fine. One shouldn't try to overcomplicate things if simple works!

I also think that Espresso is a particularly challenging case when it comes to the data handling. Very few, if any, of the other codes supported right now require the user to shuttle folders around between individual steps. At most, you have a scenario like VASP where you need to copy a charge density or wavefunction, but this is not done all the time.

Perhaps one avenue to address this in the future is to provide a common, code-agnostic suite of utilities that can somehow be prepended to or appended to each job (or a subset of jobs) in a workflow. This would give users the ability to essentially do whatever they want between job steps, one of which could be a template for file cleanup/handling. I don't quite know offhand how to do this in an ideal way, but I could see something like this being appealing. 

Definitely room for improvement here and something I will keep in mind. And thank you for a useful idea to include in my grant proposal writing :)

PR #1955, opened 2024-03-30
Better defaults for Espresso recipes

tomdemeyere:
## Summary of Changes

Following our discussions, some better default from Espresso

These parameters aim to just 'work' for any systems you can throw. These are *very* far from optimal, for optimal parameters presets are available.

### Checklist

### Notes


buildbot-princeton:

Andrew-S-Rosen:
Thanks, @tomdemeyere! This makes sense to me. One question: since there are some parameters here repeated between recipes and it's a bit bulky, do we want to factor some of them out at the top of the module like in the ONETEP recipes? Pros: Prevents code duplication; Cons: Takes another few seconds for the user to identify all the defaults.

https://github.com/Quantum-Accelerators/quacc/blob/ee3d9e79bdff93250929a0192f703e42de078399/src/quacc/recipes/onetep/core.py#L21-L28

tomdemeyere:
@Andrew-S-Rosen Ha! I am working on that right now 😅 I actually made a separated `defaults.py` in the recipe folder? I had in mind to make it easier for users to have a look at them? Your say

Andrew-S-Rosen:
I think that can make sense. My suggestion would be:
- If the defaults are only used in that module, it makes sense to have it at the top.
- If the defaults are used across modules (or likely _would_ be if new recipes were added), perhaps a dedicated `defaults.py` module makes sense. Let's call it `_defaults.py` though to indicate that this isn't a set of recipes, just like we do for `_base.py`.

But I'm flexible here.

tomdemeyere:
This should be ok to go if you are happy with it

Andrew-S-Rosen:
Thank you! Auto-merging.

Issue #1776, opened 2024-02-28
The `copy_files` keyword can fail when using the output directory from a prior `@job` with Parsl


Andrew-S-Rosen:
### Details about the quacc environment

- **quacc version**: 0.6.8
- **Python version**: 3.9+

### What is the issue?

When using the `copy_files` keyword argument and passing in something like `{directory: "*"}` where `directory` is an `AppFuture` from a prior run, Parsl won't know to autoresolve the dictionary, and a crash will occur.

Minimal examples:
- Parsl: https://github.com/Parsl/parsl/issues/3108

This will be solved upstream in Parsl via the following PR:
- https://github.com/Parsl/parsl/pull/3111

### How can we easily reproduce the issue?

Pytest examples:

Parsl:

```python
def test_copy_files(tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)
    atoms = bulk("Cu")

    @flow
    def myflow(atoms):
        result1 = relax_job(atoms)
        return relax_job(result1["atoms"], copy_files={result1["dir_name"]: "opt.*"})

    assert "atoms" in myflow(atoms).result()
```

Issue #1518, opened 2024-01-11
Add code-specific documentation, e.g. one for VASP, ORCA, etc.


Andrew-S-Rosen:
### How can we improve the documentation?

Add recipe-specific documentation beyond each docstring on how to use a given code's recipes via minimal examples. This can be pulled from the test suite.

Also includes a sample output for each recipe.

PR #1974, opened 2024-04-03
Espresso: streamline file copying / additional recipes

tomdemeyere:
## Summary of Changes

This pull request introduces several improvements and new features to Quacc Espresso, enhancing the user experience and expanding its capabilities:

**Streamlined Input/Output Naming:**

-   Quacc Espresso now enforces predefined names for **all** input and output files removing the little freedom users had for this. This ensures consistency across different calculations and reduce the potential for errors.

**Intelligent File Copying:**

- When users provide a prev_dir argument, Quacc will now determines which files to copy based on the binary and input data. This will reduce the amount of data being copied overall, I think this was one of your suggestions on a previous merge request.

**New Recipes for Electron-Phonon Calculations:**

- Two new recipes have been added to perform electron-phonon calculations and Fourier interpolation of the phonon potential. It's a little step toward future compatibility with codes such as EPW.

Some tests have been modified to be more robust, some internal functions have been modified as well.

note to self: the two recipes have to be added to the list of recipes

### Checklist

### Notes


buildbot-princeton:

Andrew-S-Rosen:
Thanks!! Will take a closer look at this in a day or two 👍 (recovering from a cold... 🤧)

src/quacc/calculators/espresso/espresso.py
```diff
@@ -71,6 +71,8 @@ def __init__(
 
         self.binary = binary
 
+        self.ase_known = self.binary in ALL_KEYS
```
Andrew-S-Rosen, line 74: Maybe change to `self.known_binary` or `self.ase_binary` or something? I find `self.ase_known` a bit cryptic. I'd also keep this as a hidden attribute with a leading `_` since this is not really something we want or expect users to query based on.

src/quacc/calculators/espresso/espresso.py
```diff
@@ -381,29 +380,29 @@ def __init__(
         self.preset = preset
         self.parallel_info = parallel_info
         self.kwargs = kwargs
-        self._user_calc_params = {}
+        self.user_calc_params = {}
```
Andrew-S-Rosen, line 383: Any reason for making this user-visible? It's not supplied by the user and is only used internally.

src/quacc/calculators/espresso/utils.py
```diff
@@ -221,20 +222,331 @@ def espresso_prepare_dir(outdir: str | Path, binary: str = "pw") -> dict[str, An
...
+        "postahc": {"input": {"ahc_dir": "ahc_dir/", "flvec": "matdyn.modes"}},
     }
 
     return outkeys.get(binary, {})
+
+
+def prepare_copy_files2(
+    parameters: dict[str, Any], binary: str = "pw"
+) -> dict[SourceDirectory, Filenames]:
+    """
```
Andrew-S-Rosen, line 267: I assume this was to replace the old one? Remember to delete the old one and rename this one when done.

src/quacc/calculators/espresso/utils.py
```diff
@@ -221,20 +222,331 @@ def espresso_prepare_dir(outdir: str | Path, binary: str = "pw") -> dict[str, An
...
+            to_copy.append(Path("pwscf.save", "charge-density.*"))
+
+        if need_wfc:
+            to_copy.append(Path("pwscf.save", "wfc*.*"))
+
+        to_copy.extend(
+            [Path("pwscf.save", "data-file-schema.*"), Path("pwscf.save", "paw.*")]
+        )
+
+    if binary == "ph":
```
Andrew-S-Rosen, line 331: Seems like these all should be `if`/`elif` blocks rather than individual `if` statements.

src/quacc/recipes/espresso/_base.py
```diff
@@ -157,14 +161,15 @@ def run_and_summarize_opt(
...
-def _prepare_atoms(
+def _prepare_calc(
     atoms: Atoms | None = None,
     preset: str | None = None,
     template: EspressoTemplate | None = None,
     profile: EspressoProfile | None = None,
     calc_defaults: dict[str, Any] | None = None,
     calc_swaps: dict[str, Any] | None = None,
     parallel_info: dict[str, Any] | None = None,
+    copy_files: SourceDirectory | dict[SourceDirectory, Filenames] | None = None,
```
Andrew-S-Rosen, line 172: Don't forget to add the docstring.

src/quacc/recipes/espresso/_base.py
```diff
@@ -218,4 +223,14 @@ def _prepare_atoms(
...
     )
 
-    return atoms
+    if isinstance(copy_files, (str, Path, list)):
+        exact_files_to_copy = prepare_copy_files(calc.user_calc_params, binary=binary)
+
+        if isinstance(copy_files, (str, Path)):
+            copy_files = {copy_files: exact_files_to_copy}
+        elif isinstance(copy_files, list):
+            copy_files = {source: exact_files_to_copy for source in copy_files}
```
Andrew-S-Rosen, line 232: This block can likely be improved without the need for so many instance checks.

Perhaps something like:

```python
if copy_files:
    if isinstance(copy_files, (str, Path)):
        copy_files = [copy_files]
    exact_files_to_copy = prepare_copy_files(calc.user_calc_params, binary=binary)
    copy_files = {source: exact_files_to_copy for source in copy_files}
```

That said, it might be helpful to not overwrite the `copy_files` variable, e.g. to return `updated_copy_files` or something similar, just for clarity.

src/quacc/recipes/espresso/bands.py
```diff
@@ -34,7 +34,7 @@ class BandsSchema(TypedDict, total=False):
 @job
 def bands_pw_job(
     atoms: Atoms,
-    copy_files: SourceDirectory | dict[SourceDirectory, Filenames],
+    copy_files: SourceDirectorySchema,
```
Andrew-S-Rosen, line 37: Personally, I'm not a fan of this type hinting approach throughout the module. It is not immediately obvious to the user what this is. Previously, we had

```python
SourceDirectory | dict[SourceDirectory, Filenames]
```

The user can likely know at a glance what a `SourceDirectory` might be or how the dictionary would be formatted, without having to dig through the codebase. However, a `SourceDirectorySchema` is quite niche. Schema also implies that it is the output of a calculation, which doesn't seem to be the case here.

Can you clarify your goal here? I can then provide a more appropriate recommendation.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -102,7 +104,7 @@ def q2r_job(
 
     Parameters
     ----------
-    prev_dir
+    copy_files
         Outdir of the previously ran ph.x calculation. This is used to copy
         the the dynamical matrix files.
```
Andrew-S-Rosen, line 109: The docstring no longer matches the parameter. Before, it was a previous directory. Now, it can be many things apparently.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -119,18 +121,9 @@ def q2r_job(
...
-    input_data = Namelist(calc_kwargs.get("input_data"))
-    input_data.to_nested(binary="q2r")
-
-    fildyn = input_data["input"].get("fildyn", "matdyn")
-
-    calc_defaults = {"input_data": {"input": {"flfrc": "q2r.fc", "fildyn": fildyn}}}
-
-    copy_files = {prev_dir: [f"{fildyn}*"]}
 
     return run_and_summarize(
```
Andrew-S-Rosen, line 125: Let's set `calc_defaults={}` just for clarity.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -151,7 +146,7 @@ def matdyn_job(
 
     Parameters
     ----------
-    prev_dir
+    copy_files
         Outdir of the previously ran q2r.x calculation. This is used to copy
         the the force constant file.
```
Andrew-S-Rosen, line 151: Same comment here about the docstring.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -168,18 +163,9 @@ def matdyn_job(
...
-    input_data = Namelist(calc_kwargs.get("input_data"))
-    input_data.to_nested(binary="matdyn")
-
-    flfrc = input_data["input"].get("flfrc", "q2r.fc")
-
-    calc_defaults = {"input_data": {"input": {"flfrc": flfrc}}}
-
-    copy_files = {prev_dir: [f"{flfrc}*"]}
 
     return run_and_summarize(
```
Andrew-S-Rosen, line 167: Same comment here about having `calc_defaults={}`.

src/quacc/runners/prep.py
```diff
@@ -77,6 +77,9 @@ def calc_setup(
         if isinstance(copy_files, (str, Path)):
             copy_files = {copy_files: "*"}
 
+        if isinstance(copy_files, list):
+            copy_files = {source: "*" for source in copy_files}
+
```
Andrew-S-Rosen, line 82: I am extremely hesitant to overload this method with more logic. I intentionally put in a fair amount of time into ensuring that things are clear for the users in terms of how `copy_files` behaves (before it had all kinds of complicated logic behind it).

If you have a list of paths and want to glob all the files in them, I would encourage you to explicitly specify that when passing in the `copy_files` keyword argument rather than have magic logic inside the function.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -463,3 +449,106 @@ def _grid_phonon_subflow(
...
+
+
+@job
+def dvscf_q2r_job(
+    copy_files: SourceDirectorySchema,
+    parallel_info: dict[str] | None = None,
+    **calc_kwargs,
+) -> RunSchema:
+    """
+    Function to carry out a basic dvscf_q2r calculation. It should allow you to
```
Andrew-S-Rosen, line 461: As a general comment, it would be helpful for users if you explain the purpose of the calculation itself. Many times throughout the modules, like here, it just says the name of the binary (e.g. `dvscf_q2r`). As a non-Espresso user, I don't know what this is, so I have no idea if the function is one I might want to use.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -463,3 +449,106 @@ def _grid_phonon_subflow(
...
+    use all the features of the dvscf_q2r binary which does not have an official
+    documentation. Here are the available keywords.
+
+    WARNING: This binary is not supported by ASE, the input_data must be in nested format.
```
Andrew-S-Rosen, line 465: Perhaps this is something then that should be added upstream to ASE?

src/quacc/recipes/espresso/phonons.py
```diff
@@ -463,3 +449,106 @@ def _grid_phonon_subflow(
...
+
+
+@job
+def postahc_job(
+    copy_files: SourceDirectorySchema,
+    parallel_info: dict[str] | None = None,
+    **calc_kwargs,
+) -> RunSchema:
+    """
+    Function to carry out a basic postahc calculation. It should allow you to
```
Andrew-S-Rosen, line 524: Same comment here about the description of the function.

Also, as you already commented, you'll need to update the recipe list as well in the docs.

Andrew-S-Rosen (commented):
Thank you for your contribution, @tomdemeyere. Overall, this looks good. The major comment I have is in relation to the `SourceDirectorySchema` and the updated `calc_setup()` behavior. We need to be very careful about ensuring that things are easily understandable for users without overloading things behind the scenes. I tried to outline what I mean with my comments below. Let me know if you'd like to discuss this further.

As a general comment, for any future contributions, please try to keep the PR somewhat focused on a specific change. Here, we have changes to file copying behavior, to the Espresso calculator, and the addition of two (largely unrelated) jobs. In principle, each of these things would be better off as individual PRs. Even if they rely on one another to some degree, that is okay --- you can number your PRs in the order they would need to be merged/reviewed. This helps a lot in the review process and if we ever need to revert something later on. It also ensures that if there is a point of contention or blocker that the discussion doesn't hold up all your work.

src/quacc/calculators/espresso/espresso.py
tomdemeyere: Will do!

src/quacc/calculators/espresso/espresso.py
tomdemeyere: It is accessed in _base.py, DeepSource was complaining that a "_" was accessed, the only reason I changed it

src/quacc/calculators/espresso/utils.py
tomdemeyere: That was meant to be deleted sorry about that

src/quacc/calculators/espresso/espresso.py
Andrew-S-Rosen: Thanks, I'll add an override in deepsource.

src/quacc/runners/prep.py
tomdemeyere: So here is the problematic:

The new "intelligent" file copying only operate if the user send a `SourceDirectory` in `copy_files`, if the user send a dict themselves we assume that they know what they are doing and the intelligent file copying will be skipped (this is also useful for specific recipe like the `grid_phonon_flow` where file handling is ... a nightmare).

The thing is: sometimes users will want to copy files from multiple `SourceDirectory` ... and still want the "intelligent" file copying to operate. Hence why I wanted to add a list.

I guess the solution here is to require the user to send a dict with a specific value? Like "automatic"? I am not really convinced. Anyway, I must say it: thanks for keeping up with espresso's file mechanics 😅 this is highly appreciated

Andrew-S-Rosen: The main issue with the currently proposed mechanism is that, while it makes sense within the context of Espresso, it is not a sufficiently obvious solution in general. For instance, if a user passes in `["dir1", "dir2"]`, they might naively expect `dir1` and `dir2` to be copied to the working directory rather than the files from within `dir1` and `dir2` copied without the parent folders. This ambiguity between "do things unpack or not unpack" was one of the key motivations for the original refactor, which is why I'm being a bit picky on this.

> The thing is: sometimes users will want to copy files from multiple SourceDirectory ... and still want the "intelligent" file copying to operate. Hence why I wanted to add a list.

What would limit the user from doing `copy_files={"dir1": "*", "dir2": "*"}` in this scenario?

> I guess the solution here is to require the user to send a dict with a specific value? Like "automatic"? I am not really convinced. 

I am also not convinced. Basically, if it "feels" like a hack or magic, then it's a no-go and this would certainly qualify. I suppose I first need to understand a bit more about your motivation here. It is not yet clear to me why the existing behavior of the function does not meet the needs of users or developers. I'm not saying there aren't reasons, but it's just not apparent to _me_ what those are yet. Once I understand that more, I can help brainstorm a possible alternative.

> Anyway, I must say it: thanks for keeping up with espresso's file mechanics 😅 this is highly appreciated

And thank you for your patience with this! I think, on the whole, these kinds of discussions have improved the codebase even if it does mean we have to go back-and-forth a bit.

Andrew-S-Rosen: I should also mention that if the main driver behind all of this is https://github.com/Quantum-Accelerators/quacc/issues/1776, then this is already solved in https://github.com/Quantum-Accelerators/quacc/pull/1942. It just requires Ben to merge the PR upstream, and a simple ping on Slack would likely be sufficient.

tomdemeyere: >> I should also mention that if the main driver behind all of this is https://github.com/Quantum-Accelerators/quacc/issues/1776

Not at all, the aim here is really to streamline everything.

>> What would limit the user from doing copy_files={"dir1": "*", "dir2": "*"} in this scenario?

Because the user provides a dict and the "intelligent" file copying would not operate, this would simply copy everything. Only sending a raw string or Path works:

``` python
if isinstance(copy_files, (str, Path, list)):
      exact_files_to_copy = prepare_copy_files(calc.user_calc_params, binary=binary)
```

Andrew-S-Rosen: @tomdemeyere: Thank you for your reply. I think I'm getting closer to understanding the situation.

The type hint for this function reads

```python
copy_files: SourceDirectory | dict[SourceDirectory, Filenames] | None = None
```

If I understand correctly, your comment is that users may need to copy files from _multiple_ `SourceDirectory`s, which is not currently possible. Your proposal is to support the behavior when a user passes in a `list[SourceDirectory]`.

If so, please ensure you update the type hint

```python
copy_files: SourceDirectory | list[SourceDirectory] | dict[SourceDirectory, Filenames] | None = None
```

Then we take the following block:

```python
        if isinstance(copy_files, (str, Path)):
            copy_files = {copy_files: "*"}

        if isinstance(copy_files, list):
            copy_files = {source: "*" for source in copy_files}
```

and please make sure to do an `elif` instead of `if` on the second check.

This would be a fine addition. Apologies. It was difficult to follow everything being done in such a large PR. This is the kind of scenario where it would be a lot clearer if a dedicated PR were made just to add support for updating the file copying mechanism, where the PR summary would summarize the need.

src/quacc/utils/files.py
```diff
@@ -23,6 +23,10 @@
     Filenames = str | Path | list[str | Path]
     SourceDirectory = str | Path
 
+    SourceDirectorySchema = (
+        SourceDirectory | dict[str, SourceDirectory] | list[SourceDirectory]
+    )
+
```
Andrew-S-Rosen, line 29: @tomdemeyere: Throughout your update to this PR, let's not lump the type hints together like this, as it is very confusing. 

A "schema" means something quite specific --- it is a structured output (typically JSON style) with various attributes and corresponding values.

Since this is a type hint that is visible to end-users through the recipes themselves, we want to be as clear as possible. In this case, explicitly enumerating the three options is better in my opinion.

src/quacc/recipes/espresso/bands.py
Andrew-S-Rosen: @tomdemeyere: I think this was a major source of my confusion for several reasons.

You have defined this `SourceDirectorySchema` as

```python
SourceDirectory | dict[str, SourceDirectory] | list[SourceDirectory]
```

However, this doesn't seem right now that I read this closer. The type hint seems like it should _actually_ be

```python
SourceDirectory | list[SourceDirectory] | dict[SourceDirectory, Filenames]
```

Is that correct? Note how there is no `dict[str, SourceDirectory]`. I do not know what that would be.

Assuming I am correct here, please use the `SourceDirectory | list[SourceDirectory] | dict[SourceDirectory, Filenames]` type hint throughout. Collapsing it all to one type hint anyway is simply more confusing to users.

src/quacc/runners/prep.py
```diff
@@ -77,6 +77,9 @@ def calc_setup(
         if isinstance(copy_files, (str, Path)):
             copy_files = {copy_files: "*"}
 
+        if isinstance(copy_files, list):
+            copy_files = {source: "*" for source in copy_files}
```
Andrew-S-Rosen, line 81: This new behavior requires a dedicated unit test.

tomdemeyere: Worth noting: even without these lines in `prep.py` espresso recipes would work fine because we transform the `list[SourceDirectory]` into a `dict[SourceDirectory, Filenames]` in `quacc.recipes.espresso._base._prepare_calc`

I added it here "just in case".

src/quacc/utils/files.py
tomdemeyere: You are right it does not make sense, I don't know what I did here... I should probably take more breaks 😅

Andrew-S-Rosen: Not me looking at the timezone difference right now... 👀

tomdemeyere: Ah yes! this happens from time to time 💀

tests/core/recipes/espresso_recipes/test_phonons.py
```diff
@@ -429,9 +433,11 @@ def test_phonon_induced_renormalization(tmp_path, monkeypatch, ESPRESSO_PARALLEL
...
         }
     }
-    c_ph_results = phonon_job(
-        c_scf_results["dir_name"], **c_ph_params, parallel_info=ESPRESSO_PARALLEL_INFO
-    )
+
+    with caplog.at_level(logging.WARNING):
+        c_ph_results = phonon_job(
+            c_scf_results["dir_name"], **c_ph_params, parallel_info=ESPRESSO_PARALLEL_INFO
+        )
```
Andrew-S-Rosen: You also have to add an assertion:

e.g.

```python
assert "Blahblah blah " in caplog.text
```

tomdemeyere:
I didn't generalize the `list[SourceDirectory]` behaviour, for now this is espresso only

I added some general purpose espresso documentation but I don't it is in the right place, (it is in quacc.calculators.espresso.espresso.Espresso for now)

Other than that, this should be ok to go if you are happy with it

src/quacc/calculators/espresso/espresso.py
```diff
@@ -326,10 +325,20 @@ def _sanity_checks(self, parameters: dict[str, Any]) -> None:
...
-    Templates are used to set the binary and input/output file names.
+    Instruction for the impatient:
+
+    1. If you are used to ASE, you can send the same parameter to Quacc's jobs `input_data`, `kpts` ...
+    2. In Quacc each job has its own running directory, keep this in mind when creating workflows.
+    You will often have to use the `copy_files` parameter to copy files from one job to another.
+    You will often want to pass only the previous `dir_name` to copy_file. This interface will
+    then take care of copying the necessary files.
+    3. By default all `pw.x` jobs will use the SSSP pseudopotentials library. This is specified in the
+    `preset` parameter.
```
Andrew-S-Rosen, line 341: Personally, I would suggest removing this information because it is not related to the calculator specifically. For instance, one could import the `Espresso` class as a standalone utility, but Points 2 and 3 would not be relevant. It is also not particularly accessible to quacc end-users because they are likely to just import the recipes themselves. We should be asking ourselves the question: are the recipes themselves clear enough to use? If not, then we should identify how to clarify _those_ docstrings instead.

src/quacc/recipes/espresso/_base.py
```diff
@@ -76,9 +83,13 @@ def run_and_summarize(
         parallel_info=parallel_info,
     )
 
+    updated_copy_files = _prepare_copy(
+        copy_files, atoms.calc._user_calc_params, atoms.calc.template.binary
+    )
```
Andrew-S-Rosen, line 88: The function takes keyword arguments, so we should specify that rather than supplying them as keyword-less positional arguments.

src/quacc/recipes/espresso/_base.py
```diff
@@ -148,9 +164,15 @@ def run_and_summarize_opt(
...
```
Andrew-S-Rosen, line 169: The function takes keyword arguments, so we should specify that rather than supplying them as keyword-less positional arguments.

src/quacc/recipes/espresso/_base.py
```diff
@@ -209,7 +231,7 @@ def _prepare_atoms(
 
     calc_flags = recursive_dict_merge(calc_defaults, calc_swaps)
 
-    atoms.calc = Espresso(
+    calc = Espresso(
```
Andrew-S-Rosen, line 234: Not sure there was a need to change it from `atoms.calc = ` but it doesn't matter much I suppose.

src/quacc/recipes/espresso/_base.py
```diff
@@ -218,4 +240,44 @@ def _prepare_atoms(
...
+def _prepare_copy(
+    copy_files: (
+        SourceDirectory
+        | list[SourceDirectory]
+        | dict[SourceDirectory, Filenames]
+        | None
+    ) = None,
+    calc_params: dict[str, Any] | None = None,
+    binary: str = "pw",
+) -> dict[SourceDirectory, Filenames]:
```
Andrew-S-Rosen, line 257: If `None` is provided for `copy_files`, it seems like this function can return `None` but that is not reflected in the type hint or docstring.

src/quacc/recipes/espresso/bands.py
```diff
@@ -43,14 +45,22 @@ def bands_pw_job(
     **calc_kwargs,
 ) -> RunSchema:
     """
-    Function to carry out a basic bands calculation with pw.x.
+    Function to carry out a basic bands structure calculation with pw.x.
+
+    First perform a normal SCF calculation as above; then use this job
+    if you are interested in calculating only the Kohn-Sham states for the given
+    set of k-point
```
Andrew-S-Rosen, line 52: "k-point" --> "k-points"

src/quacc/recipes/espresso/bands.py
```diff
@@ -43,14 +45,22 @@ def bands_pw_job(
     **calc_kwargs,
 ) -> RunSchema:
     """
-    Function to carry out a basic bands calculation with pw.x.
+    Function to carry out a basic bands structure calculation with pw.x.
+
+    First perform a normal SCF calculation as above; then use this job
```
Andrew-S-Rosen, line 50: Let's refer to the specific function since if someone views the docstring in something like VS code via a hover-over, it might not be visible immediately. Also, there is nothing above :)

You can do something like
```
[quacc.recipes.espresso.core.static_job][]
```

to have it automatically hyperlink in the built docs.

src/quacc/recipes/espresso/bands.py
```diff
@@ -107,20 +117,28 @@ def bands_pw_job(
...
+        SourceDirectory | list[SourceDirectory] | dict[SourceDirectory, Filenames]
+    ),
     parallel_info: dict[str] | None = None,
     test_run: bool = False,
     **calc_kwargs,
 ) -> RunSchema:
     """
     Function to re-order bands and computes bands-related properties with bands.x.
+    This allows to get the bands structure in a more readable way. This requires a
+    previous `bands_pw_job` calculation.
```
Andrew-S-Rosen, line 130: Let's use `[bands_pw_job][]` (no tick marks needed) to automatically hyperlink.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -39,28 +39,39 @@ class PhononDosSchema(TypedDict):
...
+        which files have to be copied over by looking at the binary and `input_data`.
+        If a dict is provided, the mode is manual, keys are source directories and values
+        are relative path to files or directories to copy. Glob patterns are supported.
     parallel_info
         Dictionary containing information about the parallelization of the
         calculation. See the ASE documentation for more information.
     test_run
         If True, a test run is performed to check that the calculation input_data is correct or
         to generate some files/info if needed.
+    is_phcg
```
Andrew-S-Rosen, line 72: Is there anything more user-friendly we can have here, or is this known by the Espresso users? I don't know what "phcg" is even based on the docstring. 

Edit: Ah, I see it's a binary. It seems to me that "is" may not be appropriate here. Perhaps `use_phcg` makes more sense?

src/quacc/recipes/espresso/phonons.py
```diff
@@ -463,3 +480,135 @@ def _grid_phonon_subflow(
...
+@job
+def dvscf_q2r_job(
+    copy_files: (
+        SourceDirectory | list[SourceDirectory] | dict[SourceDirectory, Filenames]
+    ),
+    parallel_info: dict[str] | None = None,
+    **calc_kwargs,
+) -> RunSchema:
+    """
+    Function to carry out a basic dvscf_q2r calculation. It should allow you to
```
Andrew-S-Rosen, line 494: Not immediately clear based on this sentence what a dvscf_q2r calculation actually is.

src/quacc/recipes/espresso/phonons.py
```diff
@@ -463,3 +480,135 @@ def _grid_phonon_subflow(
...
+        [quacc.calculators.espresso.espresso.Espresso][] for more information.
+
+    Returns
+    -------
+    RunSchema
+        Dictionary of results from [quacc.schemas.ase.summarize_run][].
+        See the type-hint for the data structure.
+    """
+
+    calc_defaults = {}
```
Andrew-S-Rosen, line 546: You can just set `calc_defaults={}` in the `run_and_summarize()` function here and elsewhere IMO.

Andrew-S-Rosen (commented):
Thank you very much for your contribution, @tomdemeyere! This is all looks great to me. I have left some minor comments below, all of which are trivial and should not take more than a few minutes. After they are addressed, I'll be happy to merge!

> I didn't generalize the list[SourceDirectory] behaviour, for now this is espresso only

Sounds good.

> I added some general purpose espresso documentation but I don't it is in the right place, (it is in quacc.calculators.espresso.espresso.Espresso for now)

Agreed, as commented below, that's probably not the right place. We should either clarify the recipe docstrings themselves (if needed) or, if it's meant to be more of a tutorial/guide, it should be a dedicated section in the quacc [documentation](https://github.com/Quantum-Accelerators/quacc/tree/main/docs). We don't currently have tutorials for the different codes (https://github.com/Quantum-Accelerators/quacc/issues/1518), but there's nothing stopping us from adding a mini Espresso section if we feel like it. That would probably be a bit more detailed than the text provided in this PR though. I would say let's remove it for now.

> Other than that, this should be ok to go if you are happy with it

👍

Andrew-S-Rosen:
Also, if you can fix the merge conflict that'd be great, but I'm happy to do so if that's too bothersome.

Andrew-S-Rosen:
Good to go? (The Deepsource X will resolve once this is merged)

tomdemeyere:
Yes good to go

Andrew-S-Rosen:
Beautiful. Thank you very much!

PR #2028, opened 2024-04-21
Some modifications to the Phonopy flows

tomdemeyere:
## Summary of Changes

1. Optimize the `phonon_flow` by removing of the call to `generate_displacements`. Currently, this method is called twice for each phonon_flow run, which can be time-consuming for large supercells, especially when magnetism is involved (10+ minutes).

2. Implement the ability to fix certain atoms in the supercell, which is a frequently requested feature for Phonopy but often denied due to it being a crude approximation. It can be manually implemented in the `phonon_flow` to reduce the number of displacements and computational cost. However, it is important to note that fixing atoms is an approximation and may affect the accuracy of the results. 

3. Introduce additional keywords to enhance customization options for the flow. Although I am not sure about this one because this is getting very messy... EDIT: It might be worth creating a phonopy_post_processing job? which would read the force_constant file previously written by a phonopy_flow?

The code is not ready at all, docstrings are missing etc...

### Checklist

### Notes


buildbot-princeton:

Andrew-S-Rosen:
Hi @tomdemeyere! Thank you again for your contributions. It would likely be very helpful for me if you could, at some point, split up this PR into multiple components. For instance, make a PR addressing Point 1 and then have a branch off that PR to address Point 2, doing so in the order they should be reviewed.

I would love to solve the first issue, but I was not able to identify a mechanism to do so. I'd be interested to see how you accomplish this.

For fixing of atoms, I am not opposed to this. Perhaps there is a way to have the user supply an `Atoms` object with `FixAtoms` constraints without us having to add a keyword for it? I'll have to look into this more.

I see the messiness going on in Point 3. Perhaps let's tackle that after Point 1 and 2.

tomdemeyere:
On it!

PR #2032, opened 2024-04-22
Phonopy: remove one `get_phonopy`

tomdemeyere:
## Summary of Changes

This will greatly reduce the computational cost in some cases

### Checklist

### Notes

### DIFF
--- a/src/quacc/recipes/common/phonons.py
+++ b/src/quacc/recipes/common/phonons.py
@@ -80,39 +80,36 @@ def phonon_flow(
         Dictionary of results from [quacc.schemas.phonons.summarize_phonopy][]
     """
 
+    phonon = get_phonopy(
+        atoms,
+        min_lengths=min_lengths,
+        supercell_matrix=supercell_matrix,
+        symprec=symprec,
+        displacement=displacement,
+        phonopy_kwargs=phonopy_kwargs,
+    )
+
+    supercells = [
+        phonopy_atoms_to_ase_atoms(s) for s in phonon.supercells_with_displacements
+    ]
+
     @subflow
-    def _get_forces_subflow(atoms: Atoms) -> list[dict]:
-        phonon = get_phonopy(
-            atoms,
-            min_lengths=min_lengths,
-            supercell_matrix=supercell_matrix,
-            symprec=symprec,
-            displacement=displacement,
-            phonopy_kwargs=phonopy_kwargs,
-        )
-        supercells = [
-            phonopy_atoms_to_ase_atoms(s) for s in phonon.supercells_with_displacements
-        ]
+    def _get_forces_subflow(supercells: list[Atoms]) -> list[dict]:
         return [
             force_job(supercell) for supercell in supercells if supercell is not None
         ]
 
     @job
     def _thermo_job(atoms: Atoms, force_job_results: list[dict]) -> PhononSchema:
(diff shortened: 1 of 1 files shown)


buildbot-princeton:

Andrew-S-Rosen:
Thanks for the PR. Unfortunately, this logic will not work as intended. The `Atoms` object being passed in via the `atoms` positional argument to `phonon_flow()` may be a `Future`, and this `Future` cannot work with `get_phonopy()`. In general, `@flow` patterns should have essentially no logic beyond stitching together the inputs and outputs of `@job`, `@subflow`, and/or `@flow` patterns (aside from simple things like indexing). The `@flow` is just meant to construct the directed acyclic graph of jobs. That's the reason for the pre-existing double calls to `get_phonopy()`.

If you can find a workaround that avoids the issue here, I am all ears! I was not able to surmise one.

tomdemeyere:
In French we would call that a "dumb but disciplined" solution.

Not very proud of it, I will let you judge if this is worth it

Andrew-S-Rosen:
Thanks! To be honest, this is probably the "right" way to be doing this. I think I was originally overly obsessed with reducing the number of `@job` calls for users that dispatch one Slurm job per `@job`, but that resulted in this silliness.

I'll merge this later. The only change we need to make is to ensure that the `@flow`s that call this common flow can modify the new job decorator you added.

tomdemeyere:
Ah! I didn't think about this in the slurm jobs perspective! 

In practice wouldn't it be more practical to just stop and resolve? In practice this would do the same thing without the need to launch an additional slurm job?

Andrew-S-Rosen:
> In practice wouldn't it be more practical to just stop and resolve? This would do the same thing without the need to launch an additional slurm job?

Indeed, that would be the ideal approach (then your original implementation would work fine). It opens up the following question though: what if you wanted to make a `@flow` packaged with quacc that does two steps, some arbitrary `@job` whose output is fed to the `phonon_flow`. There is no universal mechanism to block/resolve across workflow engines. If quacc as a package were just based on parsl, you'd call `.result()`, but of course calling `.result()` with no workflow engine active would cause a problem as a counterpoint. This isn't an issue if you construct some custom workflow yourself in a Python script, but it's a concern from a package management standpoint. I guess, however, there are two ways around this issue: 1) if this use case came up, `phonon_flow` could be made a `@subflow` since that's arguably what it'd be used for; or 2) we could add a custom `resolve()` function that, based on the workflow engine, would stop and resolve with the appropriate syntax. Both seem readily solvable.

In writing this out, I now feel like your original approach may be ideal and we should just force the user or developer to resolve. After all, the type hint for the positional argument is an `Atoms`, not a `Future`. 

If you agree, feel free to revert to your initial commit and remove the following tests:
- https://github.com/Quantum-Accelerators/quacc/blob/dc4fdc372ae6cce09f489e63bec0b7734dbcb80a/tests/parsl/test_emt_recipes.py#L59-L69
- https://github.com/Quantum-Accelerators/quacc/blob/dc4fdc372ae6cce09f489e63bec0b7734dbcb80a/tests/dask/test_emt_recipes.py#L58-L68

I think it should only be those two that need to be removed.

Andrew-S-Rosen:
Thanks for the suggestions! I'll merge the other PR :)

Seems tough to adjust the thermo job but am always open to suggestions.

PR #2051, opened 2024-04-24
Deprecate "mace" mlp recipes (replaced with "mace-mp-0")

### LINKED ISSUE #1972: Deprecate the `mace` method and replace it with `mace-mp-0` in the MLP recipes

Right now, the MLP recipes support the "mace" method type, which will return the MACE-MP-0 foundation model calculator. However, it would be best if we changed this to "mace-mp-0" to better capture the nature of the model and allow for versioning in the future. Since people are using the function, we should do a true deprecation warning that is active for a little while before removing it entirely.

jackpm14:
## Summary of Changes

Closes #1972 by adding support for 'mace-mp-0' and adding a deprecated warning for 'mace' in the mlp recipe. 

Adding 'mace-mp-0':
- Added as a Literal in the arguments for static_job() and relax_job() in core.py, phonon_flow() in phonons.py, and pick_calculator() in _base.py
- Added an elif conditional for 'mace-mp-0' in pick_calculator() in _base.py 
- Added 'mace-mp-0 as a Literal for tests in test_core_recipes.py in mlp_recipes test folder.

Deprecating 'mace':
- Added 'import deprecated from monty.dev' in _base.py
- Updated pick_calculator() docstring with deprecation warning according to NumPy docstring standard in _base.py
- Added mace_warning() with 'deprecated' decorator invoked in pick_calculator() in _base.py

Additional:
- Changed 'from ase.calculator.calculator' to 'from ase.calculators.calculator' in _base.py in mlp recipe to match ase documentation and other recipes.

### Checklist

- [X] I have read the ["Guidelines" section](https://quantum-accelerators.github.io/quacc/dev/contributing.html#guidelines) of the contributing guide. Don't lie! 😉
- [X] I have added relevant, comprehensive [unit tests](https://quantum-accelerators.github.io/quacc/dev/contributing.html#unit-tests).

### DIFF
--- a/src/quacc/recipes/mlp/_base.py
+++ b/src/quacc/recipes/mlp/_base.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 from functools import lru_cache
 from typing import TYPE_CHECKING
+from monty.dev import deprecated
 
 if TYPE_CHECKING:
     from typing import Literal
@@ -16,10 +17,14 @@ logger = logging.getLogger(__name__)
 
 @lru_cache
 def pick_calculator(
-    method: Literal["mace", "m3gnet", "chgnet"], **kwargs
+    method: Literal["mace-mp-0", "m3gnet", "chgnet", "mace"], **kwargs
 ) -> Calculator:
     """
     Adapted from `matcalc.util.get_universal_calculator`.
+    .. deprecated:: 0.7.6
+            method `mace` will be removed in a later version, it is replaced by 'mace-mp-0'
+            which more accurately reflects the nature of the model and allow for versioning
+            in the future.
 
     Parameters
     ----------
@@ -56,9 +61,19 @@ def pick_calculator(
 
         calc = CHGNetCalculator(**kwargs)
 
+    elif method.lower() == "mace-mp-0":
+        from mace import __version__
+        from mace.calculators import mace_mp
+
+        if "default_dtype" not in kwargs:
+            kwargs["default_dtype"] = "float64"
+        calc = mace_mp(**kwargs)
+    
     elif method.lower() == "mace":
         from mace import __version__
         from mace.calculators import mace_mp
+
+        deprecated("DEPRECATION WARNING: 'mace' has been deprecated, 'mace-mp-0' should be used instead")
 
         if "default_dtype" not in kwargs:
             kwargs["default_dtype"] = "float64"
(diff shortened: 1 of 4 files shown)


buildbot-princeton:

src/quacc/recipes/mlp/_base.py
```diff
@@ -56,10 +62,20 @@ def pick_calculator(
...
+        calc = mace_mp(**kwargs)
+
     elif method.lower() == "mace":
         from mace import __version__
         from mace.calculators import mace_mp
 
+        mace_warning()
+
         if "default_dtype" not in kwargs:
             kwargs["default_dtype"] = "float64"
```
Andrew-S-Rosen, line 80: You can condense this to

```python
elif method.lower() in ["mace", "mace-mp-0"]:
    from mace import __version__
    from mace.calculators import mace_mp


    if method.lower() == "mace":
        mace_warning()

    ...
```
to reduce code duplication.

src/quacc/recipes/mlp/_base.py
```diff
@@ -56,10 +62,20 @@ def pick_calculator(
...
+
+        if "default_dtype" not in kwargs:
+            kwargs["default_dtype"] = "float64"
+        calc = mace_mp(**kwargs)
+
     elif method.lower() == "mace":
         from mace import __version__
         from mace.calculators import mace_mp
 
+        mace_warning()
```
Andrew-S-Rosen, line 77: Ah, while I originally said we should use `monty.dev.requires`, that function (in hindsight) really is best suited as a decorator on the parent recipe. Since that doesn't make much sense here, I suggest just logging a warning with the `logging` utility instead of calling a custom `mace_warning()` function.

Andrew-S-Rosen (commented):
Thank you very much for your contribution, @jackpm14! I have left two minor comments; otherwise, this is pretty much ready to go!

Thank you for catching the type hint error too :)

src/quacc/recipes/mlp/_base.py
jackpm14: Not sure what I was thinking there, sounds good!

Andrew-S-Rosen:
Thank you, @jackpm14! I'm setting this to auto-merge. 👍

Andrew-S-Rosen:
Seems like there's a temporary bug with GitHub Actions. I'll wait a bit and try again.

PR #2058, opened 2024-04-25
Add support for function hooks in the optimization routines

Andrew-S-Rosen:
## Summary of Changes

@tomdemeyere: I wonder if something like this would enable you to do optimizer "patching" (https://github.com/Quantum-Accelerators/quacc/discussions/1556) without introducing lots of complexity to quacc. I didn't really give this much thought so maybe I'm completely wrong.


tomdemeyere:
@Andrew-S-Rosen
This is exciting! I think we're on the same page, but I have a few questions to clarify:

- If we want to use this to, for example, change the temperature linearly during an MD run, do we need to provide all the temperatures directly inside the function that modifies opt?
``` python
def my_temperature_changer(opt):
opt.set_temperatures(...)
```

Essentially, I'm wondering if we can modify the attributes of the dyn object during the dynamics. I believe that's the case, but I just want to confirm.

- If my understanding above is correct, then the main difference compared to my original idea is that "proper patching" would provide more flexibility regarding when things run. We could execute code at the beginning or end of a step, at the very start of the simulation, or only upon convergence. Additionally, we could add custom actions whenever a specific method is called, offering even greater flexibility.
- 
> without introducing lots of complexity to quacc

If you're referring to complexity on the code side, then absolutely, this option is preferable.

In all fairness, users can achieve all of this by superclassing the desired dynamics class themselves. The real question is: should quacc help this process (like providing ready-made `fn_hook` that performs specific task)? If the answer is yes, then the method showcased here takes a small step in that direction, while full patching goes all-in by providing complete control,  with the tradeoff of increased code complexity.

Andrew-S-Rosen:
> Essentially, I'm wondering if we can modify the attributes of the dyn object during the dynamics. I believe that's the case, but I just want to confirm.

That was indeed my thinking --- that by passing the `dyn` object, the user has the ability to do whatever they want to it (including accessing/modifying attributes mid-loop). I have not personally tested this, but I can't see why that wouldn't be the case.

> If my understanding above is correct, then the main difference compared to my original idea is that "proper patching" would provide more flexibility regarding when things run. We could execute code at the beginning or end of a step, at the very start of the simulation, or only upon convergence. Additionally, we could add custom actions whenever a specific method is called, offering even greater flexibility.

Yes, here we are basically providing the user with the ability to do "whatever they want" at the end of the iteration. They could, for instance, kill the optimization entirely by toggling whether it is converged or not manually. It would not be difficult to split this into a `pre_fn_hook` and `post_fn_hook` so that users can have two separate function calls at start and end of the loop. The logic is largely the same.

> In all fairness, users can achieve all of this by superclassing the desired dynamics class themselves.

This is true. One could make their own class with its own custom logic and pass that in. That said, if someone wanted to do something like "upload to the cloud after every step" it seems like it might be a bit unnecessary to have the user make a custom ASE optimizer class just for this when in reality what they want is to call a simple function they have available.

>  The real question is: should quacc help this process (like providing ready-made fn_hook that performs specific task)? If the answer is yes, then the method showcased here takes a small step in that direction, while full patching goes all-in by providing complete control, with the tradeoff of increased code complexity.

At minimum, we can give users the ability to have that flexibility. If there are a set of function hooks that are likely to be general-purpose or necessary to carry out specific recipes, I think it would be potentially worth supporting. We could have a `quacc/runners/fn_hooks.py` module to store pre-made callables, for instance.

There are likely pros/cons compared to https://github.com/Quantum-Accelerators/quacc/discussions/1556, but one reason why I have proposed this approach is that it is a very simple entry point with a lot of flexibility. One does not necessarily need to know anything about an ASE dynamics class to inject logic.

tomdemeyere:
> They could, for instance, kill the optimization entirely by toggling whether it is converged or not manually. It would not be difficult to split this into a pre_fn_hook and post_fn_hook

That's pretty much it, for example changing temperature would need to be *before* every step.

I would be happy to provide some pre-made functions once you are done working on it.

Andrew-S-Rosen:
Cool! I'll modify this PR to have a separate pre and post hook then. I think that should cover the vast majority of use cases. And I'll fix the remaining test failure because apparently SciPy optimizers do not play nicely with `irun`.

Andrew-S-Rosen:
Hmmm... it is not immediately obvious to me how to get it to activate a function _before_ each step. 🤔 I'll merge this as-is for now. Happy to modify later --- it's experimental for now.

tomdemeyere:
> Hmmm... it is not immediately obvious to me how to get it to activate a function before each step. 🤔 I'll merge this as-is for now. Happy to modify later --- it's experimental for now.

Yes, I was wondering how you were going to do it 😅. It does not *really* matter if you do it before or after, you can write a function that changes the temperature for the **next** step but conceptually for users that's not ideal to explain.

In any case I will probably do the same kind of "demo" PR for my proposal after the MD PR is merged. The way it was shown in https://github.com/Quantum-Accelerators/quacc/discussions/1556 was fairly complex, but the aim is that users do not have to deal with these patches, just import them and provide parameters. Each of them would have their own docstring, and with appropriate documentation I suspect this will not be too complicated even for beginners.

PR #2105, opened 2024-05-09
Remove redundant call to `get_phonopy` and fix bug in common phonon subflow

Andrew-S-Rosen:
This PR removes the redundant call to `get_phonopy()` in `quacc.recipes.common.phonons`. It also fixes a bug (?) where the `Phonopy` object was constructed based on the _unrelaxed_ structure rather than the relaxed structure.

### DIFF
--- a/src/quacc/recipes/common/phonons.py
+++ b/src/quacc/recipes/common/phonons.py
@@ -21,8 +21,10 @@ if TYPE_CHECKING:
 
     from quacc import Job
     from quacc.schemas._aliases.phonons import PhononSchema
+
     if has_deps:
         from phonopy import Phonopy
+
 
 @subflow
 @requires(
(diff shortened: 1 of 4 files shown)


src/quacc/recipes/common/phonons.py
```diff
@@ -71,59 +73,51 @@ def phonon_subflow(
...
         atoms = relax_job(atoms)["atoms"]
 
+    phonopy = get_phonopy(
+        atoms,
+        min_lengths=min_lengths,
+        supercell_matrix=supercell_matrix,
+        symprec=symprec,
+        displacement=displacement,
+        phonopy_kwargs=phonopy_kwargs,
+    )
```
Andrew-S-Rosen, line 118: @tomdemeyere: FYI --- It looks like `get_phonopy()` was originally being called _before_ the `relax_job`, which seems wrong. I have corrected that here.

tomdemeyere: Oops, I didn't even notice it, nice catch

PR #2108, opened 2024-05-09
Rosen phonopy fixed atoms (attempt 2)

Andrew-S-Rosen:


src/quacc/recipes/common/phonons.py
```diff
@@ -97,9 +101,16 @@ def _thermo_job(
         additional_fields: dict[str, Any] | None,
     ) -> PhononSchema:
         parameters = force_job_results[-1].get("parameters")
-        forces = [output["results"]["forces"] for output in force_job_results]
+        forces = [
+            output["results"]["forces"][~fixed_mask, :] for output in force_job_results
```
Andrew-S-Rosen, line 105: `fixed_mask` needs  to be added as a keyword argument to `_thermo_job`.

PR #2110, opened 2024-05-10
Demeyere phonopy fixed atoms v2

tomdemeyere:
## Summary of Changes

Should be much better

### Checklist

### Notes

### DIFF
--- a/src/quacc/atoms/phonons.py
+++ b/src/quacc/atoms/phonons.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import numpy as np
+from ase.constraints import FixAtoms
 from monty.dev import requires
 from pymatgen.io.ase import AseAtomsAdaptor
 from pymatgen.io.phonopy import get_phonopy_structure, get_pmg_structure
@@ -28,7 +29,6 @@ if TYPE_CHECKING:
 @requires(has_phonopy, "Phonopy not installed.")
 def get_phonopy(
     atoms: Atoms,
-    fixed_atoms: list[int] | None = None,
     min_lengths: float | tuple[float, float, float] | None = None,
     supercell_matrix: (
         tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]] | None
@@ -44,9 +44,6 @@ def get_phonopy(
     ----------
     atoms
         ASE atoms object.
-    fixed_atoms
-        List of indices of fixed atoms for the phono calculation.
-        WARNING: This is an important approximation and should be used with caution.
     min_lengths
         Minimum length of each lattice dimension (A).
     supercell_matrix
@@ -66,24 +63,31 @@ def get_phonopy(
     """
     phonopy_kwargs = phonopy_kwargs or {}
 
-    fixed_atoms = fixed_atoms or []
-    fixed_atoms = np.array([i in fixed_atoms for i in range(len(atoms))])
+    fixed_indices = np.array(
+        [
+            constr.get_indices()
+            for constr in atoms.constraints
+            if isinstance(constr, FixAtoms)
+        ]
+    )
+    
+    fixed_indices.flatten()
+    is_fixed_atoms = np.array([i in fixed_indices for i in range(len(atoms))])
 
     structure = AseAtomsAdaptor.get_structure(atoms)
     structure = SpacegroupAnalyzer(
         structure, symprec=symprec
     ).get_symmetrized_structure()
-    atoms = AseAtomsAdaptor.get_atoms(structure)
+    atoms = structure.to_ase_atoms()
 
-    fixed, non_fixed = atoms[fixed_atoms], atoms[~fixed_atoms]
+    fixed_atoms, non_fixed_atoms = atoms[is_fixed_atoms], atoms[~is_fixed_atoms]
 
-    non_fixed = AseAtomsAdaptor.get_structure(non_fixed)
+    non_fixed_atoms = AseAtomsAdaptor.get_structure(non_fixed_atoms)
 
     if supercell_matrix is None and min_lengths is not None:
-        n_supercells = np.round(np.ceil(min_lengths / atoms.cell.lengths()))
-        supercell_matrix = np.diag(np.diag([n_supercells, n_supercells, n_supercells]))
+        supercell_matrix = np.diag(np.round(np.ceil(min_lengths / atoms.cell.lengths())))
 
-    phonopy_atoms = get_phonopy_structure(non_fixed)
+    phonopy_atoms = get_phonopy_structure(non_fixed_atoms)
     phonon = phonopy.Phonopy(
         phonopy_atoms,
         symprec=symprec,
@@ -92,12 +96,15 @@ def get_phonopy(
     )
     phonon.generate_displacements(distance=displacement)
 
-    if len(fixed) > 0:
-        fixed = get_phonopy_structure(AseAtomsAdaptor.get_structure(fixed))
-        fixed = phonopy.structure.cells.get_supercell(fixed, supercell_matrix)
-        fixed = AseAtomsAdaptor.get_atoms(get_pmg_structure(fixed))
+    if fixed_atoms:
+        fixed_atoms = phonopy_atoms_to_ase_atoms(
+            phonopy.structure.cells.get_supercell(
+                get_phonopy_structure(AseAtomsAdaptor.get_structure(fixed_atoms)),
+                supercell_matrix,
+            )
+        )
 
-    return phonon, fixed
+    return phonon, fixed_atoms
 
 
 def phonopy_atoms_to_ase_atoms(phonpy_atoms: PhonopyAtoms) -> Atoms:
--- a/src/quacc/atoms/phonons.py
+++ b/src/quacc/atoms/phonons.py
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
 @requires(has_phonopy, "Phonopy not installed.")
 def get_phonopy(
     atoms: Atoms,
+    fixed_atoms: list[int] | None = None,
     min_lengths: float | tuple[float, float, float] | None = None,
     supercell_matrix: (
         tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]] | None
@@ -62,16 +63,21 @@ def get_phonopy(
     """
     phonopy_kwargs = phonopy_kwargs or {}
 
-    structure = AseAtomsAdaptor().get_structure(atoms)
+    structure = AseAtomsAdaptor.get_structure(atoms)
     structure = SpacegroupAnalyzer(
         structure, symprec=symprec
     ).get_symmetrized_structure()
+    atoms = AseAtomsAdaptor.get_atoms(structure)
+
+    fixed, non_fixed = atoms[fixed_atoms], atoms[~fixed_atoms]
+
+    non_fixed = AseAtomsAdaptor.get_structure(non_fixed)
 
     if supercell_matrix is None and min_lengths is not None:
         n_supercells = np.round(np.ceil(min_lengths / atoms.cell.lengths()))
-        supercell_matrix = np.diag([n_supercells, n_supercells, n_supercells])
+        supercell_matrix = np.diag(np.diag([n_supercells, n_supercells, n_supercells]))
 
-    phonopy_atoms = get_phonopy_structure(structure)
+    phonopy_atoms = get_phonopy_structure(non_fixed)
     phonon = phonopy.Phonopy(
         phonopy_atoms,
         symprec=symprec,
@@ -79,7 +85,13 @@ def get_phonopy(
         **phonopy_kwargs,
     )
     phonon.generate_displacements(distance=displacement)
-    return phonon
+
+    if len(fixed) > 0:
+        fixed = get_phonopy_structure(AseAtomsAdaptor.get_structure(fixed))
+        fixed = phonopy.structure.cells.get_supercell(fixed, supercell_matrix)
+        fixed = AseAtomsAdaptor.get_atoms(get_pmg_structure(fixed))
+
+    return phonon, fixed
 
 
 def phonopy_atoms_to_ase_atoms(phonpy_atoms: PhonopyAtoms) -> Atoms:
--- a/src/quacc/atoms/phonons.py
+++ b/src/quacc/atoms/phonons.py
@@ -5,11 +5,9 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import numpy as np
-from ase.constraints import FixAtoms
 from monty.dev import requires
 from pymatgen.io.ase import AseAtomsAdaptor
 from pymatgen.io.phonopy import get_phonopy_structure, get_pmg_structure
-from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
 
 try:
     import phonopy
@@ -29,14 +27,14 @@ if TYPE_CHECKING:
 @requires(has_phonopy, "Phonopy not installed.")
 def get_phonopy(
     atoms: Atoms,
-    min_lengths: float | tuple[float, float, float] | None = None,
+    min_lengths: float | tuple[float, float, float] | None = 20.0,
     supercell_matrix: (
         tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]] | None
     ) = None,
     symprec: float = 1e-5,
     displacement: float = 0.01,
     phonopy_kwargs: dict | None = None,
-) -> (Phonopy, Atoms):
+) -> Phonopy:
     """
     Convert an ASE atoms object to a phonopy object with displacements generated.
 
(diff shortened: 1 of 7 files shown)


src/quacc/atoms/phonons.py
```diff
@@ -60,22 +60,21 @@ def get_phonopy(
     """
     phonopy_kwargs = phonopy_kwargs or {}
 
-    symmetrized_structure = SpacegroupAnalyzer(
-        AseAtomsAdaptor().get_structure(atoms), symprec=symprec
-    ).get_symmetrized_structure()
-
```
Andrew-S-Rosen, line 66: Was this removed because `Phonopy()` is handling this with `symprec`? We're confident of that?

src/quacc/atoms/phonons.py
```diff
@@ -95,3 +94,33 @@ def phonopy_atoms_to_ase_atoms(phonpy_atoms: PhonopyAtoms) -> Atoms:
...
+    Get the supercell of an ASE atoms object using a supercell matrix.
+
+    Parameters
+    ----------
+    atoms
+        ASE atoms object.
+    supercell_matrix
+        The supercell matrix to use. If specified, it will override any
+        value specified by `min_lengths`.
+    Returns
```
Andrew-S-Rosen, line 115: Missing a line break before `Returns`.

src/quacc/atoms/phonons.py
```diff
@@ -95,3 +94,33 @@ def phonopy_atoms_to_ase_atoms(phonpy_atoms: PhonopyAtoms) -> Atoms:
...
+    """
+    Get the supercell of an ASE atoms object using a supercell matrix.
+
+    Parameters
+    ----------
+    atoms
+        ASE atoms object.
+    supercell_matrix
+        The supercell matrix to use. If specified, it will override any
+        value specified by `min_lengths`.
```
Andrew-S-Rosen, line 114: There is no `min_lengths` in this function. I think we can just ignore that part. Also, it must be specified since it is a positional argument.

src/quacc/recipes/common/phonons.py
```diff
@@ -48,12 +52,22 @@ def phonon_subflow(
     """
     Calculate phonon properties.
 
+    In Quacc the ASE constraints can be used to fix atoms. These atoms will
+    not be displaced during the phonon calculation. This will greatly reduce
+    the computational cost of the calculation. However, this is an important
+    approximation and should be used with caution.
```
Andrew-S-Rosen, line 58: I don't think this is relevant anymore.

src/quacc/recipes/common/phonons.py
```diff
@@ -17,13 +22,11 @@
...
     from typing import Any
 
-    from ase.atoms import Atoms
-
     from quacc import Job
     from quacc.schemas._aliases.phonons import PhononSchema
 
     if has_deps:
-        from phonopy import Phonopy
+        pass
```
Andrew-S-Rosen, line 29: Seems to be unnecessary.

src/quacc/recipes/common/phonons.py
```diff
@@ -48,12 +52,22 @@ def phonon_subflow(
...
 
+    In Quacc the ASE constraints can be used to fix atoms. These atoms will
+    not be displaced during the phonon calculation. This will greatly reduce
+    the computational cost of the calculation. However, this is an important
+    approximation and should be used with caution.
+
     Parameters
     ----------
     atoms
         Atoms object with calculator attached.
```
Andrew-S-Rosen, line 63: While we're here, this docstring seems wrong: should just be "Atoms object" (it doesn't have a calculator attached)

src/quacc/recipes/common/phonons.py
```diff
@@ -80,6 +94,27 @@ def phonon_subflow(
         Dictionary of results from [quacc.schemas.phonons.summarize_phonopy][]
     """
 
+    additional_atoms = additional_atoms or Atoms()
```
Andrew-S-Rosen, line 97: Nice. This took me a second to understand why you had done this. Very slick.

src/quacc/recipes/common/phonons.py
```diff
@@ -89,17 +124,25 @@ def _get_forces_subflow(supercells: list[Atoms]) -> list[dict]:
...
+        phonopy,
         force_job_results: list[dict],
-        t_step: float,
-        t_min: float,
-        t_max: float,
-        additional_fields: dict[str, Any] | None,
+        t_step,
+        t_min,
+        t_max,
+        additional_fields,
```
Andrew-S-Rosen, line 132: I assume you just removed the type hints for brevity's sake since they are basically implied elsewhere? I was also thinking about this since it is an internal function. Although, if we at some point enabled static type checking, having the type hints would ensure there isn't an error. I doubt we will be able to get to the point where we enable static type checking though... certainly not anytime soon.

src/quacc/recipes/emt/phonons.py
```diff
@@ -27,6 +27,7 @@ def phonon_flow(
         tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]] | None
     ) = None,
     displacement: float = 0.01,
+    additional_atoms: Atoms | None = None,
```
Andrew-S-Rosen, line 30: Note to self: we'll want to add this parameter to the other recipes calling the `phonon_subflow` (tblite and MLP I think?).

tests/core/atoms/test_phonons.py
Andrew-S-Rosen, line 1: I think this test still needs updating. `get_phonopy()` doesn't return a tuple anymore, and we aren't dealing with `FixAtoms`.

src/quacc/atoms/phonons.py
```diff
@@ -95,3 +94,33 @@ def phonopy_atoms_to_ase_atoms(phonpy_atoms: PhonopyAtoms) -> Atoms:
...
     pmg_structure = get_pmg_structure(phonpy_atoms)
     return pmg_structure.to_ase_atoms()
+
+
+def get_atoms_supercell_by_phonopy(
+    atoms: Atoms,
+    supercell_matrix: tuple[
+        tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]
+    ],
+) -> Atoms:
```
Andrew-S-Rosen, line 104: It would be ideal to have a simple unit test for this specific function.

Andrew-S-Rosen (commented):
@tomdemeyere: Thank you for taking the time to revisit this. It is certainly a lot cleaner and is very easy to follow now! I have some minor comments below, none of which are anything crazy.

I do just have one question: the `additional_atoms` is an interesting approach. I'd be interested to hear more about how you intend to use this in practice (or see people using it in practice).

In surface catalysis, I am most familiar with the opposite scenario to some degree. To calculate thermochemical corrections, people will invoke weak coupling between adsorbate and the metal surface. They'll then vibrate the adsorbate atoms and, perhaps, a few of the adjacent atoms on the surface while fixing the rest. They'll then use the harmonic approximation (e.g. with ASE utilities) to calculate the corrections. This is implicitly assuming that when you compare the energies of two different states along the reaction coordinate, the vibrational modes of the slab are effectively unchanged.

Here, it looks like it's almost the opposite perspective. I guess that's simply because here you are trying to calculate the phonon spectra of the material? If doing so, wouldn't fixing a few adsorbate atoms not really make a big difference compared to the size of the slab?

tests/core/recipes/emt_recipes/test_emt_phonons.py
```diff
@@ -70,4 +70,22 @@ def test_phonon_flow_v4(tmp_path, monkeypatch):
     assert output["results"]["thermal_properties"]["temperatures"][-1] == 1000
     assert output["results"]["force_constants"].shape == (8, 8, 3, 3)
     assert "mesh_properties" in output["results"]
-    assert output["atoms"] != atoms
+    assert output["atoms"] == atoms
```
Andrew-S-Rosen, line 73: Based on the test, it looks like the original `atoms` object is getting mutated somewhere? That doesn't seem ideal... 😅

Andrew-S-Rosen: Oops, that was a me problem. Fixing...

tomdemeyere:
Overall additional atoms can be seen as the same to "fixed_atoms". I renamed it to put an emphasis on the fact that these atoms are merely spectators of the Phonons calculation, only having an implicit influence through the forces. When you think about this has nothing to do with "fixing".


If you want to calculate an adsorbate on a slab that way you would just send the adsorbate part to regular atoms and the slab as "additional_atoms".


This could be renamed as "fixed_atoms" as well, I am trying to explore the semantic here

Andrew-S-Rosen:
> Overall additional atoms can be seen as the same to "fixed_atoms". I renamed it to put an emphasis on the fact that these atoms are merely spectators of the Phonons calculation, only having an implicit influence through the forces. When you think about this has nothing to do with "fixing".
> 
> If you want to calculate an adsorbate on a slab that way you would just send the adsorbate part to regular atoms and the slab as "additional_atoms".
> 
> This could be renamed as "fixed_atoms" as well, I am trying to explore the semantic here

To clarify, I understand that part. My question was about understanding why and when this is relevant, given that it is the opposite from what is usually invoked in the literature.

Andrew-S-Rosen:
Historically, in such scenarios, I have used ASE's `Vibrations` module (specifying the atoms to vibrate with `indices`) and then passed that information into ASE's `HarmonicThermo` class. However, using phonopy for this seems like it might be even better because it would be possible to run the displacement calculations in parallel rather than sequentially.

At some point, I was going to add `HarmonicThermo` to [`quacc.runners.thermochemistry`](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/runners/thermo.py) to support this, but the phonopy approach is great.

src/quacc/recipes/common/phonons.py
```diff
@@ -48,12 +52,22 @@ def phonon_subflow(
...
     Parameters
     ----------
     atoms
         Atoms object with calculator attached.
     force_job
         The static job to calculate the forces.
+    additional_atoms
+        Additional atoms to add to the supercells i.e. fixed atoms.
+        These atoms will not be displaced during the phonon calculation.
+        Useful for adsorbates on surfaces with weak coupling etc.
```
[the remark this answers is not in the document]
tomdemeyere: I completely see where this confusion can get from, this is why I am not entirely sure about the current name.

Although this forces the users to stop and think before using this important approximation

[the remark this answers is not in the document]
Andrew-S-Rosen: One downside of `fixed_atoms` is that it potentially implies that they are sites within `atoms` (the first positional argument) that are fixed, whereas in reality this is a separate `Atoms` object. That's one reason why I like `additional_atoms` since it makes it clear that this is a separate `Atoms` object, although one would understand that if they read the type hint. Of course, `additional_atoms` doesn't inherently specify anything about them being fixed until you read the docstring.

I think either could be okay, personally.

Andrew-S-Rosen:
What if we rename the first positional argument to `unfixed_atoms` and then `additional_atoms` as `fixed_atoms`? Not sure if that helps or hurts...

tomdemeyere:
Claude 3 Opus is suggesting (😅):

1. primary_atoms (instead of atoms) and auxiliary_atoms
2. atoms and fixed_atoms
3. displaced_atoms (instead of atoms) and nondisplaced_atoms

In this order of "preference", I might actually lean towards 2, (let's just be pragmatic?)

Andrew-S-Rosen:
Ooh I like 3! It's better than my suggestion because it doesn't accidentally imply a `FixAtoms` constraint. But I'm okay with 2 if you have a strong opinion about it.

tomdemeyere:
That's a good argument let's go with that, I just added an extra underscore

src/quacc/recipes/emt/phonons.py
```diff
@@ -92,15 +98,16 @@ def phonon_flow(
         parameters=job_params,
         decorators=job_decorators,
     )
-    if run_relax:
+    if run_relax and not non_displaced_atoms:
```
tomdemeyere, line 101: This is a little bit problematic, another option is to send both displaced and non_displaced and optimise them and separate them again...

Andrew-S-Rosen: I see. I mean, I guess we can ask if we really even need a relaxation step beforehand? Presumably one could just call `relax_job()` before calling the flow and it would be exactly the same (although without a tight force tolerance).

If there's a desire to keep this though, then yeah I think the only route would be to relax `displaced_atoms+non_displaced_atoms` as a single `combined_atoms` and then pass in `combined_atoms[:len(displaced_atoms)]` etc. to `phonon_subflow`.

src/quacc/atoms/phonons.py
tomdemeyere: What is this expected to do? I was afraid that it would translate/modify atoms object: this might surprise users that have to place atoms to specific place in unit cells (coulomb corrections, ...) but after trying it seems that it does not do anything? Is it for metadata?

Andrew-S-Rosen: The idea is that if you're doing a relaxation on a structure with many degrees of freedom, it almost certainly will be P1 symmetry just due to slight fluctuations. This makes the number of displacements _enormous_ even though it is near a higher symmetry. This line symmetrizes the structure, based on the value of `symprec`, to ensure the phonon displacements are calculated for the symmetrized structure. The default `symprec` is 1e-5, which is a relatively small value. If the structure being passed through the workflow is already at the high symmetry space group or is not within that `symprec` cutoff, there will be no change. If the user doesn't want any symmetrization, they can always toggle `symprec`. 

Now, `symprec` is also passed to `Phonopy`, so my original question was about if Phonopy as already doing this or not. The documentation for the Python API of Phonopy does not make it clear.

Andrew-S-Rosen: That said, we can simply leave a note in the docstring and rely on the user to symmetrize their structure before passing it to the flow if we prefer.

tomdemeyere: I would expect Phonopy to make sure of that through the symprec keyword yes:

``` python
slab = bulk("Pt", cubic=True) * (4, 4, 4)
slab.rattle(0.001)

phonon = get_phonopy(slab, supercell_matrix=[1, 1, 1])

print(Symmetry(phonon.get_supercell(), symprec=0.001*3*2).get_international_table())
```
"Fm-3m (225)"

Andrew-S-Rosen: We can remove it then

tests/core/recipes/emt_recipes/test_emt_phonons.py
Andrew-S-Rosen: I think this was related to https://github.com/Quantum-Accelerators/quacc/issues/2230.

tomdemeyere:
@Andrew-S-Rosen 

I did not forget this PR, as you may know I am very busy currently, I will definitely come back to it ASAP!

Andrew-S-Rosen:
There is no rush at all :) Good luck with wrapping up the PhD!

tomdemeyere:
I have some more time to do what I like now :)

I added some of your suggestions, namely:

- The type hinting in the internal `_thermo_job`
- Fixed the unnecessary comments
- Added a test for the `get_phonopy_supercell`

You will see that I added the `non_displaced_atoms` object to `additional_fields` in the `phonopy_subflow`. I did it because otherwise the result dict does not contain any information about these atoms at all. There might be a better way to do this? I will let you decide.

src/quacc/recipes/common/phonons.py
```diff
@@ -89,17 +117,25 @@ def _get_forces_subflow(supercells: list[Atoms]) -> list[dict]:
     @job
     def _thermo_job(
         atoms: Atoms,
-        phonopy: Phonopy,
+        phonopy,
```
Andrew-S-Rosen, line 120: Looks like we lost a type hint.

src/quacc/recipes/common/phonons.py
```diff
@@ -5,34 +5,36 @@
...
-
 
 @subflow
 @requires(has_phonopy, "Phonopy must be installed. Run `pip install quacc[phonons]`")
 @requires(has_seekpath, "Seekpath must be installed. Run `pip install quacc[phonons]`")
 def phonon_subflow(
-    atoms: Atoms,
+    displaced_atoms: Atoms,
     force_job: Job,
+    non_displaced_atoms: Atoms | None = None,
```
Andrew-S-Rosen, line 37: Out of curiosity, what is the motivation to have two separate `Atoms` objects rather than one `Atoms` objects for the full thing and specific indices to not displace? It seems like it might be a bit awkward if they are separate `Atoms` objects because wouldn't the user have to split up their `Atoms` object into two before calling this function? I'm sure there's a reason for your suggestion --- just trying to pick up on what it is. We may have gone over this a long time ago.

src/quacc/recipes/common/phonons.py
```diff
@@ -5,34 +5,36 @@
...
-    if has_phonopy:
-        from phonopy import Phonopy
-
 
 @subflow
 @requires(has_phonopy, "Phonopy must be installed. Run `pip install quacc[phonons]`")
 @requires(has_seekpath, "Seekpath must be installed. Run `pip install quacc[phonons]`")
 def phonon_subflow(
-    atoms: Atoms,
+    displaced_atoms: Atoms,
```
Andrew-S-Rosen, line 35: As an aside, `displaced_atoms` is a bit awkward of a argument name here because it sounds as if the `Atoms` have already been displaced (past tense) when in reality they are _going_ to be displaced. Not sure I have better alternative per se but just commenting.

Andrew-S-Rosen (commented):
@tomdemeyere: What a surprise! Thanks for taking care of this. I think it will be a very nice feature.

I left a few very minor comments but am happy to merge thereafter. I think your changes make a lot of sense.

src/quacc/recipes/common/phonons.py
```diff
@@ -31,6 +32,7 @@ def phonon_subflow(
     atoms: Atoms,
     force_job: Job,
     relax_job: Job | None = None,
+    fixed_atoms: list[int] | None = None,
```
tomdemeyere: Do we want to come back to this approach?

src/quacc/recipes/common/phonons.py
tomdemeyere: In commit "https://github.com/Quantum-Accelerators/quacc/pull/2110/commits/24608c9046a1b0a8cb2bbfaf0691e5a2fc60e127" I left a comment.

I think we moved out of this approach to the "FixAtoms" approach, but that did not work as well so ended up here.

Andrew-S-Rosen: @tomdemeyere: Apologies if we're revisiting history here (it's been a while...). I'm not sure the commit shows the commit, but my question is mostly about what if we had an `Atoms` object as input with a `list[int]` of indices to not displace. Was there an issue with that approach? If so, we can stick with what you have here

src/quacc/recipes/common/phonons.py
Andrew-S-Rosen: @tomdemeyere: Ah, we did that before. Sorry for the memory issues..........

Let me read up on what happened before.

Andrew-S-Rosen:
@tomdemeyere: Sorry about the back-and-forth. Could you remind me the motivation to switch from specifying a `list[int]` to two `Atoms` objects?

Otherwise, once the type hint is fixed, I'll merge based on the above discussion.

tomdemeyere:
@Andrew-S-Rosen I don't really remember, this might be linked to other problems at the time.

It seems that changing to the index approach is rather straightforward, and might make more sense?

Andrew-S-Rosen:
Let's give it a go and see what happens! It seems a tad easier on the user in my opinion. Since it should also support negative indexing, that should make things pretty trivial for the user since it's usually the last N indices where an adsorbate is.

tomdemeyere:
@Andrew-S-Rosen Tell me what you think.

I also made sure that the Schema returns the same `Atoms` object as the one being sent. I also included `non_displaced_atoms` and `displaced_atoms` when relevant.

PR #2163, opened 2024-05-17
Add support for dynamically changing settings with an activated workflow engine

### LINKED ISSUE #2147: Improve dynamic changing of `SETTINGS` parameters for decorated functions with active workflow engines

Recently, we introduced a function [`change_settings`](https://quantum-accelerators.github.io/quacc/reference/quacc/settings.html#quacc.settings.change_settings) that can be used to _temporarily_ modify the settings of some function calls in a script. See [here](https://quantum-accelerators.github.io/quacc/user/settings/settings.html#modifying-the-global-settings-in-a-script) for details. An example is like that shown below:

```python
from quacc import change_settings

with change_settings({"RESULTS_DIR": "/new/path/to/store/results"}):
    pass  # Your calculation here
```

This works wonderfully. However, as noted on that same page, if you dispatch the calculation to a remote machine (as is common with the use of a workflow engine), there is no knowledge of changes to in-memory parameters like the global `SETTINGS` object. We should come up with a smart and user-friendly way to address this.

For instance, let's say you want to change `SETTINGS.DEBUG` to `True` for a `@job` called `relax_job`. Most likely, it can be solved by:
1. Stripping the `@job`, `@flow`, or `@subflow` decorator (`from quacc import strip_decorator`) from the imported function. In this case, it'd be a `@job`.
2. Dynamically redefining the function with an update to `SETTINGS.DEBUG` inside of the function definition. Important: we must retain the original function's name, signature, type hints, etc. so presumably `functools.wraps` here. 
3. Redecorating it with the original decorator. 
4. Returning the redecorated function to the user

The tools to do this are likely already in `quacc.wflow_tools.customizers`, but the exact logic to get it right in an automated fashion currently escapes me.

For background context, refer to https://github.com/Quantum-Accelerators/quacc/discussions/2145 and [my answer therein](https://github.com/Quantum-Accelerators/quacc/discussions/2145#discussioncomment-9435337).

honghuikim:
## Summary of Changes

Closes #2147. 

### Checklist

### Notes


Andrew-S-Rosen:
Wonderful, thanks for kicking this off! I'll review in the next few days and try to address the remaining issue you raised.

src/quacc/settings.py
```diff
@@ -552,3 +552,33 @@ def change_settings(changes: dict[str, Any]) -> QuaccSettings:  # type: ignore
...
+        The decorator associated with `func`.
+
+    Returns
+    -------
+    Callable
+        The updated function.
+    """
+    from quacc import change_settings, strip_decorator
+
+    def concatenated(*args, **kwargs):
```
Andrew-S-Rosen, line 580: Can you see if decorating `concatenated` with `@wraps` still causes the tests to pass? It is important to retain the original function's metadata: https://docs.python.org/3/library/functools.html#functools.wraps

honghuikim: Yes, it passes all the tests under `parsl` in my local. I added a new commit.

Andrew-S-Rosen:
Thank you, @honghuikim! Overall, this looks great. That said, I can indeed confirm the issue you raised in #2147, which I have reproduced as follows:

```bash
!quacc set WORKFLOW_ENGINE parsl
```

```python
import parsl
parsl.load()

from quacc.recipes.emt.slabs import static_job
from quacc import change_settings_wf, job
from ase.build import bulk

atoms = bulk("Cu")
static_job1 = change_settings_wf(static_job, {"RESULTS_DIR": "~/test1"}, job)
static_job2 = change_settings_wf(static_job1, {"RESULTS_DIR": "~/test2"}, job)

static_job1(atoms).result()

**Comment by Andrew-S-Rosen** [comment]
I'll give this some more thought too. We're close! Wish I had a better answer in the meantime.

**Comment by honghuikim** [comment]
@Andrew-S-Rosen  getting back to this issue took longer than I expected.
I've been digging into this issue and found some other weird behaviors. Unfortunately, I haven't found a general solution for flow and subflow, but I think it's worth discussing the examples below to get there.

## 1.  `change_settings_wf` works perfectly for `job`, but not for `flow` and `subflow`
It doesn't work as we want because of the same reason with the nested structure as below.
```python
with change_settings({"RESULTS_DIR": "~/test2"}):
    with change_settings({"RESULTS_DIR": "~/test1"}):
        static_job(atoms)
```
The situation above is (I think) equivalent to the below example.
```python
@flow
def test_flow(atoms):
    test_job1(atoms)
    test_job2(atoms)

new_test_flow = change_settings_wf(test_flow, {"RESULTS_DIR": "~/test"}, flow)
# it cannot influence the inner jobs (test_job1 and test_job2)
```
This is the reason why we get issues in the `bulk_to_slabs_flow` example you provided.
The below example resolves the issue by modifying inner jobs directly (it's not the solution that we want though).
```python
changes = {"RESULTS_DIR": "~/test3"}
relax_job_ = change_settings_wf(relax_job, changes, job)
static_job_ = change_settings_wf(static_job, changes, job)

@flow
def bulk_to_slabs_flow_(atoms):
    return bulk_to_slabs_subflow(
        atoms,
        relax_job_,
        static_job=static_job_,
    )

atoms = bulk("Cu")
bulk_to_slabs_flow_new = change_settings_wf(bulk_to_slabs_flow_, changes, flow)
bulk_to_slabs_flow_new(atoms).result()
```
But from here, I've found some weird results (see below 2, 3, 4).

## 2. unexpected output 1 
The above `bulk_to_slabs_flow` example has a nested structure;
```python
flow:
    subflow:
        jobs
```
So if we run the below example, outputs expected to be not stored in `~/test3`
```python
import parsl
parsl.load()

changes = {"RESULTS_DIR": "~/test3"}
bulk_to_slabs_subflow_ = change_settings_wf(bulk_to_slabs_subflow, changes, subflow)

@flow
def bulk_to_slabs_flow_(atoms):
    return bulk_to_slabs_subflow_(
        atoms,
        relax_job,
        static_job=static_job,
    )

atoms = bulk("Cu")
bulk_to_slabs_flow_(atoms).result()
```
But this stores  two `relax job`s in `~/test3` and two `relax job`s , four `static job`s in `Path.cwd()`.
(this might be the only case for my working environment)

## 3. unexpected output 2
The below `write_file_flow` also has a nested structure (flow(job)), but `change_settings_wf` makes them save files in the desired directory.
```python
@job
def write_file(name="job"):
    with open(Path(f"{SETTINGS.RESULTS_DIR}/{name}.txt"), "w") as f:
        f.write("test file")

@flow
def write_file_flow(name="flow"):
    return write_file(name=name)

write_file_new = change_settings_wf(write_file, {"RESULTS_DIR": "~/job"}, job)
write_file_flow_new = change_settings_wf(write_file_flow, {"RESULTS_DIR": "~/flow"}, flow)

write_file_new().result()
write_file_flow_new().result()
```

## 4. unexpected output 3
Running the  pytest of `parsl/test_syntax.py` multiple times on local gives inconsistent outputs, either pass or fail.
I suspect there are unwanted SETTINGS calling between functions, and this will not be the case in using HPC.


## suggestions, not optimal
* If we can strip all decorators in a given Callable object (including itself) and redecorate them exactly at the same place, we can solve the issues.
* We might be able to handle this in the decorating step?

Andrew-S-Rosen:
@honghuikim: Thank you so much for your detailed investigation here. This is a _really_ tough (but important) problem, and I appreciate the time and energy you have already put into this.

Unfortunately, I also remain unsure about the best path forward here.

> If we can strip all decorators in a given Callable object (including itself) and redecorate them exactly at the same place, we can solve the issues.

It's worth noting that there is already a mechanism in quacc that does basically this. Take a look at the green textbox at the bottom of [this page](https://quantum-accelerators.github.io/quacc/user/wflow_engine/wflow_engines2.html#running-a-user-constructed-dynamic-workflow) for an example. Basically, all `@flow`s in quacc allow the user to modify the underlying `@job` (and `@subflow`) decorators via a custom `job_decorators` keyword argument. [Here](https://github.com/Quantum-Accelerators/quacc/blob/ebd32d66cbf82b89925b7fad55a6b584ce6b027a/src/quacc/recipes/emt/slabs.py#L63-L68) is the logic where that's being done in the EMT `bulk_to_slabs_flow` recipe.

Since this is already in place, I wonder if we should support adding a custom keyword argument to the decorators to modify `SETTINGS` on-the-fly. For instance, maybe something like

```python
@job(settings_swap={"RESULTS_DIR": "~/test"})
```

would then modify the `SETTINGS.RESULTS_DIR` for just that `@job`. From a user-facing perspective, that might be

```python
bulk_to_slabs_flow(..., job_decorators={"static_job": {"settings_swap": {"RESULTS_DIR": "~/test"}}})
```

to modify the `RESULTS_DIR` setting for just the static job in the flow.

> We might be able to handle this in the decorating step?

I suppose that this is somewhat similar to what I outlined above.

Thoughts?

honghuikim:
@Andrew-S-Rosen  Thank you for the reply and suggestions. I totally agree with your outline. I think I slightly misread your reply yesterday, which resulted in a slightly different implementation below.
Basically I just made another argument named `changes` in `customize_funcs` and applied the changes using `change_settings_wf`.
```python
def customize_funcs(
    #other original kwargs,
    changes: dict[str, dict[str, Any]] | None=None,
) -> tuple[Callable, ...] | Callable:
    # original part
        if change := changes.get("all"):
            func_ = change_settings_wf(func_, change, decorator)
        if change := changes.get(names[i]):
            func_ = change_settings_wf(func_, change, decorator)

@flow
def bulk_to_slabs_flow(
    #other original kwargs,
    job_changes: dict[str, dict[str, Any]] | None = None,
) -> list[RunSchema | OptSchema]:
```
I checked that this works without errors for `bulk_to_slabs_flow`, but this will require adding an additional input argument (e.g. `job_changes`) to all pre-defined `flow`s.
So, I think what you suggested looks more robust and less annoying
```python
@job(settings_swap={"RESULTS_DIR": "~/test"})
```
It's simply the problem of choosing the position to use the `change_settings` function. Which one do you think is better?

Andrew-S-Rosen:
Wonderful to hear that your similar approach worked!

I see merits of both approaches. That said, the main benefit I see with my proposed example is that it is also possible for users to update the settings of an imported job (rather than just a flow). For instance, this could be done with the `redecorate` function described [in the tutorial](https://quantum-accelerators.github.io/quacc/user/wflow_engine/wflow_engines2.html#running-a-user-constructed-serial-workflow). If we take your approach, we have only added functionality for modifying the settings of jobs that are already within a pre-made flow.

I could be mistaken. This is just me brainstorming. You're the one who has done the difficult thinking on this (thank you!).

Andrew-S-Rosen:
Sounds good! Ultimately, I will defer to you about how to best proceed here. 👍

honghuikim:
I just have successfully implemented the approach with minimal changes to the original framework. 
The implementation itself seems nice, but it seems to have the same issue as the "unexpected output 3". I guess there exist interventions between tasks' QuaccSettings? but not sure what exactly is happening. 

It's getting late here, so I'll add commits with unit tests tomorrow morning. Hopefully, everything works smoothly!

Andrew-S-Rosen:
Thank you very much, @honghuikim! Hopefully I can try and sort out what is going wrong with the flakey Parsl tests. Parsl runs the `@job`s concurrently, so what I imagine is happening is that the setting change on one `@job` is impact the settings of the concurrent `@job`. This, naturally, should not be the case. I'll dig into it once there's a minimal example for me to debug.

honghuikim:
@Andrew-S-Rosen I just added a unittest for Parsl. With this PR, we can modify settings of `job` and `flow`. 
The current implementation doesn't solve the problem in the nested structure (repeated redecoration), but this can be easily resolved. I'll add this commit soon.

For reproducing the inconsistent Parsl output,
```python
import parsl
parsl.load()

from quacc import  job
from ase.build import bulk

atoms = bulk("Cu")

bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"})}).result()
```
There will be eight output directories in total, and four of them should be stored in `~/test`, but you will see that it's not always the case once you run it multiple times.
If we set ` job_decorators={"all": job(settings_swap={"RESULTS_DIR": "~/test"})}`, all the eight directories will be saved in `~/test`.

src/quacc/settings.py
```diff
@@ -569,3 +570,36 @@ def change_settings(changes: dict[str, Any]):
...
+        The settings to apply within the context manager.
+
+    Returns
+    -------
+    Callable
+        The wrapped function.
+    """
+    from quacc import change_settings
+
+    if hasattr(func, "__changed__") and func.__changed__:
```
Andrew-S-Rosen, line 593: You can likely do

```python
if getattr(func, "__changed__", False):
```

src/quacc/settings.py
```diff
@@ -569,3 +570,36 @@ def change_settings(changes: dict[str, Any]):
...
+    else:
+        original_func = func
+
+    @wraps(original_func)
+    def wrapper(*args, **kwargs):
+        with change_settings(changes):
+            return original_func(*args, **kwargs)
+
+    wrapper.__changed__ = True
+    wrapper.original_func = original_func
```
Andrew-S-Rosen, line 604: When you do `@wraps(original_func)`, it automatically stores the original function under `wrapper.__wrapped__`, so there shouldn't be a need to manually add on a new attribute to store the original function.

src/quacc/settings.py
```diff
@@ -569,3 +570,36 @@ def change_settings(changes: dict[str, Any]):
...
+    -------
+    Callable
+        The wrapped function.
+    """
+    from quacc import change_settings
+
+    if hasattr(func, "__changed__") and func.__changed__:
+        original_func = func.original_func
+    else:
+        original_func = func
```
Andrew-S-Rosen, line 596: With the knowledge that `.__wrapped__` is always there with `@wraps`, this logical can be simplified to something like:

```python
original_func = func.__wrapped__ if getattr(func, "__changed__", False) else func
```

src/quacc/wflow_tools/decorators.py
```diff
@@ -616,6 +619,9 @@ def wrapper(
     ):
         return func(*f_args, **f_kwargs)
 
+    if hasattr(func, "__changed__") and func.__changed__:
```
Andrew-S-Rosen, line 622: We can probably use `getattr` with a default here again, if I'm understanding `getattr` corectly.

src/quacc/wflow_tools/decorators.py
```diff
@@ -616,6 +619,9 @@ def wrapper(
...
+        wrapper.__changed__ = func.__changed__
+        wrapper.original_func = func.original_func
```
Andrew-S-Rosen, line 624: Worth double-checking if this is still needed?

Andrew-S-Rosen (commented):
@honghuikim: Thank you very much for your continued work on this!

I cleaned up the test slightly but confirmed that it makes sense. Not to worry about other workflow engines --- I can take care of the intricacies there when we confirm one of the others (e.g. Parsl) works as expected.

I left a few comments below that may clean things up a little bit more, but it's nothing major. Also, feel free to add your name [here](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/about/contributors.md#contributors) if you'd like 😄 Hopefully we can get the remaining loose ends sorted out!

I was, however, unable to reproduce the error you are seeing with the `bulk_to_slabs_flow()` (which I agree was very important to test). I tried running `bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"})})` four sequential times in an IPython console, and I was left with the appropriate number of folders in `~/test` (a total of 16). Could you clarify a bit about how to reproduce the erratic behavior? It definitely sounds like something important to be solved.

src/quacc/settings.py
honghuikim: Thank you! I changed it to getattr

src/quacc/settings.py
honghuikim: Sounds good, but using the `__wrapped__` attribute might require slight modifications.

I tried to run pytest on `test_syntax.py` with all the modifications you suggested, but it failed on `test_change_settings_redecorate`. I think it's because `wrappers.__wrapped__` is not always the "real" original function especially when multiple wrappers wrap a function (e.g. nested calling of `change_settings_wrap` or combination with wrappers from workflow engines (e.g. `python_app`)).

so we can either assign a new attribute as I did, or find the real original function like below.
```python
def find_original_func(func):
    while hasattr(func, "__wrapped__"):
        func = func.__wrapped__
    return func
```
The above approach shouldn't cause any issues, but I'm currently encountering a failure with pytest. I'll let you know once it's resolved.

Andrew-S-Rosen: Ah, I understand now. Thanks for clarifying. Doing the `while` loop approach probably isn't wise because we don't want to accidentally go "too deep" (the "original function" might have had a `@wraps` from some other cause). The custom attribute is okay for now.

However, I do have one question: can you clarify why this logic is necessary in general? I understand that it is needed so that if you call `change_settings_wrap()` on a `@job` twice that it will override the inner change. Is this necessary to address though? I cannot really think of a scenario where someone would willingly call `change_settings_wrap()` twice on a `@job`.

We had this example in a prior discussion:

```python
import parsl
parsl.load()

from quacc.recipes.emt.core import static_job
from quacc import change_settings_wf, job
from ase.build import bulk

atoms = bulk("Cu")
static_job1 = change_settings_wf(static_job, {"RESULTS_DIR": "~/test1"}, job)
static_job2 = change_settings_wf(static_job1, {"RESULTS_DIR": "~/test2"}, job)

static_job1(atoms).result()
static_job2(atoms).result()
```

But when would one ever do this in practice? If you wanted two separate `static_job` calls with different settings, you should be able to do:

```python
import parsl
parsl.load()

from quacc.recipes.emt.core import static_job
from quacc import change_settings_wf, job
from ase.build import bulk

atoms = bulk("Cu")
static_job1 = change_settings_wf(static_job, {"RESULTS_DIR": "~/test1"}, job)
static_job2 = change_settings_wf(static_job, {"RESULTS_DIR": "~/test2"}, job) # note: I change this line!

static_job1(atoms).result()
static_job2(atoms).result()
```

If this does not work as-is, then it means the underlying `static_job` is getting mutated, which we don't want to happen (e.g. we can copy it first). But if it works as is, which seems like the case based on a quick glance of the source code, then what is the remaining scenario where we need to handle nested calls to `change_settings_wrap`? I just want to make sure we aren't adding unnecessary functionality. It's possible I'm missing the intention here.

honghuikim: @Andrew-S-Rosen Thank you for bringing up this discussion! I also had the exact same concern a few days ago, that is in what situation do we really face the nested calling of `change_settings_wrap`. My short answer is to cover the situation where users not intentionally rechange the function already changed, so that the `change_settings_wrap` can work as they thought. I agree that this should not be done with intention, but this can be troublesome if we globally change a specific job at first and accidentally try to change the job again in a code like below.
```python
static_job = change_settings_wf(static_job, {"RESULTS_DIR": "~/test1"}, job) #literally rewrite the static_job
# some other codes run
static_job_specific = change_settings_wf(static_job, {"RESULTS_DIR": "~/test2"}, job)
```

I was also thinking of adding another feature to `change_settings_wrap` that can freeze the change so that next `change_settings_wrap` call cannot change it again. But I didn't want to make this logic too complicated, and result in overriding the previous change.

At the moment, I think it's better to throw an error (or warning) if `change_settings_wrap` is applied on a function that already has changed (e.g. `if getattr(func, "__changed__", False):`), and not adding the new attribute (`original_func`). So basically we have two options; 1) tolerate (mis)double calling of `change_settings_wrap` (this was my intent) or 2) warn the mistake to users.

Andrew-S-Rosen: @honghuikim: Ah, that's a good example. I can see how someone would accidentally make that kind of error now if they just pick the same name as the original function for convenience.

At least for right now, I think I am okay with either option (the "smart" approach or raising a warning/error). I will leave it up to you to do what you think is best. The `original_func` behavior is entirely independent of the proper functioning of `change_settings_wf` in most cases, which is the important part anyway.

This is mostly a note to myself, but if we want to make sure that the "double application" of `change_settings_wrap` works appropriately for all workflow engines, we will need to make sure that we duplicate the behavior below (which is specifically for Parsl) elsewhere as well:
https://github.com/honghuikim/quacc/blob/40cf20feba9c79ce0f388232264aeb85e9a38f4c/src/quacc/wflow_tools/decorators.py#L622-L624

Andrew-S-Rosen (commented):
@honghuikim: Alright! I have added tests for all the other workflow engines, and everything is working as expected. 👍 (Note: I moved your tests to `test_customizers.py` instead of being in `test_syntax.py`)

This means, at least for me, there is really only one remaining question left before this can be merged: you said there was a sporadic error with Parsl when applying the changes several times to the `bulk_to_slabs_flow`. Can you still reproduce this error? I am having trouble doing so but want to make sure we address any lingering issues like this if they are still around. We don't have any tests for changing the settings with a "real" recipe like `bulk_to_slabs_flow` (although maybe we should...), so I haven't been focusing on it that much. But I did try it out locally and couldn't see the problem you were seeing.

src/quacc/settings.py
```diff
@@ -569,3 +570,33 @@ def change_settings(changes: dict[str, Any]):
...
+
+    original_func = func.original_func if getattr(func, "__changed__", False) else func
+
+    @wraps(original_func)
+    def wrapper(*args, **kwargs):
+        with change_settings(changes):
+            return original_func(*args, **kwargs)
+
+    wrapper.__changed__ = True
+    wrapper.original_func = original_func
```
Andrew-S-Rosen, line 601: Could we call this `._original_func` since it's not meant to be part of the public API? Thanks!

src/quacc/settings.py
honghuikim: @Andrew-S-Rosen  Sounds good! I'm leaning toward keeping the `original_func`, and of course we can get back to this discussion once we get unexpected behavior from it. In that situation, raising a warning/error will be an easy solution.

Can we test the "double application" of `change_settings_wrap` on other workflow engines just with the current implementation? I tried it on my local but I got some errors since I have no experience using other workflow engines except Parsl. If we pass the wrapped `_func` (input to `job` decorator) to workflow engines, I think `__changed__` and `_original_func` attributes should be preserved (and this is what it is in the current implementation?).

src/quacc/settings.py
honghuikim: Of course, thanks for suggestion. I'll update it.

honghuikim:
@Andrew-S-Rosen  Thank you very much for adding the other tets!

Yes, I can still get inconsistent output from it. I tested it on Linux and MacOS, and they have different outputs (first one has only one directory saved in the changed location while latter one has three directories in there). Let me try some others too so that you can generate this behavior. If this ends to the only error that I have, then we can conclude that my working environment has some problems.

src/quacc/settings.py
Andrew-S-Rosen: > Can we test the "double application" of change_settings_wrap on other workflow engines just with the current implementation?

We can't with the current setup because

https://github.com/honghuikim/quacc/blob/40cf20feba9c79ce0f388232264aeb85e9a38f4c/src/quacc/wflow_tools/decorators.py#L622-L624

is only being applied for Parsl. I will take a look later and see if I can get the same logic and tests going with the other workflow engines. No need for you to worry about it since I'm probably the only person who knows 5 different workflow engines...

Andrew-S-Rosen:
Thanks! Any additional info would be a big help. Quite odd that you're getting different behavior on different operating systems! Ironically, I was trying Linux...

honghuikim:
I don't think this is due to different OS. I also have that odd behavior in the same OS..

honghuikim:
I found out that assigning changes to all jobs doesn't show such odd behavior.  
```python 
bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"}),
                                        "static_job": job(settings_swap={"RESULTS_DIR": "~/"})}).result()
```
I still get some random number of directories saved in either `~/test` or `~/`(default RESULTS_DIR) with below code.
```python 
bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"})}).result()
```

src/quacc/settings.py
honghuikim: I see. I also just tested redun and prefect, and they didn't pass the double-calling test.

Now I'm leaning toward throwing an error/warning for the double-calling... I thought this should cover most of other workfklow engines, but if not I think it's better to throw an error/warning. We can still keep the Parsl's double-calling implementation or make it consistent with others. 

What do you think?

Andrew-S-Rosen:
Thank you for the detailed report, @honghuikim! I am finally able to reproduce the behavior (perhaps because I ran it as a Python script instead of in IPython now, or maybe just a coincidence). I will need to look into this a bit more to see what might be going on, almost certainly before the end of the week. I'll let you know what I find.

------

For future me, here is the full example I used:

```python
import parsl
from ase.build import bulk
from quacc import job

parsl.load()

atoms = bulk("Cu")

bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"})}).result()
bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"})}).result()
bulk_to_slabs_flow(atoms, job_decorators={"relax_job": job(settings_swap={"RESULTS_DIR": "~/test"})}).result()
```

Andrew-S-Rosen:
Good news and bad news!

First, the good news. I figure out what's going on (I think). If you replace the following lines

https://github.com/Quantum-Accelerators/quacc/blob/d716ba2ceae4e4bdb73f92f22d54a6ee8376ea77/src/quacc/settings.py#L570-L571

with `pass` (so nothing happens instead), the folders seem to write out to the correct directory.

The bad news is that this is a big mess if true. My suspicion is that the global `SETTINGS` variable is shared across multiple threads. So, when we change `SETTINGS` in one thread, we change it in all concurrently running processes. The way the `change_settings` context handler works is that it only changes the `SETTINGS` temporarily for that `@job`. When the `@job` is done running, it puts the `SETTINGS` back to what it was originally.

I think that's the cause of our issue here. We are changing `SETTINGS.RESULTS_DIR` to `~/test` in one process (`relax_job`), but in another process (`static_job`) it is the default `~`. Depending on when the concurrent process actually accesses that `SETTINGS.RESULTS_DIR` variable, it could either be `~/test` or `~` depending on the status of concurrently running processes.

Again, I'm not 100% certain, but it's possible that global variables are shared over concurrent threads, which is a bit of a mess. Even if we didn't use a context handler here, any changes to `SETTINGS` in one `@job` would be influencing another `@job` even if those `@job`s are completely unrelated to one another.

😅

honghuikim:
@Andrew-S-Rosen Thank you for digging into this error! I have the same suspicions, and I think you are right. 
I tried to make a copied SETTINGS for each job when using `change_settings`, but it didn't work out as in your above example... Can we re-initialize the QuaccSettings in a job? If possible, that could be the solution. 
I haven't tested it on HPC, but do you think this erroneous behavior exists in HPC that uses different compute nodes for jobs?

Andrew-S-Rosen:
> Can we re-initialize the QuaccSettings in a job? If possible, that could be the solution.

While I'm not 100% certain how to do this offhand, it is potentially plausible. I will try to think about how to do this later in the week.

> I haven't tested it on HPC, but do you think this erroneous behavior exists in HPC that uses different compute nodes for jobs?

This will depend on how the jobs are launched. If every job is launched on a unique node (or unique Slurm job), it should work as expected. But if you "pack" a single node with multiple concurrently running processes, it might get... interesting.

honghuikim:
I see, thank you very much for investigating this more. I will also dig into it more.
I have very little knowledge about other workflow engines, but do you think this is the only error with Parsl?

Andrew-S-Rosen:
Thanks for your help too! I recognize that this is actually a very complex problem for a first contribution and am grateful for all you have already done thus far to help. 

> I have very little knowledge about other workflow engines, but do you think this is the only error with Parsl?

I think the error is likely to occur anytime Python's multiprocessing utilities are used, which would be the case for all the other workflow engines as well (except Jobflow).

honghuikim:
Thanks for the detailed explanation!
I was hoping that the `SETTINGS` variable is not global for jobs when I found out the inconsistent output from Parsl, but now it seems exactly the case.. 
I guess this PR needs to be paused/closed unless we find a solution for this issue.

Andrew-S-Rosen:
I'll give it a look and see what we can do. I am not opposed to removing it as a global variable if I can make everything else still work as expected.

Andrew-S-Rosen:
Currently trying to patch multithreading behavior on #2276. I think I'm onto something. I'll let you know if I succeed. 👍

Andrew-S-Rosen:
@honghuikim: I think I got it! 

When you have a moment, would you be willing to try out the branch in #2277 and see if it works as you expect on your machine? If so, then I will go ahead and merge #2276 (which would fix the concurrency issue with settings), update the current PR, and then we can merge #2163.

honghuikim:
@Andrew-S-Rosen  Amazing! It works great! I tested `bulk_to_slabs_flow` and some other tests with the branch #2277, and now it's working just as we expected. It seems like we are now setting `settings` attribute for each thread?  Thank you very much for the fix!

Andrew-S-Rosen:
@honghuikim: Wonderful! Thank you for your testing! Yes, you are right --- the main changes are in the `__init__.py` [here](https://github.com/Quantum-Accelerators/quacc/pull/2277/files#diff-946e3fdb0961ed618a71033ca644b84c4496c41e5dcf890b7280043d6b86e067R44-R86). There, I take advantage of a concept known as [thread-local data](https://docs.python.org/3/library/threading.html#thread-local-data). This was new to me, but it seems to work well.

I'll go ahead and start merging things, after adding a few more tests just to be sure. Thank you again for your enormous help. I am incredibly grateful for it!

Andrew-S-Rosen:
I'm now going to go ahead and close this PR in favor of https://github.com/Quantum-Accelerators/quacc/pull/2277 since there are a few minor changes I made there to make tests pass. Your contributions are still marked in the commits and will be recognized when merged, so there should be no change in credit --- just a change in which PR is merged.

honghuikim:
@Andrew-S-Rosen Sounds good! Thank you for wrapping this up.

Andrew-S-Rosen:
Of course! All merged. I'm going to wait a couple of days before minting a new release just in case something unusual pops up, but things are looking good!

PR #2218, opened 2024-06-06
Support in run_calcs for models that implement arbitrary properties, and support for MLPs that don't have forces

zulissimeta:
## Summary of Changes

Small change to `run_calc` so that arbitrary properties can be requested (including but not limited to energy/forces), and small change to the mlp.core.static_job so that forces are not already required for those calculators. This is also useful for ML models that predict properties beyond energy/forces. 

Currently, if a MLP calculator does not implement forces, mlp.core.static_job will throw a PropertyNotImplemented error.

### DIFF
--- a/src/quacc/recipes/mlp/core.py
+++ b/src/quacc/recipes/mlp/core.py
@@ -21,7 +21,10 @@ if TYPE_CHECKING:
 
 @job
 def static_job(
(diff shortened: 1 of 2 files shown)


buildbot-princeton:

Andrew-S-Rosen:
Thanks, @zulissimeta! Seems logical enough. Setting this to auto-merge (I just fixed the docs build error, which was because the `get_forces` docstring was still left behind).

PR #2242, opened 2024-06-12
Vibrational analysis recipe for VASP

benshi97:
## Summary of Changes

\>> New VASP recipe for performing vibrational analysis (using ASE Vibrations) to obtain thermodynamic properties.


buildbot-princeton:

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,72 @@ def non_scf_job(
...
+    VibThermoSchema
+        Dictionary of results, specified in [quacc.schemas.ase.summarize_vib_and_thermo][].
+        See the type-hint for the data structure.
+    """
+    calc_defaults = {
+        "ediff": 1e-8,
+        "ibrion": -1,
+        "isym": 0,
+        "istart": 1,
+        "algo": "all",
```
Andrew-S-Rosen, line 376: We probably don't want to enforce algo = All. I use it a lot, but it's not the best algorithm for all materials and isn't specific to getting an accurate vibrational frequency analysis (to the best of my knowledge). Let's drop this one.

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,72 @@ def non_scf_job(
...
+
+    Returns
+    -------
+    VibThermoSchema
+        Dictionary of results, specified in [quacc.schemas.ase.summarize_vib_and_thermo][].
+        See the type-hint for the data structure.
+    """
+    calc_defaults = {
+        "ediff": 1e-8,
+        "ibrion": -1,
```
Andrew-S-Rosen, line 373: We can remove the ibrion = -1 setting since nsw = 0 is all we need to enforce a static calculation. Internally, quacc will [automatically](https://github.com/Quantum-Accelerators/quacc/blob/90065b4e557dd34cada408bfccaaed9725f8f6ab/src/quacc/calculators/vasp/params.py#L267-L271) drop ibrion and related flags if NSW = 0 anyway.

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,72 @@ def non_scf_job(
...
+        "isym": 0,
+        "istart": 1,
```
Andrew-S-Rosen, line 375: Generally, we rely on VASP to automatically determine the appropriate value for ISTART. If a WAVECAR exists in the directory, it will be automatically set to 1, so no need for us to enforce that.

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,72 @@ def non_scf_job(
...
+        "algo": "all",
+        "lcharg": False,
+        "lwave": True,
+        "nsw": 0,
+        "ismear": 0,
```
Andrew-S-Rosen, line 380: In this case, we should be setting ISMEAR = -5 since that's how you ensure you get highly accurate forces that are needed for a vibrational frequency analysis.

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,72 @@ def non_scf_job(
...
+        "sigma": 0.05,
```
Andrew-S-Rosen, line 381: If we set ISMEAR = -5, then we remove sigma since it is not used for ISMEAR < 0.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
```diff
@@ -848,3 +849,14 @@ def test_mp_relax_flow_custom(tmp_path, patch_nonmetallic_taskdoc):
             job_params={"mp_metagga_relax_job": {"nsw": 0}},
         )
         assert output["relax2"]["parameters"]["nsw"] == 0
+
+
+def test_freq_job():
+    atoms = molecule("H2")
+    atoms.pbc = True
+    atoms.set_cell([3, 3, 3])
```
Andrew-S-Rosen, line 857: Let's instead do

```python
atoms.center(vacuum=5, axis=2)
```

or something like that which is more akin to what one would do in reality.

Andrew-S-Rosen (commented):
Hi @benshi97! Overall, this looks plenty good to me. 

Two things:
1. I have a variety of comments related to the choice of settings. Please review them and make the changes as you see fit.
2. Since you likely don't want to vibrate every single atom in the case of something like an adsorbate on a surface, I'd suggest adding an additional keyword argument (e.g. `displaced_indices`) that is `list[int]` and will apply a `FixAtoms` constraint on the `Atoms` object accordingly to ensure that the ASE `Vibrations` module only vibrates the atoms you want vibrated. The only tricky thing is that we need to make sure that the resulting thermochemistry is computed appropriately and also that the vibrations schema is appropriate (i.e. just making sure that everything is handled as it should be).

A few other small things:
- Could you update the recipe list? https://github.com/Quantum-Accelerators/quacc/blob/main/docs/user/recipes/recipes_list.md
- Feel free to add your name here (obviously it'll be added with the other workflows you're doing): https://github.com/Quantum-Accelerators/quacc/blob/main/docs/about/contributors.md

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,61 @@ def non_scf_job(
...
+        `None` to remove a pre-existing key entirely. For a list of available
+        keys, refer to [quacc.calculators.vasp.vasp.Vasp][].
+
+    Returns
+    -------
+    VibThermoSchema
+        Dictionary of results, specified in [quacc.schemas.ase.summarize_vib_and_thermo][].
+        See the type-hint for the data structure.
+    """
+    calc_defaults = {"ediff": 1e-8, "isym": 0, "lcharg": False, "lwave": True, "nsw": 0}
```
Andrew-S-Rosen, line 371: Would you be okay with `ediff = 1e-7`? This is what Materials Project uses for accurate forces, so it'd be quite consistent. But if you disagree, that's fine. 

Also, once the Custodian patch is merged, we can/should add `sigma=-5` back in so we can ensure the forces are very accurate.

src/quacc/recipes/vasp/_base.py
```diff
@@ -120,3 +123,74 @@ def run_and_summarize_opt(
...
+    energy: float = 0.0,
+    temperature: float = 298.15,
+    pressure: float = 1.0,
+    thermo_method: Literal["ideal_gas", "harmonic"] = "ideal_gas",
+    preset: str | None = None,
+    calc_defaults: dict[str, Any] | None = None,
+    calc_swaps: dict[str, Any] | None = None,
+    vib_kwargs: dict[str, Any] | None = None,
+    copy_files: SourceDirectory | dict[SourceDirectory, Filenames] | None = None,
+    additional_fields: dict[str, Any] | None = None,
```
Andrew-S-Rosen, line 139: Super minor, but could you swap the order so it matches the function above?

src/quacc/recipes/vasp/_base.py
```diff
@@ -120,3 +123,74 @@ def run_and_summarize_opt(
...
+        Dictionary of custom kwargs for [quacc.runners.ase.Runner.run_vib][]
+    copy_files
+        Files to copy (and decompress) from source to the runtime directory.
+
+    Returns
+    -------
+    VibThermoSchema
+        Dictionary of results, specified in [quacc.schemas.ase.summarize_vib_and_thermo][].
+        See the type-hint for the data structure.
+    """
```
Andrew-S-Rosen, line 172: Could you add the missing docstring for `additional_fields`?

src/quacc/recipes/vasp/_base.py
```diff
@@ -120,3 +123,74 @@ def run_and_summarize_opt(
...
+    calc_flags = recursive_dict_merge(calc_defaults, calc_swaps)
+
+    calc = Vasp(atoms, preset=preset, **calc_flags)
+    vibrations = Runner(atoms, calc, copy_files=copy_files).run_vib(
+        vib_kwargs=vib_kwargs
+    )
+    if thermo_method == "harmonic":
+        thermo_analysis = ThermoRunner(
+            atoms, vibrations.get_frequencies(), energy=energy
+        ).run_harmonic()
```
Andrew-S-Rosen, line 184: Could you also add one very small test to run a VASP recipe with the harmonic method so this gets covered? You have already tested the harmonic thermo functionality elsewhere, so this can be pretty minimal --- just to make sure we don't break it later.

src/quacc/recipes/vasp/core.py
```diff
@@ -320,3 +325,63 @@ def non_scf_job(
...
+    preset
+        Preset to use from `quacc.calculators.vasp.presets`.
+    energy
+        Potential energy in eV. If 0, then the output is just the correction.
+    temperature
+        Temperature in Kelvins.
+    pressure
+        Pressure in bar.
+    vib_kwargs
+        Dictionary of kwargs for the [ase.vibrations.Vibrations][] class.
```
Andrew-S-Rosen, line 358: Could you add a docstring for the missing `thermo_method`?

src/quacc/runners/thermo.py
```diff
@@ -103,3 +103,27 @@ def run_ideal_gas(self, spin_multiplicity: int | None = None) -> IdealGasThermo:
...
         )
+
+    def run_harmonic_thermo(self) -> HarmonicThermo:
+        """
+        Create a HarmonicThermo object for a molecule from a given vibrational analysis.
+        This works for free gases, solids and adsorbates.
+
+        Returns
+        -------
+        HarmonicThermo object
```
Andrew-S-Rosen, line 114: To make sure this renders appropriately, could you do

```
Returns
--------
HarmonicThermo
    The ASE `HarmonicThermo` object
```

src/quacc/schemas/ase.py
```diff
@@ -255,12 +255,18 @@ def summarize_vib_and_thermo(
...
-        igt,
-        temperature=temperature,
-        pressure=pressure,
-        charge_and_multiplicity=charge_and_multiplicity,
-    )
+
+    if isinstance(thermo_analysis, HarmonicThermo):
+        thermo_task_doc = _summarize_harmonic_thermo(
+            ht=thermo_analysis, temperature=temperature, pressure=pressure
+        )
```
Andrew-S-Rosen, line 262: This line will get covered when you make the minimal VASP harmonic thermo test and/or when you add a brief test for `summarize_vib_and_thermo` directly as commented below.

src/quacc/schemas/ase.py
```diff
@@ -437,3 +443,55 @@ def _summarize_ideal_gas_thermo(
...
+                temperature, verbose=settings.DEBUG
+            ),
+            "entropy": ht.get_entropy(temperature, verbose=settings.DEBUG),
+            "zpe": ht.get_ZPE_correction(),
+        }
+    }
+
+    atoms_metadata = {}
+
+    return atoms_metadata | inputs | results
```
Andrew-S-Rosen, line 497: If we stick with `{}` then there's no need to concatenate it to the other dictionaries. However, I think we should probably be generating an `atoms_metadata` like what was done for the ideal gas case because otherwise the output dictionary from a run will not have the resulting `Atoms` object in it, which means the user can't do a subsequent job. All outputs in quacc need to support doing `results["atoms"]` basically.

I'm guessing this was done because `HarmonicThermo` does not have an `.atoms` attribute. I would suggest adding an `atoms: Atoms = None` keyword argument to `summarize_vib_and_thermo` that you can use to pass in the output `Atoms` object in your VASP recipe, where it will be turned into metadata here.

Make the `Atoms` object a required positional argument to `_summarize_harmonic_thermo` though so other developers don't forget to pass it in.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
```diff
@@ -849,3 +850,12 @@ def test_mp_relax_flow_custom(tmp_path, patch_nonmetallic_taskdoc):
...
         assert output["relax2"]["parameters"]["nsw"] == 0
+
+
+def test_freq_job():
+    atoms = molecule("H2")
+    atoms.pbc = True
+    atoms.center(vacuum=1)
+    output = freq_job(atoms, kpts=(1, 1, 1))
+    assert output["parameters"]["ediff"] == 1e-07
+    assert len(output["results"]["vib_freqs_raw"]) == 3 * len(atoms)
```
Andrew-S-Rosen, line 861: Could we also test the "actual" number of vibrational modes (3N-5 for linear)? Just to make sure that's being done correctly...

tests/core/schemas/test_ase.py
```diff
@@ -345,6 +347,25 @@ def test_summarize_ideal_gas_thermo(tmp_path, monkeypatch):
...
+    assert results["parameters_thermo"]["vib_freqs"] == [0.34 / invcm]
+    assert results["results"]["energy"] == 0
+    assert_close(results["results"]["helmholtz_energy"], 0.16999995401497991, rtol=1e-5)
+    assert_close(results["results"]["internal_energy"], 0.1700006085385999, rtol=1e-5)
+    assert_close(results["results"]["entropy"], 2.1952829783392438e-09, rtol=1e-5)
+    assert_close(results["results"]["zpe"], 0.17, rtol=1e-5)
+
+    # test document can be jsanitized and decoded
+    d = jsanitize(results, strict=True, enum_values=True)
+    MontyDecoder().process_decoded(d)
```
Andrew-S-Rosen, line 366: This is great. Could you also test the "parent" `summarize_harmonic_thermo` function once so we make sure it's working as expected?

Andrew-S-Rosen (commented):
Hi @benshi97! Thank you for these updates. This looks splendid! I only have a few very minor comments. I'll be happy to merge thereafter!

tests/core/schemas/test_ase.py
benshi97: I have added an additional unit test for `summarize_vib_and_thermo` which tests for both the `_summarize_harmonic_thermo` and `_summarize_ideal_gas_thermo` situation.

tests/core/recipes/vasp_recipes/mocked/test_vasp_recipes.py
benshi97: It seems that the symmetry isn't being read for the 'ideal_gas' case; so the "vib_freqs" doesn't differ from "vib_freq_raw"

src/quacc/schemas/vasp.py
Andrew-S-Rosen, line 1: Hi @benshi97: Thanks again! It looks like there are a total of 5 test failures, only one of which is due to the incorrect number of vibrational modes. At least 3 of those failures are due to this edit.

Could you ensure the tests pass, except the one you called out? Afterwards, I'll be happy to fix the symmetry detection and will merge this in. Looks great!

Andrew-S-Rosen:
Closing in favor of https://github.com/Quantum-Accelerators/quacc/pull/2362.

PR #2297, opened 2024-06-21
job patterns for partitioning lists and mapping onto them

zulissimeta:
## Summary of Changes

This draft PR adds job patterns that are common in high-throughput workflows. When running many jobs on the same flow (say `result = map(my_job, range(1000))`) , many workflow managers will get sad (network/db/load issues) with too many jobs. This PR is inspired by the dask bag partitions and map operator. 

For example:
```
@job
def testjob(**kwargs):
    print(kwargs)

@flow
def testflow():
    num_partitions = 2
    result = map_partitioned_lists(
        testjob,
        num_partitions=num_partitions,
        test_arg_1=partition([1,2,3,4,5], num_partitions),
        test_arg_2=partition(["a", "b", "c","d","e"], num_partitions),
    )

testflow()
```

should yield:
```
{'test_arg_1': 1, 'test_arg_2': 'a'}
{'test_arg_1': 2, 'test_arg_2': 'b'}
{'test_arg_1': 3, 'test_arg_2': 'c'}
{'test_arg_1': 4, 'test_arg_2': 'd'}
{'test_arg_1': 5, 'test_arg_2': 'e'}
```
but run in only two jobs (instead of 5). 

In addition, by using a specified number of partitions, logic flows can quickly move from one step to the next without waiting for any intermediate results.
```
many_results = generate_many_objects()

# Cannot continue until many_results is generated as the number of jobs to be generated is unknown
final_results = [analysis_job(result) for result in results]
```
while 
```
many_results = generate_many_objects()
partitions_results = partition(many_results, num_partitions=10)
final_results = mapped_partitioned_lists(analysis_job, result=partitioned_results, num_partitions=10)
```
is fine since it is clear there is a first job, a second partition job that yields 10 tasks, and a final mapping job that has 10 jobs (one per partition).

### DIFF
--- a/src/quacc/wflow_tools/job_patterns.py
+++ b/src/quacc/wflow_tools/job_patterns.py
@@ -1,0 +1,157 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from quacc.wflow_tools.decorators import job
+from quacc.wflow_tools.customizers import strip_decorator
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+@job
+def partition(a, n) -> list[Any]:
+    """
+    Given a list, partition it into n roughly equal lists
+
+    Parameters
+    ----------
+    a
+        the list to partition
+    n
+        the number of partitions to output
+
+    Returns
+    -------
+    list[Any]
+        n lists constructed from a
+    """
+    k, m = divmod(len(a), n)
+    return [a[i * k + min(i, m) : (i + 1) * k + min(i + 1, m)] for i in range(n)]
+
+
+def map_partitioned_lists(
+    func: callable,
+    num_partitions: int,
+    unmapped_kwargs: dict[str, Any] | None = None,
+    **mapped_kwargs: dict[str, list[list[Any]]],
+):
+    """
+    Given list-of-lists parameters (say a list of batches that we want to map over),
+    apply func to each element of each list
+
+    For example:
+    @job
+    def testjob(**kwargs):
+        print(kwargs)
+
+    @flow
+    def testflow():
+        num_partitions = 2
+        result = map_partitioned_lists(
+            testjob,
+            num_partitions=num_partitions,
+            test_arg_1=partition([1,2,3,4,5], num_partitions),
+            test_arg_2=partition(["a", "b", "c","d","e"], num_partitions),
+        )
+
+    testflow()
+
+    should yield:
+    {'test_arg_1': 1, 'test_arg_2': 'a'}
+    {'test_arg_1': 2, 'test_arg_2': 'b'}
+    {'test_arg_1': 3, 'test_arg_2': 'c'}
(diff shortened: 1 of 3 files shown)


buildbot-princeton:

src/quacc/wflow_tools/job_patterns.py
```diff
@@ -0,0 +1,155 @@
...
+        for i in range(num_partitions)
+    ]
+
+
+@job
+def map_partition(
+    func: callable, unmapped_kwargs: dict[str, Any] | None = None, **mapped_kwargs
+) -> list[Any]:
+    """
+    Wrap a function to handle special Parsl arguments.
```
Andrew-S-Rosen, line 103: Wrong docstring?

tests/core/wflow/test_job_patterns.py
Andrew-S-Rosen, line 1: Let's also add a workflow engine (e.g. Prefect) test for good measure.

Andrew-S-Rosen (commented):
Thanks, @zulissimeta! This is an interesting PR.

If I understand correctly, it seems like this is logic that would need to be put within the `@flow`s themselves, right? So, the ideal use case here is someone is importing a pre-made `@job` (e.g. `quacc.recipes.mlp.core import relax_job`) and making a custom `@flow` for themselves from that, right? It would perhaps be a bit difficult to justify when/where to add this logic within pre-made `@flow`s in quacc itself (i.e. in `quacc.recipes`).

Some (non-substantial) comments below while I await your reply.

zulissimeta:
> Thanks, @zulissimeta! This is an interesting PR.
> 
> If I understand correctly, it seems like this is logic that would need to be put within the `@flow`s themselves, right? So, the ideal use case here is someone is importing a pre-made `@job` (e.g. `quacc.recipes.mlp.core import relax_job`) and making a custom `@flow` for themselves from that, right? It would perhaps be a bit difficult to justify when/where to add this logic within pre-made `@flow`s in quacc itself (i.e. in `quacc.recipes`).
> 
> Some (non-substantial) comments below while I await your reply.

Yes, exactly! If you want a flow that does many jobs in parallel (for example, if you made a bulk_to_adsorbates_flow that generated hundreds of possible adsorbate configuration and had ML potentials to make relaxations fast), this would be helpful. 

Partitioning and batching would also be helpful if doing lots of inference; then you could partition a list, and apply a function to take batches and do ML inference quickly, rather than running one MLP relaxation as a separate job.

Andrew-S-Rosen:
Got it, thanks! Since it is pretty independent from existing recipe logic, I don't have much reservation about this. We will just want to add a brief section to the documentation somewhere highlighting how it can be used since it's somewhat of an "advanced" (but useful!) feature. I am happy to take care of the docs though.

Andrew-S-Rosen:
Thanks! Setting this to auto-merge now.

Issue #782, opened 2023-08-26
Address all `mypy` type hinting issues


arosen93:

abastola0:
@Andrew-S-Rosen  can you please assign me on this issue?

jayanthkoduri:
@Andrew-S-Rosen I need to contribute to Open Source as part of my Course work.
Can you please assign me this.

Andrew-S-Rosen:
@abastola0, @jayanthkoduri: Feel free to work on a PR. I don't assign issues. This issue is large enough where it won't be possible to solve it completely with one PR.

If you both plan to work on this, I recommend commenting here about which modules you plan to work on so you don't overlap. Like I mentioned, don't try to do them all.

ugognw:
This is a large undertaking! Running `mypy --strict` locally, I get ~700 errors. If you're interested in being able to run `mypy --strict` right away, I can create a PR with all outstanding errors suppressed. (Not by hand 😅, I've written a [tool](https://github.com/ugognw/mypy-upgrade?tab=readme-ov-file) that can automatically suppress `mypy` errors with `# type: ignore [error-code]` comments.) You can then start using `mypy` in CI to prevent introducing new type hint errors while you clear the old ones.

Andrew-S-Rosen:
Thank you! :)

PR #2314, opened 2024-06-30
Address some mypy type hinting issues

AitElBadaoui:
## Summary of Changes

This merge request addresses some of the type hinting issues identified by MyPy.
Ref: https://github.com/Quantum-Accelerators/quacc/issues/782.

### Requirements

### Notes

### DIFF
--- a/src/quacc/runners/prep.py
+++ b/src/quacc/runners/prep.py
@@ -138,7 +138,7 @@ def calc_cleanup(
         symlink_path.unlink(missing_ok=True)
 
 
-def terminate(tmpdir: Path, exception: Exception) -> Exception:
+def terminate(tmpdir: Path, exception: Exception) -> None:
     """
     Terminate a calculation and move files to a failed directory.
 
--- a/src/quacc/runners/prep.py
+++ b/src/quacc/runners/prep.py
@@ -138,7 +138,7 @@ def calc_cleanup(
         symlink_path.unlink(missing_ok=True)
 
 
-def terminate(tmpdir: Path, exception: Exception) -> None:
+def terminate(tmpdir: Path | str, exception: Exception) -> None:
     """
     Terminate a calculation and move files to a failed directory.
 
@@ -158,6 +158,7 @@ def terminate(tmpdir: Path, exception: Exception) -> None:
     Exception
         The exception that caused the calculation to fail.
     """
+    tmpdir = Path(tmpdir)
     settings = get_settings()
     job_failed_dir = tmpdir.with_name(tmpdir.name.replace("tmp-", "failed-"))
     tmpdir.rename(job_failed_dir)
--- a/src/quacc/runners/prep.py
+++ b/src/quacc/runners/prep.py
@@ -138,7 +138,7 @@ def calc_cleanup(
         symlink_path.unlink(missing_ok=True)
 
 
-def terminate(tmpdir: Path | str, exception: Exception) -> Exception:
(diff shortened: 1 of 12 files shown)


buildbot-princeton:

AitElBadaoui:
Hello @Andrew-S-RosenThe,
The missing line in the coverage is because none of the jobs call the run_and_summarize function with a None value in option_defaults. I'm not sure if it's worth it to mock a job just to cover this one line of code. Please let me know what you think.

To resolve the remaining MyPy issues, I need to understand the objective of this MR: https://github.com/Quantum-Accelerators/quacc/pull/1903. The changes introduced have led to multiple MyPy errors due to assigning tuples as default values to variables of different data types. Is _DEFAULT_SETTING meant to always be empty ?

Thank you !

Andrew-S-Rosen:
Hi @AitElBadaoui! Thank you so much for putting together this PR. I really appreciate it!! These all look like great changes and, in fact, may even reduce some bugs.

As for the test coverage, don't worry about it for this PR. No need to worry about test coverage for type hinting changes here.

As for PR #1903, I can see how that might introduce some problems in hindsight. The reasoning behind the PR was that I thought `_DEFAULT_SETTING` (rather than `None`) would make it clearer to the end-user that a default setting was being used (quacc has a set of [default settings](https://quantum-accelerators.github.io/quacc/user/settings/settings.html) it imports at runtime). However, I see the problem with the type hinting. I chose to use `()` as a placeholder because `()` was simple and immutable. I also thought this approach would be nice because then `None` could actually mean something distinct from using the defaults. But the problem is that something like `qchem_cmd: str = _DEFAULT_SETTING` is clearly going to be flagged as a problem by `mypy` since `_DEFAULT_SETTING` is a tuple (even though the actual setting being used at runtime is indeed a `str`).

I guess since the type hint is now "wrong", would you recommend that I switch back to `None` in those cases? I can do that in a separate PR if you think that is the best approach.

src/quacc/runners/prep.py
```diff
@@ -138,7 +138,7 @@ def calc_cleanup(
         symlink_path.unlink(missing_ok=True)
 
 
-def terminate(tmpdir: Path | str, exception: Exception) -> Exception:
+def terminate(tmpdir: Path, exception: Exception) -> None:
```
Andrew-S-Rosen, line 141: For this one, could you update the docstring to be

```
Returns
-------
None
```

Andrew-S-Rosen, line 141: It'd be nice to also support `Path | str` for `tmpdir`. Maybe the better option is to leave the original type hint for that and instead do `tmpdir = Path(tmpdir)` in the function itself?

Andrew-S-Rosen (commented):
As for a review, here are my (minor) comments. Otherwise, this is in great shape for merging!

src/quacc/runners/prep.py
AitElBadaoui: That sounds good, I'll do that.

AitElBadaoui:
Yes, I agree that we should switch back to None. After that, I can create a new PR to address other mypy issues. 
Thank you for the review!

Andrew-S-Rosen:
@AitElBadaoui: wonderful! I will take are of that and merge this PR when I'm back at my computer in an hour or so. 👍

Andrew-S-Rosen:
Merging this one. Thanks!

PR #2315, opened 2024-06-30
Update default type handling

Andrew-S-Rosen:


AitElBadaoui:
@Andrew-S-Rosen, yes it works! Thank you for that. Just a small suggestion: What do you think about creating the class `class _DefaultSettingType: pass` in a single file like `type.py` or `typing.py`, and then importing it in all other modules? This way, we can avoid updating multiple files if the type needs to be changed in the future and also prevent code duplication.

Andrew-S-Rosen:
Excellent idea! I was thinking about this but wasn't sure where to put it. Great suggestion. I'll take care of that shortly :)

PR #2326, opened 2024-07-03
Fix mypy issues in calculators and schema modules

AitElBadaoui:
## Summary of Changes

Fixed mypy issues in the calculators and schema modules.

FYI: @Andrew-S-Rosen, we reduced the number of mypy errors from 320 to 160. There are still over 130 errors in `types.py`, which I will try to fix in another PR.

#### The remaining issues are linked to design problems such as:

- The `Atoms` used in Quacc for the moment overload the `ASE Atoms` class with attributes and functions like `charges`, `as_dict()`, and `from_dict()`, which mypy does not accept. Should we ignore these, or create a `QuaccAtom` class with the necessary attributes and functions? what do you think ?

### Requirements

### Notes

### DIFF
--- a/src/quacc/atoms/phonons.py
+++ b/src/quacc/atoms/phonons.py
@@ -17,6 +17,7 @@ if has_phonopy:
 
 if TYPE_CHECKING:
     from ase.atoms import Atoms
+    from numpy._typing import NDArray
 
     if has_phonopy:
         from phonopy.structure.atoms import PhonopyAtoms
@@ -27,7 +28,7 @@ def get_phonopy(
     atoms: Atoms,
     min_lengths: float | tuple[float, float, float] | None = None,
     supercell_matrix: (
-        np.ndarray
+        NDArray
         | tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]]
         | None
     ) = None,
--- a/src/quacc/atoms/phonons.py
+++ b/src/quacc/atoms/phonons.py
@@ -27,7 +27,9 @@ def get_phonopy(
     atoms: Atoms,
     min_lengths: float | tuple[float, float, float] | None = None,
     supercell_matrix: (
-        tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]] | None
+        np.ndarray
+        | tuple[tuple[int, int, int], tuple[int, int, int], tuple[int, int, int]]
+        | None
     ) = None,
     symprec: float = 1e-5,
     displacement: float = 0.01,
--- a/src/quacc/atoms/phonons.py
+++ b/src/quacc/atoms/phonons.py
@@ -17,7 +17,7 @@ if has_phonopy:
 
 if TYPE_CHECKING:
     from ase.atoms import Atoms
-    from numpy._typing import NDArray
+    from numpy.typing import NDArray
 
     if has_phonopy:
         from phonopy.structure.atoms import PhonopyAtoms
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -195,7 +195,7 @@ class EspressoTemplate(EspressoTemplate_):
 
         Path(directory, f"{prefix}.EXIT").touch()
 
-    def read_results(self, directory: os.PathLike | str) -> dict[str, Any]:
+    def read_results(self, directory: os.PathLike) -> dict[str, Any]:
         """
         The function that should be used instead of the one in ASE EspressoTemplate to
         read the output file. It calls a customly defined read function. It also adds
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -215,7 +215,7 @@ class EspressoTemplate(EspressoTemplate_):
         results = {}
         if self.binary == "pw":
             atoms = read(Path(directory) / self.outputname, format="espresso-out")
-            results = dict([atom.calc.properties() for atom in atoms])
+            results = dict(atoms.calc.properties())
         elif self.binary in ["ph", "phcg"]:
             with Path(directory, self.outputname).open() as fd:
                 results = read_espresso_ph(fd)
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -215,7 +215,7 @@ class EspressoTemplate(EspressoTemplate_):
         results = {}
         if self.binary == "pw":
             atoms = read(Path(directory) / self.outputname, format="espresso-out")
-            results = dict(atoms.calc.properties())
+            results = dict([atom.calc.properties() for atom in atoms])
         elif self.binary in ["ph", "phcg"]:
             with Path(directory, self.outputname).open() as fd:
                 results = read_espresso_ph(fd)
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -195,7 +195,7 @@ class EspressoTemplate(EspressoTemplate_):
 
         Path(directory, f"{prefix}.EXIT").touch()
 
-    def read_results(self, directory: Path | str) -> dict[str, Any]:
+    def read_results(self, directory: os.PathLike | str) -> dict[str, Any]:
         """
         The function that should be used instead of the one in ASE EspressoTemplate to
         read the output file. It calls a customly defined read function. It also adds
--- a/src/quacc/calculators/espresso/espresso.py
+++ b/src/quacc/calculators/espresso/espresso.py
@@ -222,7 +222,8 @@ class EspressoTemplate(EspressoTemplate_):
         elif self.binary == "dos":
             with Path(directory, "pwscf.dos").open() as fd:
                 lines = fd.readlines()
-                fermi = float(re.search(r"-?\d+\.?\d*", lines[0])[0])
+                match = re.search(r"-?\d+\.?\d*", lines[0])
+                fermi = float(match.group(0)) if match else None
                 dos = np.loadtxt(lines[1:])
             results = {"dos_results": {"dos": dos, "fermi": fermi}}
         elif self.binary == "projwfc":
@@ -279,7 +280,7 @@ class EspressoTemplate(EspressoTemplate_):
 
         return parameters
 
-    def _sanity_checks(self, parameters: dict[str, Any]) -> None:
+    def _sanity_checks(self, parameters: dict[str, Any]) -> dict[str, Any]:
         """
         Function that performs sanity checks on the input_data. It is meant
         to catch common mistakes that are not caught by the espresso binaries.
@@ -293,7 +294,7 @@ class EspressoTemplate(EspressoTemplate_):
         Returns
         -------
         dict
-            The modified parameters dictionary.
+            The modified dictionary parameters.
         """
         input_data = parameters.get("input_data", {})
 
--- a/src/quacc/calculators/vasp/vasp.py
+++ b/src/quacc/calculators/vasp/vasp.py
@@ -26,7 +26,7 @@ from quacc.schemas.prep import set_magmoms
 from quacc.utils.dicts import sort_dict
 
 if TYPE_CHECKING:
-    from typing import Literal
+    from typing import Any, Literal
 
     from ase.atoms import Atoms
 
@@ -149,7 +149,7 @@ class Vasp(Vasp_):
         self.kwargs = kwargs
 
         # Initialize for later
-        self.user_calc_params = {}
+        self.user_calc_params: dict[str, Any] = {}
 
         # Cleanup parameters
         self._cleanup_params()
@@ -312,4 +312,5 @@ class Vasp(Vasp_):
             run_custodian(directory=directory)
             return 0
 
-        return subprocess.call(command, shell=True, stdout=out, cwd=directory)
+        stdout = Path.open(Path(out), "w") if out else None
+        return subprocess.call(command, shell=True, stdout=stdout, cwd=directory)
--- a/src/quacc/calculators/vasp/vasp.py
+++ b/src/quacc/calculators/vasp/vasp.py
@@ -312,5 +312,8 @@ class Vasp(Vasp_):
             run_custodian(directory=directory)
             return 0
 
-        stdout = Path.open(Path(out), "w") if out else None
-        return subprocess.call(command, shell=True, stdout=stdout, cwd=directory)
+        if out:
+            with Path(out).open("w") as file:
+                return subprocess.call(command, shell=True, stdout=file, cwd=directory)
+        else:
+            return subprocess.call(command, shell=True, stdout=None, cwd=directory)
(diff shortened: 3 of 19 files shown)


buildbot-princeton:

src/quacc/schemas/ase.py
```diff
@@ -170,10 +169,12 @@ def summarize_opt_run(
 
     # Get trajectory
     if not trajectory:
-        trajectory = read(dyn.trajectory.filename, index=":")
-    trajectory_results = [atoms.calc.results for atoms in trajectory]
+        atoms_trajectory = read(dyn.trajectory.filename, index=":")  # type: ignore[union-attr]
+    else:
+        atoms_trajectory = trajectory
```
Andrew-S-Rosen, line 174: Could we reverse the logic so it's `if trajectory` first and then `else`?

src/quacc/utils/dicts.py
```diff
@@ -95,7 +97,7 @@ def _recursive_dict_pair_merge(
     """
     dict1 = dict1 or ({} if dict1 is None else dict1.__class__())
     dict2 = dict2 or ({} if dict2 is None else dict2.__class__())
-    merged = safe_dict_copy(dict1)
+    merged = safe_dict_copy(dict(dict1))
```
Andrew-S-Rosen, line 100: I guess this and the change on line 184 are needed because `safe_dict_copy` and `sort_dict` are only type hinted for `dict`. Should we be type hinting them for `MutableMapping` instead?

src/quacc/wflow_tools/decorators.py
```diff
@@ -3,19 +3,16 @@
...
+Job = TypeVar("Job", bound=Callable[..., Any])
+Flow = TypeVar("Flow", bound=Callable[..., Any])
+Subflow = TypeVar("Subflow", bound=Callable[..., Any])
 
-if TYPE_CHECKING:
-    from typing import Any, Callable
 
-
-def job(_func: Callable | None = None, **kwargs) -> Job:
+def job(_func: Job | None = None, **kwargs) -> Job | Callable:
```
Andrew-S-Rosen, line 15: I am a bit confused about this one and the ones similar to it. The `@job` decorator is generally applied to a plain function (`Callable`). The output of this is what we are defining internally as a `Job` object. So, in this case, why would `_func` be type hinted as `Job`? I think this would be confusing because it looks like `job` applies to a `Job`, but it applies to a plain old Python function. Same comment for the flow and subflow.

Andrew-S-Rosen (commented):
@AitElBadaoui: Thank you very much for your edits once again!

I have several questions below. Would you be able to address them when you get a chance? Again, most of these are genuine questions rather than critiques.

Andrew-S-Rosen:
> The Atoms used in Quacc for the moment overload the ASE Atoms class with attributes and functions like charges, as_dict(), and from_dict(), which mypy does not accept. Should we ignore these, or create a QuaccAtom class with the necessary attributes and functions? what do you think ?

Interesting. I think this is a fair point. We do that patching [here](https://github.com/Quantum-Accelerators/quacc/blob/3558a7426f130dab287f4b6a5c8dcdf3b9924fa7/src/quacc/__init__.py#L42-L44) and perhaps a few other places. This is a technical necessity for some workflow tools and is very niche. For now, I would be okay with ignoring them to be honest. This whole business is coming from the fact that sometimes an `Atoms` object is being used and sometimes it is ["secretly" an `MSONAtoms` object](https://github.com/materialsproject/pymatgen/blob/4a2c3dffbd6644159ca6b60e7c6ef42de151ace4/src/pymatgen/io/ase.py#L49). Probably not worth digging into here, but... ignoring it seems wise for now.

Andrew-S-Rosen:
@buildbot-princeton test this please

src/quacc/atoms/phonons.py
```diff
@@ -17,6 +17,7 @@
 
 if TYPE_CHECKING:
     from ase.atoms import Atoms
+    from numpy._typing import NDArray
```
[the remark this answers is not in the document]
AitElBadaoui: My bad, I should not use `_typing`. I've updated the code. Thanks for pointing it out!

src/quacc/calculators/espresso/espresso.py
```diff
@@ -195,7 +195,7 @@ def _test_run(parameters: dict[str, Any], directory: Path) -> None:
 
         Path(directory, f"{prefix}.EXIT").touch()
 
-    def read_results(self, directory: Path | str) -> dict[str, Any]:
+    def read_results(self, directory: os.PathLike | str) -> dict[str, Any]:
```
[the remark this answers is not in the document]
AitElBadaoui: There is a difference between the types. The `PathLike` type allows different path representations to be used interchangeably. It includes `str` and `bytes`, as well as instances of `pathlib.Path` and any other custom objects that implement the `__fspath__` method. So, while a `Path` is a `PathLike`, a `PathLike` is not necessarily a Path.

We need to use the same type as the function in the parent class `EspressoTemplate`. This class defines the `read_results` method with a directory attribute of type `PathLike`.

NOTE: I will remove the | str as it is unnecessary.

src/quacc/runners/ase.py
```diff
@@ -199,6 +199,8 @@ def run_opt(
             },
             optimizer_kwargs,
         )
+        if optimizer_kwargs is None:
+            optimizer_kwargs = {}
```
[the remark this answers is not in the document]
AitElBadaoui: The issue stems from `optimizer_kwargs` being typed as `dict[str, None] | None`. Despite `recursive_dict_merge` assigning a dictionary to `optimizer_kwargs`, mypy still considers it potentially `None` because of its initial typing declaration. That's why I added the check

src/quacc/calculators/espresso/espresso.py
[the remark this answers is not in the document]
Andrew-S-Rosen: Thank you! I was a bit confused by the | str, but this makes sense. It was just extraneous.

src/quacc/utils/dicts.py
AitElBadaoui: Yes, that makes sense. I updated all type hints from `dict` to `MutableMapping` in `dicts.py`

src/quacc/calculators/vasp/vasp.py
```diff
@@ -312,4 +312,5 @@ def _run(
             run_custodian(directory=directory)
             return 0
 
-        return subprocess.call(command, shell=True, stdout=out, cwd=directory)
+        stdout = Path.open(Path(out), "w") if out else None
```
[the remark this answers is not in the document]
AitElBadaoui: You right about the file not being automatically closed after the `subprocess.call` execution . I initially thought that `call` would handle the file closure, but it actually requires manual handling.

```suggestion
        if out:
            stdout = Path(out)
            with open(stdout, "w") as file:
                return subprocess.call(command, shell=True, stdout=file, cwd=directory)
        else:
            return subprocess.call(command, shell=True, stdout=None, cwd=directory)
```
Thank you!

AitElBadaoui:
> Interesting. I think this is a fair point. We do that patching [here](https://github.com/Quantum-Accelerators/quacc/blob/3558a7426f130dab287f4b6a5c8dcdf3b9924fa7/src/quacc/__init__.py#L42-L44) and perhaps a few other places. This is a technical necessity for some workflow tools and is very niche. For now, I would be okay with ignoring them to be honest. This whole business is coming from the fact that sometimes an `Atoms` object is being used and sometimes it is ["secretly" an `MSONAtoms` object](https://github.com/materialsproject/pymatgen/blob/4a2c3dffbd6644159ca6b60e7c6ef42de151ace4/src/pymatgen/io/ase.py#L49). Probably not worth digging into here, but... ignoring it seems wise for now.

Sounds good. I've already ignored this type of error.

src/quacc/wflow_tools/decorators.py
AitElBadaoui: If the output of the `_func` is the job, we can change the signature of the job as follows:
```python
def job(_func: Callable[..., Job] | None = None, **kwargs) -> Job | Callable:
```
What do you think ?

Andrew-S-Rosen: The output of `_func` itself is anything. It can be any function at all. The output of `job(_func)` is a `Job`, by definition.

So, the type hinting here is still a bit unclear to me. The `_func` is any `Callable` (i.e. any plain function), so `Callable[..., Job]` does not seem accurate. It would just be `Callable`.

src/quacc/schemas/vasp.py
```diff
@@ -293,7 +292,7 @@ def _bader_runner(path: Path | str) -> BaderSchema:
 
 
 def _chargemol_runner(
-    path: str, atomic_densities_path: str | None = None
+    path: Path | str, atomic_densities_path: str | None = None
```
Andrew-S-Rosen, line 295: Just so I know for the future: should I be replacing all `Path | str` with `os.PathLike` or is that not wise?

src/quacc/calculators/vasp/vasp.py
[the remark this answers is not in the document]
Andrew-S-Rosen: Good to know.

I assume mypy was throwing an error because `stdout` needs to be the file handle and not a `Path` object? I'm surprised it knows that because I don't see any type hinting within `subprocess.py` itself in the standard library. It's also interesting because it was definitely working with the `Path` object, even though the `Popen` docstring says file handle.

src/quacc/runners/ase.py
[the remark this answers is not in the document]
Andrew-S-Rosen: Interesting. I wonder why that is. Is it because we are overwriting `optimizer_kwargs` with a variable of the same name? If so, perhaps we should be using a new name instead? It seems odd that if `recursive_dict_merge` doesn't return `None` in its type hint that mypy would think `optimizer_kwargs` could be `None`.

src/quacc/schemas/vasp.py
AitElBadaoui: I believe the best option is to use  `Path | str` exclusively with Pathlib, and avoid combining `Pathlib` and `os.path`.

src/quacc/runners/ase.py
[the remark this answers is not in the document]
AitElBadaoui: Yes, using a different name should resolve the issue. Another potential solution is to use an assert statement. Since we are confident that optimizer_kwargs will not be None, we can include the following in the code:
```python
assert optimizer_kwargs is not None
```
what do you think ?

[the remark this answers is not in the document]
Andrew-S-Rosen: Personally, I think a different name would be wiser. Perhaps `merged_optimizer_kwargs`? This might be less prone to error than something like `optimizer_kwargs_`.

Having an `assert` statement in the code should really only be there as an internal way to say "this _might_ conceivably happen, but it really shouldn't." In reality, it is actually impossible for it to be `None` based on the way the code is written. There is no scenario where it would be `None` here.

src/quacc/wflow_tools/decorators.py
AitElBadaoui: Ah, that clarifies things. The generic type in the code initially led me to believe the input and output have the same type.  Assuming I'm correct, `_func` is likely executed by various workflow engines (`Covalent`, `Dask`, etc.) based on the configuration. Consequently, both `_func` and `job` end up being of type Callable[..., Any].

Here is the new types:

```python
Job = Callable[..., Any]
Flow = Callable[..., Any]
Subflow = Callable[..., Any]


def job(_func: Callable[..., Any] | None = None, **kwargs) -> Job:
...
```

Andrew-S-Rosen: Ah, yes! This looks a lot more like what I'd expect!

Technically, some of the workflow engines return a class rather than a function object. For instance, parsl returns a `PythonApp` class, jobflow returns its own `Job` class, Dask returns a `Delayed` class, and so on. But I have just been referring to these all as `Job`. In other words, I have been defining `Job` in a somewhat circular way as "a `Job` is simply whatever is returned by `def job`. 😅 

I guess we can see if mypy is okay with this type hint as-is!

src/quacc/wflow_tools/decorators.py
```diff
@@ -192,7 +189,7 @@ def wrapper(*f_args, **f_kwargs):
         return _func
 
 
-def flow(_func: Callable | None = None, **kwargs) -> Flow:
+def flow(_func: Flow | None = None, **kwargs) -> Flow | Callable:
```
Andrew-S-Rosen, line 192: Just a reminder for this one to be corrected.

src/quacc/wflow_tools/decorators.py
```diff
@@ -357,7 +354,7 @@ def workflow(a, b, c):
         return _func
 
 
-def subflow(_func: Callable | None = None, **kwargs) -> Subflow:
+def subflow(_func: Subflow | None = None, **kwargs) -> Subflow | Callable:
```
Andrew-S-Rosen, line 357: Just a reminder for this one to be corrected.

Andrew-S-Rosen:
@buildbot-princeton test this please

src/quacc/wflow_tools/decorators.py
AitElBadaoui: No problem with mypy after this change! All the mypy issues in `decorators.py` are resolved.

src/quacc/utils/dicts.py
Andrew-S-Rosen, line 1: I have just one final question: I'm not 100% sure that `MutableMapping` is even needed anywhere to begin with. Does mypy complain if we just leave everything as `dict` instead of `MutableMapping` altogether? If so, that would allow us to avoid the `dict(d)` calls in a few places.

AitElBadaoui: We need `MutableMapping` for two reasons:
- Compatibility with `Namelist`: The class inherits from `UserDict`, which in turn inherits from `MutableMapping`. 
- Some parts of the code (mentioned in the following lines) rely on the distinction between `MutableMapping` and regular dictionaries
https://github.com/Quantum-Accelerators/quacc/blob/3558a7426f130dab287f4b6a5c8dcdf3b9924fa7/src/quacc/utils/dicts.py#L100-L115

src/quacc/calculators/vasp/vasp.py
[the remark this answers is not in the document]
AitElBadaoui: I don't believe it worked previously because, if you examine the code of the subprocess class below:

```  python
       if stdout is None:
            pass
        elif stdout == PIPE:
        ...
        elif stdout == DEVNULL:
           ...
        elif isinstance(stdout, int):
            ...
        else:
            # Assuming file-like object
            c2pwrite = stdout.fileno() # quacc case
``` 
Using `Path()` for stdout will result in an AttributeError: 'XXX' object has no attribute 'fileno' error.

Mypy determines the types of modules using stubs. For the standard library and Python builtins, these stubs are defined in the [typeshed](https://github.com/python/typeshed) repository, which mypy use.

src/quacc/utils/dicts.py
Andrew-S-Rosen: Ah, thank you!!

Andrew-S-Rosen:
Thank you, @AitElBadaoui!! This should be good to go. I will merge later this evening :)

PR #2367, opened 2024-07-28
Change structure relaxation to optional in Espresso grid_phonon_flow

nerdwarrior:
## Summary of Changes

This PR modifies `quacc.recipes.espresso.phonons.grid_phonon_flow` to only run a `quacc.recipes.espresso.core.relax_job` if `run_relax` is `True`. To account for cases where `run_relax==False`, an option for specifying `prev_outdir` is provided. To account for cases where concurrent jobs are desired, an option for specifying `copy_files` is provided. The docstring has been updated to match.

There does not appear to be a unit test for grid_phonon_flow, and I am not running these edits on a cluster, so the changes are yet untested. The majority are simple copy-pastes from existing jobs. I am not entirely confident that the `copy_files` option will not break things, but it seems that the `ph_init_job` is standard and safely isolated from the q-point parallelization portion.

### Requirements

- [N/A] My PR has relevant, comprehensive [unit tests](https://quantum-accelerators.github.io/quacc/dev/contributing.html#unit-tests).

### Notes


buildbot-princeton:

nerdwarrior:
@Andrew-S-Rosen Tests passed on the patch from our discussion. Could you kindly review?

Andrew-S-Rosen:
@nerdwarrior: Thank you very much for your PR! 

For some background context, there is actually a test for `grid_phonon_flow` but it's in a bit of an unusual spot. The test can be found [here](https://github.com/nerdwarrior/quacc/blob/patch/espresso-standalone-grid-ph/tests/dask/test_espresso_recipes.py). In short, since we are running real Quantum Espresso calculations, it turned out the `grid_phonon_flow` test was taking too long when run serially (perhaps unsurprising). So, I have this test running with Dask as the backend, which will take advantage of the concurrency offered by multiple cores on GitHub Actions and finish in a reasonable time.

The `test_grid_phonon_flow_single` test [here](https://github.com/nerdwarrior/quacc/blob/25df32df5fb6f0f53dc9750b2b3aa95144433030/tests/dask/test_espresso_recipes.py#L37) only takes 11 s to run](https://github.com/Quantum-Accelerators/quacc/actions/runs/10127694142/job/28005790070?pr=2367#step:8:400) and `test_grid_phonon_flow_gamma` only takes 9 s to run. Would you be willing to copy one of those tests and run them with `run_relax=False` to test your new addition? Unfortunately, there is no easy way to `assert` that a specific change was made here, but we can at least make sure it doesn't crash. What are your thoughts?

src/quacc/recipes/espresso/phonons.py
```diff
@@ -342,6 +350,19 @@ def grid_phonon_flow(
...
+    copy_files
+        Source directory or directories to copy files from. If a `SourceDirectory` or a
+        list of `SourceDirectory` is provided, this interface will automatically guess
+        which files have to be copied over by looking at the binary and `input_data`.
+        If a dict is provided, the mode is manual, keys are source directories and values
+        are relative path to files or directories to copy. Glob patterns are supported.
+    prev_outdir
+        The output directory of a previous calculation. If provided, Quantum Espresso
+        will directly read the necessary files from this directory, eliminating the need
+        to manually copy files. The directory will be ungzipped if necessary.
```
Andrew-S-Rosen, line 365: We probably want to think about how to clarify the language here. For instance, if `run_relax == True`, the user-specified `prev_outdir` will do nothing because the `prev_outdir` used in `ph_init_job` is the `pw_job` relaxation job that was run immediately beforehand. Perhaps we can specify in the docstring that this `prev_outdir` keyword argument is specifically used in the call to `ph_init_job` in the scenario where `run_relax` is `False`?

src/quacc/recipes/espresso/phonons.py
```diff
@@ -342,6 +350,19 @@ def grid_phonon_flow(
...
         single q-point.
+    run_relax
+        If True, structure relaxation is performed using [quacc.recipes.espresso.core.relax_job][]
+        before running parallelized phonon jobs.
+    copy_files
+        Source directory or directories to copy files from. If a `SourceDirectory` or a
+        list of `SourceDirectory` is provided, this interface will automatically guess
+        which files have to be copied over by looking at the binary and `input_data`.
+        If a dict is provided, the mode is manual, keys are source directories and values
+        are relative path to files or directories to copy. Glob patterns are supported.
```
Andrew-S-Rosen, line 361: Similarly, this `copy_files` keyword argument does not apply to all jobs. In fact, this keyword argument might potentially cause some confusion for users. If the user runs `grid_phonon_flow` with `run_relax = True` but a user also specifies `copy_files`, then the files being copied to the runtime directory in `ph_init_job` may not be those associated with the relaxed structure used as input (i.e. the output from `pw_job`).

Any suggestions on how you think we should approach this?

src/quacc/recipes/espresso/phonons.py
```diff
@@ -451,9 +472,11 @@ def _grid_phonon_subflow(
         decorators=job_decorators,
     )
 
-    pw_job_results = pw_job(atoms)
+    if run_relax:
+        pw_job_results = pw_job(atoms)
+        prev_outdir = pw_job_results["dir_name"]
 
-    ph_init_job_results = ph_init_job(prev_outdir=pw_job_results["dir_name"])
+    ph_init_job_results = ph_init_job(copy_files=copy_files, prev_outdir=prev_outdir)
```
Andrew-S-Rosen, line 479: We don't need `copy_files` or `prev_outdir` as keyword arguments to `grid_phonon_flow` because these can be specified via the pre-existing `job_params` keyword argument.

For instance, 

```python
grid_phonon_flow(..., job_params={"ph_init_job": {"copy_files": "blah", "prev_outdir": "blahblah"}})
```

So, I would suggest the following changes:
1. Remove the `copy_files` and `prev_outdir` keyword arguments to `grid_phonon_flow`
2. Remove the associated docstrings
3. Remove `copy_files=copy_files` in your call to `ph_init_job`.

Maybe have something  like:

```python
    if run_relax:
        pw_job_results = pw_job(atoms)
        ph_init_job_results = ph_init_job(prev_outdir=pw_job_results["dir_name"])
    else:
        ph_init_job_results = ph_init_job()
```

Thoughts?

nerdwarrior: > After thinking on this more, we don't need `copy_files` or `prev_outdir` as keyword arguments to `grid_phonon_flow` because these can be specified via the pre-existing `job_params` keyword argument.
>
> So, I would suggest the following changes:
> 
> 1. Remove the `copy_files` and `prev_outdir` keyword arguments to `grid_phonon_flow`
> 2. Remove the associated docstrings
> 3. Maybe have something like:
> 
> ```python
>     if run_relax:
>         pw_job_results = pw_job(atoms)
>         ph_init_job_results = ph_init_job(prev_outdir=pw_job_results["dir_name"])
>     else:
>         ph_init_job_results = ph_init_job()  # note to nerdwarrior: copy_files and/or prev_outdir will be implicitly specified here via the `customize_funcs()` function above these lines.
> ```
> 
> Thoughts? If it might help users, we could mention in the docstring for the `run_relax` keyword argument that if the user specifices `run_relax=False`, they may be interested in specifying the `copy_files` and/or `prev_outdir` keyword argument of the `ph_init_job` via `job_params`.

I like this, and I agree that the extra arguments are redundant. However, I still see the issue

> Similarly, this copy_files keyword argument does not apply to all jobs. In fact, this keyword argument might potentially cause some confusion for users. If the user runs grid_phonon_flow with run_relax = True but a user also specifies copy_files, then the files being copied to the runtime directory in ph_init_job may not be those associated with the relaxed structure used as input (i.e. the output from pw_job).

being relevant. In addition to removing the explicit `copy_files` and `prev_outdir` arguments, I propose overriding `run_relax` to `False` if either `copy_files` or `prev_outdir` is specified in `job_params`. My rationale for this is that any user who goes out of their way to specify `copy_files` or `prev_outdir` likely has a specific behavior in mind, and we would not want a default behavior to surprise them. In other words, user-specific parameters should take precedence, and I think we can somewhat safely infer intentions to have `run_relax==False`. 

If this makes sense to you, I will get to implementing it soon. The needed changes are:

- Removing `copy_files` and `prev_outdir` arguments and docstrings
- Checking for `copy_files` and `prev_outdir` in `job_params` and adjusting `run_relax` accordingly
- Updating the `run_relax` docstring to specify that its value will be overridden if either `copy_files` or `prev_outdir` is specified in `job_params`. Further, if `run_relax` is false, `copy_files` or `prev_outdir` should be specified.
  - I suppose the existing phonon jobs can properly handle cases where `copy_files` or `prev_outdir` are not specified, since that relies on the existing phonon job implementation? I can check this when I get the chance.
- Updating the `job_params` docstring to note that `copy_files` or `prev_outdir` can be specified if users want to run concurrent or sequential tasks with only the phonon calculations and not the relaxation (e.g., they have run their own relaxation). We should note that doing so overrides any `run_relax` setting to `False`. Further, we should specify that one of `copy_files` or `prev_outdir` must be provided if `run_relax==False`.
- Copy/sort out the tests

Perhaps this logic is a tad hairy to write out, but my sense is that it is a robust way to get things running in the way that users expect. The tradeoff is that clear documentation is needed, but I think the docstrings should suffice. What do you think?

Andrew-S-Rosen: Thanks for the thorough reply!

In general, my philosophy is to never override a user-specific parameter (even if it is implicitly supplied). Trying to inject too much "smart" logic tends to have a habit of backfiring later on, especially as the code grows in complexity even if it makes sense in this particular case. So, I'd prefer not to modify `run_relax` on-the-fly. Trying to interpret user behavior is a bit of a can of worms that is best avoided.

Personally, I don't see a major issue with the `copy_files` behavior. If a user is manually specifying `copy_files` for the `ph_init_job` _and_ they have kept `run_relax=True`, they will get exactly what they ask for: it will run a relaxation, and then `ph_init_job` will run with the user-specified files copied over. Sure, this might not make much sense from a practical standpoint, but ultimately that is what the user requested and what was ultimately delivered. After all, the user would have to go out of their way to specify `copy_files` for the `ph_init_job` specifically within `job_params`. It is not something that is likely to happen by accident.

For some background context, the motivation for having a `relax_job` as the first step within the phonon flows is that typically one needs to do a "tighter" relaxation than a standard relaxation (in order to not have fictitious imaginary modes). So, even if a user has a "relaxed" structure, it can still be beneficial to re-relax it with tighter settings (this will usually only take a step or two). Of course, this should be made optional though --- hence the motivation for this PR.

> I suppose the existing phonon jobs can properly handle cases where copy_files or prev_outdir are not specified, since that relies on the existing phonon job implementation? I can check this when I get the chance.

They should work fine as-is. The `ph_init_job` (by default) is using the relax job as its `prev_outdir`. Then each `phonon_job` call is relying on `copy_files` determined internally. Generally, these aren't things that users should need to think about. The file handling is all taken care of by quacc. But of course, if you aren't running a `relax_job`, you need to tell the `ph_init_job` some information about where to continue from. You won't need to touch any of the other jobs' file-handling keyword arguments.

P.S. Please don't feel obligated to take my word as final. If you strongly disagree, let me know.

CC'ing @tomdemeyere just for his own context, although his participation here is by no means required.

nerdwarrior: This all makes sense to me. I understand re: "smart" logic. I have a (perhaps unhealthy and needing-untraining) proclivity for including it; I agree that what you said gives a more faithful and predictable output.

I will go ahead and update these soon, as you specified; I just have a packed couple of days ahead of me. Perhaps we should also add a note in the `job_params` docstring that specifying a `prev_outdir` with `relax_job==True` will result in the user-specified `prev_outdir` being overridden (my sense is that this is true given the `recursive_dict_merge` steps that happen each time a job is called?).

Andrew-S-Rosen: Thanks! And no rush at all. If you have time whenever, feel free to take a stab at it. I appreciate the conversation here regardless!

>  Perhaps we should also add a note in the job_params docstring that specifying a prev_outdir with relax_job==True will result in the user-specified prev_outdir being overridden (my sense is that this is true given the recursive_dict_merge steps that happen each time a job is called?).

Any parameters passed via `job_params` (which is a keyword argument for ever `@flow`-decorated function in quacc) will override the defaults. For instance, if you decide you want to run a relaxation with a very tight convergence (let's say `conv_thre` of 1e-20 instead of the default 1e-12 in quacc), you can do so via `job_params` and the call to `recursive_dict_merge` will ensure the user's parameter(s) are always taken as priority. 

However, things are a bit different with `prev_outdir`. Note that `prev_outdir` is not in the list of default parameters. The way that my example was constructed is reproduced below:

```python
    if run_relax:
        pw_job_results = pw_job(atoms)
        ph_init_job_results = ph_init_job(prev_outdir=pw_job_results["dir_name"])
    else:
        ph_init_job_results = ph_init_job()  # note to nerdwarrior: copy_files and/or prev_outdir will be implicitly specified here via the `customize_funcs()` function above these lines.
```

You can see here that if the user specifies `run_relax=True` _and_ decides to (somewhat counter-intuitively) specify a `prev_outdir` for `ph_init_job`, the job will actually crash. That's because `prev_outdir=pw_job_results["dir_name"]` is specified in quacc _and_ the `prev_outdir` supplied by the user will attempt to be used. Since a single keyword argument cannot take two different values, an error will be raised. This is actually not a bad thing because, after all, it does not make sense to do this in the first place. The error is a clue to the user that something is wrong here.

On the flipside, if the user specifies `run_relax=False` and decides to specify a `prev_outdir` for `ph_init_job` via `job_params` (as would be expected), everything will work as-expected. Even though you don't see a `prev_outdir` keyword argument in the `ph_init_job_results = ph_init_job()` call, it is indeed in there if the user specifies it in `job_params. That's because the [`customize_funcs()`](https://github.com/Quantum-Accelerators/quacc/pull/2367/files#diff-8bdf5ce1f1234043fcf2694c61e347b04fdf41671bf88492779cccb06e5da3b9R467-R472) call actually returns a new function with keyword arguments pre-supplied (this is something called a "partial function" --- a bit technical).

tomdemeyere: Is it worth putting in place a security, if `run_relax=False` then the user **must** specify either `copy_files` or `prev_outdir`, otherwise QuAcc complains?

tomdemeyere: Thinking about it a little bit more, wouldn't it make more sense to completely remove the relax step from this flow? Some arguments:

1. This flow is already complicated enough
2. This might trouble beginners even more? (especially the part with folders)
3. Experts (which are definitely the target audience for this flow) already know that a relax is necessary?

@Andrew-S-Rosen What do you think?

Andrew-S-Rosen: > Is it worth putting in place a security, if run_relax=False then the user must specify either copy_files or prev_outdir, otherwise QuAcc complains?

Perhaps. The same question could, perhaps, be asked about the `phonon_job` itself, right? What happens if the user doesn't specify `copy_files` or `prev_outdir`? It presumably assumes that you have everything in the current working directory, but if you are creating a unique folder for every job (as quacc does by default), then that's not going to be do anything useful for you. A reasonable question is: should we require one of these two parameters be set for `phonon_job` itself (and all other jobs that do not take `Atoms` objects as input, such as `dos_job` or `bands_pp_job` or `post_processing_job`). Personally, I would argue that we should add a guard here for all of these "processing"-type jobs. Of course, this also means that if a user forgets to do this in a flow, the underlying job will throw the proper error.

> Thinking about it a little bit more, wouldn't it make more sense to completely remove the relax step from this flow? Some arguments:

I also had been on the fence about this. I can see both arguments. The pro for keeping the relax step is that (in my experience) most users will want to do a tight relaxation beforehand anyway. The con is, of course, increased complexity for a scenario where the user can readily construct a flow themselves of relax+phonons.

I am not opposed to removing the `relax_job` step, but if we do so, we should be consistent about it. We should also do it for `phonon_dos_flow`. And, furthermore, do it in the other phonon recipes where this is done, such as the [toy EMT example](https://github.com/Quantum-Accelerators/quacc/blob/36dde0fb1d70a7083547b29b31ef83c6a9b3300d/src/quacc/recipes/emt/phonons.py#L32). If we go down this road, we should also make sure to update the docstring of the functions to inform the user that they should provide a tightly relaxed structure (maybe even mentioning the `"input_data": {"control": {"forc_conv_thr": 5.0e-5}, "electrons": {"conv_thr": 1e-12}}` as a suggestion).

------

tl;dr -- both posts of yours, @tomdemeyere, are valid to be thinking about.

My vote for the first comment is we should add a guard statement for all Espresso jobs that don't take `Atoms` as inputs to ensure `copy_files` or `prev_outdir` are specified.

My vote for the second comment is... undecided. But I am happy to defer to the consensus on this.

**With the above information in mind, what do you both vote for?**

nerdwarrior: > If we go down this road, we should also make sure to update the docstring of the functions to inform the user that they should provide a tightly relaxed structure (maybe even mentioning the "input_data": {"control": {"forc_conv_thr": 5.0e-5}, "electrons": {"conv_thr": 1e-12}} as a suggestion).

I very much like this suggestion. I think it keeps a reminder of the original intention, while making `grid_phonon_flow` more flexible.

> My vote for the first comment is we should add a guard statement for all Espresso jobs that don't take `Atoms` as inputs to ensure `copy_files` or `prev_outdir` are specified.

I agree; I think this might be an error-prone area for first-time users.

> My vote for the second comment is... undecided. But I am happy to defer to the consensus on this. Maybe I'm leaning slightly towards removal now?

I am leaning toward removal. My sense is that `grid_phonon_flow` is much "less of" a flow than others. To me, running a flow means that I want to run a common back-to-back set of calculations that yield a result I want. A job means running a single calculation. Though `grid_phonon_flow` is a flow in syntax and execution, its usage is much closer to that of a job---basically, a q-point-parallelized version of `phonon_job`. Removing the relaxation would cement this position, where`grid_phonon_flow` can act as a "drop-in" replacement for `phonon_job`. 

Given the above, it no longer seems quite so inconsistent to remove relaxation from only `grid_phonon_flow`, to @Andrew-S-Rosen's point below. 

> I am not opposed to removing the `relax_job` step, but if we do so, we should be consistent about it. We should also do it for `phonon_dos_flow`. And, furthermore, do it in the other phonon recipes where this is done, such as the [toy EMT example](https://github.com/Quantum-Accelerators/quacc/blob/36dde0fb1d70a7083547b29b31ef83c6a9b3300d/src/quacc/recipes/emt/phonons.py#L32).

Andrew-S-Rosen: Very useful to hear! Thank you again. I am going to play around with some things in https://github.com/Quantum-Accelerators/quacc/pull/2372. I will ping you both when I'm done. @nerdwarrior: No need to further modify this PR, but I would certainly appreciate a review when I'm done toying around with things. I will seek to implement the suggestions here.

Andrew-S-Rosen:
Closing in favor of #2372.

PR #2372, opened 2024-07-29
Remove preliminary `relax_job` calls in phonon flows (joint with @nerdwarrior)

Andrew-S-Rosen:
This PR follows up conversations in #2367 wherein we agreed to remove the `relax_job` calls at the start of the phonon flows.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 ## Summary of Changes
 
-\>> Provide context and a description of your changes. Make sure to reference any associated issues. <<
+\>> Provide context and a description of your changes here. Make sure to reference any associated issues. <<
 
 ### Requirements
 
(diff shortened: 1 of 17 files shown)


Andrew-S-Rosen:
@nerdwarrior: I have implemented @tomdemeyere's suggestion (based on your helpful feedback) of not having a preliminary relaxation job in the `grid_phonon_flow`. For internal consistency, I have done this for all such phonon flows in quacc.

Would you be willing to take a look and see if `grid_phonon_flow` now looks more intuitive to you? You don't need to review the other recipes here.

This PR does _not_ do any checks on whether the user has specified `copy_files` or `prev_outdir`. I will leave that for a separate PR to keep this PR focused on a single task, should someone be interested in implementing it.

tomdemeyere:
Thanks for exploring the idea! Originally, I wasn't assuming that you would apply this to all the concerned flows! But to be completely honest, I totally understand the need for consistency.

I think these changes will be beneficial for people who have more of a "Lego" mind. This will allow them to build their workflows using a combination of elementary functions, instead of using functions that appear slightly more complex, potentially obscuring a well-thought construction.

For people having more of a "black-box" mindset, well... this isn't such a big change, after all QuAcc is all about workflows, and I don't think this will cause many issues.

Also, thinking more technically:

1 - This might help people manage job decorations when performing complex workflows. Not sure.
2 - If I remember, at some point we stumbled upon a case where a flow would rather have a job resolved mid-flow, this kind of thinking might avoid these issues?

Andrew-S-Rosen:
Great to hear! I also think this is ultimately the right way to go for the reasons you outlined. In the spirit of "keep everything as simple as possible but not simpler", this PR seems to get closer to that goal. I like your elementary function comment... it is the elementary reaction of programming. 😄 

> 1 - This might help people manage job decorations when performing complex workflows. Not sure.

There should be the same amount of flexibility, but it will admittedly be easier to achieve customization since the user doesn't need to play around with `job_params` for the relaxation step.

> 2 - If I remember, at some point we stumbled upon a case where a flow would rather have a job resolved mid-flow, this kind of thinking might avoid these issues?

Hmmm... potentially. Honestly, I can't quite remember. But simplicity is a good thing in my mind!

Thanks again for the helpful discussion. I greatly appreciate it! I'll let this PR sit for a bit until @nerdwarrior takes a look, but I think it should address the various points of concern!

nerdwarrior:
@Andrew-S-Rosen Thank you for going through this, and @tomdemeyere thank you for the helpful suggestions! I'm on the same page with all that's been discussed, and I also agree about liking the simplicity. I appreciate everyone waiting!

Andrew-S-Rosen:
Wonderful! I'll merge, and we can improve further if needed. Thanks again for the valuable comments! 

No worry about the (very short) wait :) we are in no rush over here.

PR #2391, opened 2024-08-06
Add tests for `find_recent_logfile()`

### LINKED ISSUE #2389: Add a test for `find_recent_logfile`
### What would you like to report?

The following function has become untested. We should test it! It is not too hard :)

https://github.com/Quantum-Accelerators/quacc/blob/8ca88c22dff6196690e44942169340c43383dbc7/src/quacc/utils/files.py#L238-L265

mikeweltevrede:
## Summary of Changes
Add tests for `find_recent_logfile()` to close #2389. Note that I also noticed that there might be some unintended functionality (see `test_find_recent_logfile_only_checks_files_matching_the_extension`) and added a test that is expected to fail.

### Requirements


buildbot-princeton:

tests/core/utils/test_files.py
```diff
@@ -176,3 +182,44 @@ def test_check_logfile(tmp_path):
...
+def test_find_recent_logfile_only_checks_files_matching_the_extension(tmp_path):
+    """
+    This test double checks that find_recent_logfile only matches file extensions and does not hit files if the desired
+    extension is somewhere else in the file name or path.
+    """
+    with open(tmp_path / "first_log.out", "w"):
+        ...
+
+    actual = find_recent_logfile(tmp_path, logfile_extensions=["log"])
+    assert actual is None
```
Andrew-S-Rosen, line 225: Ah, good catch! Rather than reinforce a bug, let's try to address it properly.

https://github.com/Quantum-Accelerators/quacc/blob/69e4fc075f3775ddf3abc81705b05a6f48251ce8/src/quacc/utils/files.py#L262

The above is the source of the problem. We are checking for the `ext: str` in `str(f)`, and as you noted, that will also catch a match anywhere within `str(f)` --- not just the extension.

I believe the proper fix for this would instead be:

```python
 if ext in f.suffixes and f_path.stat().st_mtime > mod_time: 
```

Andrew-S-Rosen (commented):
@mikeweltevrede: Thank you very much for taking the time to write tests for this function, which in fact discovered a bug! That's why we write thorough tests, and I appreciate your attention to detail with this.

I would be happy to merge this PR once my one comment is addressed. Don't hesitate to let me know if you have any questions.

tests/core/utils/test_files.py
mikeweltevrede: @Andrew-S-Rosen That makes sense. I decided not to fix it in the first case because some libraries are very strict about separating different functionalities in PRs even to this extent. Will fix and push :)

mikeweltevrede: I fixed the issue. I tried a little bit with `glob` and `rglob` to simplify this (you would be able to move the `if ext in str(f)` to that line and remove some levels of nesting) but I was unable to make it work quickly when `logfile_extensions` is a list of multiple extensions (note that you would also need to handle whether users would provide the extension as `.log` or `log`). Maybe a nice item to work on by others.

It would be something like this, assuming the user provides the period in the `logfile_extensions`:

```python
    mod_time = 0.0
    logfile = None
    if isinstance(logfile_extensions, str):
        pattern = logfile_extensions
        logfile_extensions = [logfile_extensions]
    else:
        pattern = f"[{']['.join(logfile_extensions)}]"

    for f_path in Path(directory).expanduser().rglob(pattern):
        if f_path.stat().st_mtime > mod_time:
            mod_time = f_path.stat().st_mtime
            logfile = f_path.resolve()
    return logfile
```

Andrew-S-Rosen:
Fantastic, thank you @mikeweltevrede! I appreciate your help with this. This is wonderful. I am happy to merge.

PR #2460, opened 2024-09-11
resolve all quacc/prefect flow futures to data make prefect happy 

zulissimeta:
## Summary of Changes

Prefect3 now requires all jobs to be resolved before a flow exits. If you return a job that is still running, this causes warnings and problems.

Also, return futures/states is good in that you can avoid passing around data until it's needed, but annoying in that you may ended up with complex structures that have prefect futures embedded inside them. 

This PR introduces a function `resolve_futures_to_results` modeled on the prefect function `resolve_futures_to_states` that will recursively crawl python collections and resolve any futures to their final results. This function is called on any results that are returned by a quacc flow so that they are properly resolved before the flow finishes.

### Requirements

### Example code

```
import time

from quacc import flow, job


@job
def slow_math():
    # An independent piece of work that might take a while
    time.sleep(2)
    return 5


@flow()
def math_flow():
    return slow_math()


@flow()
def double_flow():
    a = math_flow()
    b = math_flow()
    return a + b


math_flow()
double_flow()
```

Before these patches, prefect would throw warnings about how tasks were not being resolved within flows. After, each flow finishes cleanly.

### DIFF
--- a/docs/user/basics/wflow_syntax.md
+++ b/docs/user/basics/wflow_syntax.md
@@ -193,8 +193,7 @@ graph LR
         return output2
 
 
-    future = workflow(1, 2, 3)  #  (3)!
-    result = future.result()  #  (4)!
+    result = workflow(1, 2, 3)  #  (3)!
     print(result)  # 9
     ```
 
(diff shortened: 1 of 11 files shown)


buildbot-princeton:

Andrew-S-Rosen:
Note to self: We have to update the documentation (examples) to reflect this change in behavior.

zulissimeta:
> Note to self: We have to update the documentation (examples) to reflect this change in behavior.

docs updated!

Andrew-S-Rosen:
Thanks, @zulissimeta! I assume this doesn't break concurrency of things, so I'm happy to merge this. Overall, it seems much more intuitive for end users anyway.

Would be good to get something like this in Prefect so that upstream code changes don't break our utility function. But for now, let's not let perfect be the enemy of good enough.

Issue #2457, opened 2024-09-07
mp_gga_relax_job pseudopotential parsing bug


zulissimeta:
### Details about the quacc environment

- **quacc version**: 0.11.3
- **Python version**: 3.11

### What is the issue?

The pseudopotential symbols are not getting split correctly in the VASP set converter. 

```
  File "../lib/python3.11/site-packages/quacc/calculators/vasp/params.py", line 471, in _convert
    potcar_setups[k] = v.split(k)[-1]
                       ^^^^^^^^^^
ValueError: empty separator
```

This appears to be because the list of pseudopotential symbols (which is separated by `\n` is not getting split into a list, so each character is treated separately. 

### How can we easily reproduce the issue?

```python
from ase.build import bulk
from quacc import flow
from quacc.recipes.vasp.mp import mp_gga_relax_job


@flow
def test():
    return mp_gga_relax_job(bulk("Cu"))


test()
```

Andrew-S-Rosen:
Thank you for the report, @zulissimeta! This looks like a new bug due to the release of atomate2 0.0.15. We aren't testing it on CI yet because of a dependency conflict (https://github.com/materialsproject/atomate2/pull/970). Happy to accept a fix.

Andrew-S-Rosen:
Fixed by @chiang-yuan in #2463!

PR #2463, opened 2024-09-17
add line split for potcar_symbols

chiang-yuan:
@Andrew-S-Rosen  I found the problem. It is not `emmet-core` or `pymargen` version issue. `input_set.potcar` needs line splitting similar to what is done at [line 424](https://github.com/chiang-yuan/quacc/blob/ea34aa63b76d7cdd3c1eed0d674827dd04a16337/src/quacc/calculators/vasp/params.py#L424)

I tested the vasp workflow locally and didn't run into any errors

This will close issue #2457 and close https://github.com/materialsproject/atomate2/pull/984

### DIFF
--- a/src/quacc/calculators/vasp/params.py
+++ b/src/quacc/calculators/vasp/params.py
@@ -449,7 +449,7 @@ class MPtoASEConverter:
         )
         self.incar_dict = input_set.incar
         self.pmg_kpts = input_set.kpoints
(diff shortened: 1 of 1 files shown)


buildbot-princeton:

chiang-yuan:
okay it fails the previous test. looks like we need to do version specific test here. 

I was using the following versions

```
emmet-core==0.84.2rc7
atomate2==0.0.15
pymatgen==2024.9.10
```

PR #2511, opened 2024-10-23
Prefect utils fix resultrecord

zulissimeta:
## Summary of Changes

The resolve_futures_to_results seems to have a bug when the result is a "ResultRecord" rather than a "BaseResult". This came up when running a simple phonon flow:

```python
from quacc.recipes.emt.phonons import phonon_flow
from ase.build import bulk

cu_bulk = bulk('Cu')
phonon_flow(cu_bulk)
```

This change handles the case, and adds the phonon flow above as a test case for prefect (nice since it has a bunch of subflows/jobs/etc.

### Requirements


buildbot-princeton:

Andrew-S-Rosen:
@zulissimeta: Thanks! Good catch. This is the function adapted from the upstream Prefect, right? If so, should we be opening an issue there to mark it?

zulissimeta:
> @zulissimeta: Thanks! Good catch. This is the function adapted from the upstream Prefect, right? If so, should we be opening an issue there to mark it?

Yup - not sure this is common though - it only happens because we're chaining subflows to job which is perhaps less common in prefect?

Andrew-S-Rosen:
> Yup - not sure this is common though - it only happens because we're chaining subflows to job which is perhaps less common in prefect?

I think that would still be valid prefect though, so I'd classify it as a bug. Ultimately, I would like to excise this custom patching from quacc entirely, but until I have someone to do that, we shall merge...

Issue #2429, opened 2024-08-17
Add support for more foundation models


Andrew-S-Rosen:
### What new feature would you like to see?

We should add the pre-trained [SevenNet MLP](https://github.com/MDIL-SNU/SevenNet/tree/main) to the supported MLP methods since it performs so well on https://matbench-discovery.materialsproject.org.

This would entail adding it as a supported calculator method in `quacc.recipes.mlp._base.pick_calculator`, updating the `method` type hints in the `quacc.recipes.mlp` functions to include SevenNet, adding some tests to the test suite, and adding `sevenn` to `pyproject.toml`.

Also see https://github.com/orbital-materials/orb-models.

PR #2553, opened 2024-11-30
feat: add sevennet support to mlp recipes

mamunm:
## Summary of Changes
Added **sevennet** to the MLP recipes based on issue #2429. 

Key changes include:

1. Added sevennet implementation to mlp recipes
2. Created unit tests to validate sevennet functionality
3. Ensured compatibility with existing MLP workflows

## Requirements

-[X] My PR is focused on a single feature addition [#2429 ].
-[X] My PR has relevant, comprehensive unit tests.
-[X] My PR is on a custom branch (feature/sevennetMLP).

### DIFF
--- a/src/quacc/recipes/mlp/_base.py
+++ b/src/quacc/recipes/mlp/_base.py
@@ -64,11 +64,11 @@ def pick_calculator(
         if "default_dtype" not in kwargs:
             kwargs["default_dtype"] = "float64"
         calc = mace_mp(**kwargs)
-    
+
     elif method.lower() == "sevennet":
         from sevenn import __version__
         from sevenn.sevennet_calculator import SevenNetCalculator
-        
+
         if "model" not in kwargs:
             kwargs["model"] = "7net-0"
         if "device" not in kwargs:
--- a/src/quacc/recipes/mlp/_base.py
+++ b/src/quacc/recipes/mlp/_base.py
@@ -16,7 +16,7 @@ LOGGER = getLogger(__name__)
 
 @lru_cache
 def pick_calculator(
-    method: Literal["mace-mp-0", "m3gnet", "chgnet"], **kwargs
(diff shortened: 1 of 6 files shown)


buildbot-princeton:

src/quacc/recipes/mlp/_base.py
```diff
@@ -64,6 +65,16 @@ def pick_calculator(
             kwargs["default_dtype"] = "float64"
         calc = mace_mp(**kwargs)
 
+    elif method.lower() == "sevennet":
+        from sevenn import __version__
+        from sevenn.sevennet_calculator import SevenNetCalculator
+
+        if "model" not in kwargs:
+            kwargs["model"] = "7net-0"
```
Andrew-S-Rosen, line 73: 7net-0 is the default model, so it seems like we should probably not specify this.

src/quacc/recipes/mlp/_base.py
```diff
@@ -64,6 +65,16 @@ def pick_calculator(
...
+        if "device" not in kwargs:
+            kwargs["device"] = "cpu"
```
Andrew-S-Rosen, line 75: The default device is "auto". Is there a reason to override this default by forcing CPU?

Andrew-S-Rosen (commented):
@mamunm: Thank you very much for your PR! I have just two questions before I can merge this.

src/quacc/recipes/mlp/_base.py
mamunm: I added both the default settings so the user knows what model and devices are being used if they want to dig deeper. But it's redundant, and I'll remove them in the next push.

Andrew-S-Rosen: The default setting is "auto" for device I believe, but in any case let's remove them so we are always consistent with the underlying calculator.

Andrew-S-Rosen:
Thanks, merging!

mamunm:
> Thanks, merging!

Thanks for the review!

PR #2384, opened 2024-08-04
Bump ase from 3.23.0 to 3.24.0

Andrew-S-Rosen:

PR #2574, opened 2024-12-19
feat: add orb-models support to mlp recipes

mamunm:
## Summary of Changes
Added **orb-models** to the MLP recipes based on issue #2429. 

Key changes include:

1. Added orb-models implementation to mlp recipes
2. Created unit tests to validate orb-models functionality
3. Ensured compatibility with existing MLP workflows

## Requirements

-[X] My PR is focused on a single feature addition [#2429 ].
-[X] My PR has relevant, comprehensive unit tests.
-[X] My PR is on a custom branch (feature/orbnet-models).

## Note to the reviewer: 
1. orb-models outputs are somewhat more stochastic than other tests, so I used a `rel` flag to limit the accuracy of the pytest comparison. I set it to `1e-4` after some failed checks.


pyproject.toml
```diff
@@ -45,7 +45,7 @@ covalent = ["covalent>=0.234.1-rc.0; platform_system!='Windows'", "covalent-clou
 dask = ["dask[distributed]>=2023.12.1", "dask-jobqueue>=0.8.2"]
 defects = ["pymatgen-analysis-defects>=2024.10.22", "shakenbreak>=3.2.0"]
 jobflow = ["jobflow[fireworks]>=0.1.14", "jobflow-remote>=0.1.0"]
-mlp = ["matgl>=1.1.2", "chgnet>=0.3.3", "mace-torch>=0.3.3", "torch-dftd>=0.4.0", "sevenn>=0.10.1"]
+mlp = ["matgl>=1.1.2", "chgnet>=0.3.3", "mace-torch>=0.3.3", "torch-dftd>=0.4.0", "sevenn>=0.10.1", "orb-models>=4.1.0", "pynanoflann @ git+https://github.com/dwastberg/pynanoflann#egg=af434039ae14bedcbb838a7808924d6689274168"]
```
Andrew-S-Rosen, line 48: We can't put direct git links in `pyproject.toml` because this prevents us from uploading the code to PyPI. Instead, we should check to see if this import exists via `mportlib.util.find_spec` and if it doesn't, we raise a clear error to the user informing them to install the code and how. It should also be mentioned in the recipe docstring.

src/quacc/recipes/mlp/_base.py
```diff
@@ -16,7 +16,7 @@
 
 @lru_cache
 def pick_calculator(
-    method: Literal["mace-mp-0", "m3gnet", "chgnet", "sevennet"], **kwargs
+    method: Literal["mace-mp-0", "m3gnet", "chgnet", "sevennet", "orb-models"], **kwargs
```
Andrew-S-Rosen, line 19: Would it be clearer to just call it "orb"?

src/quacc/recipes/mlp/_base.py
```diff
@@ -71,6 +71,15 @@ def pick_calculator(
 
         calc = SevenNetCalculator(**kwargs)
 
+    elif method.lower() == "orb-models":
+        from orb_models import __version__
+        from orb_models.forcefield import pretrained
+        from orb_models.forcefield.calculator import ORBCalculator
+
+        orb_model = kwargs.pop("model", "orb_v2")
```
Andrew-S-Rosen, line 79: Why are we popping the model keyword argument out? We shouldn't modify keyword arguments specified by the user. It would be better to do

```python
if "model" not in kwargs:
    kwargs["model"] = "orb_v2"
```

tests/requirements-mlp.txt
```diff
@@ -2,3 +2,5 @@ chgnet==0.4.0
 mace-torch==0.3.9
 torch-dftd==0.5.1
 sevenn==0.10.3
+orb-models==4.1.0
+pynanoflann @ git+https://github.com/dwastberg/pynanoflann#egg=af434039ae14bedcbb838a7808924d6689274168
```
Andrew-S-Rosen, line 6: We can keep the git link in `requirements-mlp.txt`. But can we remove the #egg=af434039ae14bedcbb838a7808924d6689274168?

Andrew-S-Rosen (commented):
@mamunm Thanks for your contribution! Some comments below.

src/quacc/recipes/mlp/_base.py
mamunm: We need to obtain the pre-trained model as an instance of `GraphRegressor` and pass it to `ORBCalculator`, instead of providing the model name as a string like we do in other calculators. Since the `model` won't be used elsewhere, I  popped it from the `kwargs`, but I realize that this may not be the best approach. In the recent update, I switched to using `kwargs.get()`.

mamunm:
Tests in MLPS are failing because `uv` cannot install `pynanoflann`, whereas `pip` can. I am trying to figure out how to resolve this.

Andrew-S-Rosen:
Hmm, it should be able to I think? I do so here https://github.com/Quantum-Accelerators/quacc/pull/2384/files

Andrew-S-Rosen:
Thank you for the changes, @mamunm! This looks good to me now! Let me know when you are happy to merge.

src/quacc/recipes/mlp/_base.py
Andrew-S-Rosen: Ah, yes I like this better.

mamunm:
@Andrew-S-Rosen Thanks for the review! I am happy to merge it.

PR #2587, opened 2024-12-30
Fix sorting with MP sets

### LINKED ISSUE #2583: MP GGA VASP workflows resort and cause problems with per-atom defaults like LDA+U

- **quacc version**: latest
- **Python version**: 3.11

### What is the issue?

The ASE Vasp calculator resorts the atoms when writing POSCARs, which is a problem when any specific script (like the scripts to guess MP-like LDA+U or magmom settings). 

This is a problem because using the mp_gga_relax_job will silently cause problems if the input atoms object does not have the same order as what the Vasp calculator will plan to do. See the script below for the minimum reproducible example. Running the MP GGA relax jobs on atoms objects with random sorting and LDAU settings will also have problems, but the calculator won't actually throw errors as VASP will just ignore more LDA+U parameters than the number of species. 

```python
from quacc.calculators.vasp import Vasp
from ase import Atoms
atoms = Atoms('HOH', cell=[10,10,10])
atoms.set_calculator(Vasp(atoms, ldau=True, ldauu=[1,2,3]))
atoms.calc.write_input(atoms)
```

Notice that in this example, INCAR has LDAUU=[1,2,3], while the POSCAR has been resorted to not be HOH but instead be H2O, so that there are only two species.

Andrew-S-Rosen:
Closes #2583, I think.

### DIFF
--- a/src/quacc/calculators/vasp/params.py
+++ b/src/quacc/calculators/vasp/params.py
@@ -398,8 +398,10 @@ class MPtoASEConverter:
         if atoms is None and prev_dir is None:
             raise ValueError("Either atoms or prev_dir must be provided.")
         self.atoms = atoms
-        self.structure = AseAtomsAdaptor.get_structure(atoms) if self.atoms else None
         self.prev_dir = prev_dir
+        if self.atoms:
+            self.ase_sort, self.ase_resort = Vasp_()._make_sort(self.atoms)
+            self.structure = AseAtomsAdaptor.get_structure(self.atoms[self.ase_sort])
 
     def convert_dict_set(self, dict_set: DictSet) -> dict:
         """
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
Pinging @zulissimeta: I _believe_ this PR closes #2583. It certainly fixes a bug -- hopefully it's the same one you're seeing.

Previously, if we had `Atoms("HOHMn")`, the `LDAUU` flag would _incorrectly_ be made as `LDAUU = 0 0 0 3.9`. The reason this is incorrect is that the POSCAR generated from ASE will have only _three_ element types at the top: `H O Mn`. Therefore, there will be a mismatch between the LDAUU entries and the corresponding elements any time that there is merging of elements into a single element type.

With this patch, the `LDAUU` flag would _correctly_ be made as `LDAUU = 0 0 0 3.9` since we generate the MP input set on the ASE-sorted structure so that it is ultimately written out in the correct way. This PR does not make any changes to a user's `Atoms` object --- I try not to mutate anything that might be user-facing.

If this does not solve your particular issue, please provide the sample `Atoms` object or `POSCAR` along with a minimal reproducible example of the exact issue so that I can troubleshoot it appropriately.

PR #2676, opened 2025-02-24
Add filter kwargs

zulissimeta:
## Summary of Changes

The ASE-based run_opt used uses a FrechetCellFilter. That class can take kwargs, for example applying a pressure during the relaxation. This PR simply adds a filter_kwargs to run_opt so these can flow down to the FrechetCellFilter. 

### Requirements


buildbot-princeton:

tests/core/recipes/emt_recipes/test_emt_recipes.py
```diff
@@ -93,6 +93,24 @@ def test_relax_job(tmp_path, monkeypatch):
...
+        relax_cell=True,
+        opt_params={
+            "fmax": 0.03,
+            "optimizer": FIRE,
+            "filter_kwargs": {"scalar_pressure": 0.001},
+        },
+        asap_cutoff=True,
+    )
+    assert output["nsites"] == len(atoms)
+    assert output["parameters"]["asap_cutoff"] is True
```
Andrew-S-Rosen, line 112: Is there anything we can `assert` here to make sure that the filter kwargs were used or is that not super feasible? Happy to merge either way.

Andrew-S-Rosen:
@buildbot-princeton add to whitelist

tests/core/recipes/emt_recipes/test_emt_recipes.py
zulissimeta: Done!

Andrew-S-Rosen:
@zulissimeta Thanks for the change! Can you briefly explain to me what the following is testing? If the `filter_kwargs` were not passed through, wouldn't the two expressions be equal and therefore the assertion would still hold `True`? I think I am misunderstanding the intention with the test.

```python
    assert (
        output_fire_pressure["results"]["energy"] - output_fire["results"]["energy"]
        < 0.1
    )
```

zulissimeta:
> @zulissimeta Thanks for the change! Can you briefly explain to me what the following is testing? If the `filter_kwargs` were not passed through, wouldn't the two expressions be equal and therefore the assertion would still hold `True`? I think I am misunderstanding the intention with the test. Are you trying to do a `pytest.approx()` here or does the sign convention matter, or...?
> 
> ```python
>     assert (
>         output_fire_pressure["results"]["energy"] - output_fire["results"]["energy"]
>         < 0.1
>     )
> ```

Good catch! Edited to use volume (which should definitely be smaller when you apply pressure!).

PR #2679, opened 2025-02-25
Fairchem calculator

zulissimeta:
## Summary of Changes

This PR add support for FAIR Chemistry ASE calculators. https://fair-chem.github.io/

### Requirements

- [x] My PR is focused on a [single feature addition or bugfix](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/best-practices-for-pull-requests#write-small-prs).
- [x] My PR has relevant, comprehensive [unit tests](https://quantum-accelerators.github.io/quacc/dev/contributing.html#unit-tests).
- [x] My PR is on a [custom branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-and-deleting-branches-within-your-repository) (i.e. is _not_ named `main`).

Note: If you are an external contributor, you will see a comment from [@buildbot-princeton](https://github.com/buildbot-princeton). This is solely for the maintainers.

PR #2680, opened 2025-03-02
Add support for Fairchem calculator

Andrew-S-Rosen:
Closes #2679.


Andrew-S-Rosen:
@zulissimeta: Everything largely looks good. Thanks so much for handling the CI refactoring. 

If you take a look at the error log, you'll see the following. I just want to confirm that it is okay to update the test numbers before proceeding with the merge.

```
FAILED tests/core/recipes/mlp_recipes/test_core_recipes.py::test_static_job[fairchem] - assert -4.098316669464111 == -4.09 ± 4.1e-04
  
  comparison failed
  Obtained: -4.098316669464111
  Expected: -4.09 ± 4.1e-04
FAILED tests/core/recipes/mlp_recipes/test_core_recipes.py::test_relax_job[fairchem] - assert -32.80327224731445 == -32.7361946105957 ± 3.3e-03
  
  comparison failed
  Obtained: -32.80327224731445
  Expected: -32.7361946105957 ± 3.3e-03
FAILED tests/core/recipes/mlp_recipes/test_core_recipes.py::test_relax_cell_job[fairchem] - assert -32.82823944091797 == -32.73428726196289 ± 3.3e-03
  
  comparison failed
  Obtained: -32.82823944091797
  Expected: -32.73428726196289 ± 3.3e-03
===== 3 failed, 31 passed, 7 skipped, 342 deselected in 171.77s (0:02:51) ======
```

tests/requirements-mlp.txt
```diff
@@ -1,6 +1,12 @@
...
 sevenn==0.10.4
 orb-models==0.4.2
+fairchem-core==1.5.0
 pynanoflann@git+https://github.com/dwastberg/pynanoflann
+# Requirements from fairchem-core
+torch_geometric==2.6.1
+-f https://data.pyg.org/whl/torch-2.4.1+cpu.html
+torch_scatter==2.1.2
+torch_sparse==0.6.18
+torch_cluster==1.6.3
```
Andrew-S-Rosen, line 12: This is not a blocker, but if these are true requirements of `fairchem-core`, they should ideally be added to `fairchem-core` as an optional set of "extras" that can be installed via something like `pip install fairchem-core[pyg]` that way if a release like `torch_geometry==3.0.0` one day breaks `fairchem-core`, a version `<=` pin can be specified in the `pyproject.toml` of `fairchem-core` appropriately. You can still instruct people to install the PyG-recommended packages, but having the package names in `pyproject.toml` would ensure that the proper versions are always installed.

One of the annoying things about having them all in quacc's `requirements.txt` is that now dependabot will open PRs for four additional packages corresponding to each dependency release even though the actual maintenance and testing should be done on the `fairchem-core` side since it is only impacting that package.

zulissimeta: Let's try without these requirements and see what happens. Torch-geometric is pretty finicky about versions and install which is why we have these explicitly installed that way. We aren't actually pinned that hard to those versions, it's just important that torch-geometric and torch are consistent with each other.

zulissimeta:
Yes, please update with the real numbers! I just put in some place holders (I think copying the orb numbers) and was hoping that it was going to be easy once the tests ran. 

> @zulissimeta: Everything largely looks good. Thanks so much for handling the CI refactoring.
> 
> If you take a look at the error log, you'll see the following. I just want to confirm that it is okay to update the test numbers before proceeding with the merge.
> 
> ```
> FAILED tests/core/recipes/mlp_recipes/test_core_recipes.py::test_static_job[fairchem] - assert -4.098316669464111 == -4.09 ± 4.1e-04
>   
>   comparison failed
>   Obtained: -4.098316669464111
>   Expected: -4.09 ± 4.1e-04
> FAILED tests/core/recipes/mlp_recipes/test_core_recipes.py::test_relax_job[fairchem] - assert -32.80327224731445 == -32.7361946105957 ± 3.3e-03
>   
>   comparison failed
>   Obtained: -32.80327224731445
>   Expected: -32.7361946105957 ± 3.3e-03
> FAILED tests/core/recipes/mlp_recipes/test_core_recipes.py::test_relax_cell_job[fairchem] - assert -32.82823944091797 == -32.73428726196289 ± 3.3e-03
>   
>   comparison failed
>   Obtained: -32.82823944091797
>   Expected: -32.73428726196289 ± 3.3e-03
> ===== 3 failed, 31 passed, 7 skipped, 342 deselected in 171.77s (0:02:51) ======
> ```

.github/workflows/tests.yaml
```diff
@@ -353,6 +359,45 @@ jobs:
...
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+          cache-dependency-path: tests/requirements**.txt
```
janosh, line 380: you could consider using [`astral-sh/setup-uv`](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#faq) in place of `pip install uv` below

pyproject.toml
```diff
@@ -45,7 +45,8 @@ covalent = ["covalent>=0.234.1-rc.0; platform_system!='Windows'", "covalent-clou
 dask = ["dask[distributed]>=2023.12.1", "dask-jobqueue>=0.8.2"]
 defects = ["pymatgen-analysis-defects>=2024.10.22", "shakenbreak>=3.2.0"]
 jobflow = ["jobflow>=0.1.14", "jobflow-remote[gui]>=0.1.0"]
-mlp = ["matgl>=1.1.2", "chgnet>=0.3.3", "mace-torch>=0.3.3", "torch-dftd>=0.4.0", "sevenn>=0.10.1", "orb-models>=0.4.1"]
+mlp1 = ["chgnet>=0.3.3", "torch-dftd>=0.4.0", "sevenn>=0.10.1", "orb-models>=0.4.1", "fairchem-core>=1.4.0"]
+mlp2 = ["mace-torch>=0.3.3", "matgl>=1.1.2"]
```
janosh, line 49: are these dep sets split based on required `e3nn` version? maybe worth renaming to something more descriptive?

janosh (commented):
nice to see this! 👍

.github/workflows/tests.yaml
Andrew-S-Rosen: Ah, good to know! Thanks!

pyproject.toml
Andrew-S-Rosen: It's a mess 😅 matgl and chgnet have torch compatibility issues, and then there are also e3nn version issues too. I definitely should rename them though!

.github/workflows/tests.yaml
Andrew-S-Rosen: > Using actions/setup-python can be faster, because GitHub caches the Python versions alongside the runner.

That part seems not ideal. Unclear to me what the difference is between that and `enable-cache: true` with `setup-uv` though.

janosh: `enable-cache: true` refers to the package cache whereas that comment is talking about caching Python itself. you can use both actions together which is what I usually do. haven't tried `setup-uv` on its own though, could be that Python caching makes little difference

PR #2693, opened 2025-03-07
Extend elastic properties to MLPs, and include elastic property analysis

zulissimeta:
## Summary of Changes

This PR extends elasticity support and addresses a few quality-of-life issues for running elastic workflows:
* Add elastic property support to the MLP recipes
* Add an ElasticSchema to include context on the deformations (needed for final elastic property fitting)
* Add a pre-relax option to the elastic property workflow
* Calculate the ElasticityDoc alongside the others results

### Requirements


src/quacc/types.py
```diff
@@ -528,6 +530,12 @@ class ThermoSchema(AtomsSchema):
         parameters_thermo: ParametersThermo
         results: ThermoResults
 
+    class ElasticSchema(TypedDict):
+        deformed_structure_set: DeformedStructureSet
+        deformed_results: list[RunSchema | OptSchema]
+        undeformed_result: RunSchema | OptSchema
+        elasticity_doc: ElasticityDoc
```
Andrew-S-Rosen, line 537: Please add a docstring for the class so it renders appropriately in the built documentation.

tests/core/recipes/mlp_recipes/test_elastic_recipes.py
Andrew-S-Rosen, line 1: We will merge https://github.com/Quantum-Accelerators/quacc/pull/2695 first since I anticipate that will require only one small change. Please make sure that this test reflects changes to https://github.com/Quantum-Accelerators/quacc/pull/2695 after the merge.

Andrew-S-Rosen (commented):
Thanks, @zulissimeta! This largely looks sensible to me. I have a few suggestions and requests in my review comments, mostly about user-facing choices and documentation.

src/quacc/recipes/common/elastic.py
```diff
@@ -4,24 +4,66 @@
...
 
     from quacc import Job
+    from quacc.types import ElasticSchema, OptSchema, RunSchema
+
+
+@job
+def deformations_to_elastic_tensor(
+    undeformed_result: OptSchema | RunSchema,
+    deformed_structure_set: DeformedStructureSet,
+    results: list[dict],
```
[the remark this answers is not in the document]
zulissimeta: Done!

src/quacc/recipes/emt/elastic.py
```diff
@@ -66,11 +67,17 @@ def bulk_to_deformations_flow(
         [relax_job, static_job],
         param_swaps=job_params,
         decorators=job_decorators,
-    )
+    )  # type: ignore
+
+    if pre_relax:
+        undeformed_result = relax_job_(atoms, relax_cell=True)
+    else:
+        undeformed_result = static_job_(atoms)
```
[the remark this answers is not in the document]
zulissimeta: Think this should be fixed now.

src/quacc/recipes/mlp/elastic.py
```diff
@@ -0,0 +1,85 @@
...
+    from ase.atoms import Atoms
+
+    from quacc.types import ElasticSchema
+
+
+@flow
+def bulk_to_deformations_flow(
+    atoms: Atoms,
+    run_static: bool = False,
+    pre_relax: bool = True,
```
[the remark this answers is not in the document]
zulissimeta: Done!

src/quacc/recipes/mlp/elastic.py
```diff
@@ -0,0 +1,85 @@
...
+        ["relax_job", "static_job"],
+        [relax_job, static_job],
+        param_swaps=job_params,
+        decorators=job_decorators,
+    )  # type: ignore
+
+    if pre_relax:
+        undeformed_result = relax_job_(atoms, relax_cell=True)
+    else:
+        undeformed_result = static_job_(atoms)
```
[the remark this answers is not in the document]
zulissimeta: Same as above, done!

src/quacc/types.py
zulissimeta: Think I got it, but most of the types don't have docstrings so not 100% sure what you were looking for.

Andrew-S-Rosen: What do you know, you're right. I suppose it's not really needed then!

src/quacc/recipes/emt/elastic.py
```diff
@@ -43,6 +44,8 @@ def bulk_to_deformations_flow(
     ----------
     atoms
         Atoms object
+    pre_relax:
```
Andrew-S-Rosen, line 47: Could you remove the `:` here so it renders appropriately in the built docs? Thanks!

src/quacc/recipes/emt/elastic.py
```diff
@@ -66,11 +69,21 @@ def bulk_to_deformations_flow(
...
         decorators=job_decorators,
-    )
+    )  # type: ignore
+
+    if pre_relax:
+        if run_static:
+            undeformed_relax = relax_job_(atoms, relax_cell=True)
+            undeformed_result = static_job_(undeformed_relax["atoms"])
+        else:
+            undeformed_result = relax_job_(atoms, relax_cell=True)
```
Andrew-S-Rosen, line 79: We should run `undeformed_result = relax_job_(atoms, relax_cell=True)` first regardless. And then `if run_static` we can run the static. We have some code duplication here otherwise. See below.

src/quacc/recipes/emt/elastic.py
```diff
@@ -66,11 +69,21 @@ def bulk_to_deformations_flow(
...
+    else:
+        undeformed_result = static_job_(atoms)
```
Andrew-S-Rosen, line 81: This code block has some problems. For instance, it will run a static job even if `run_static=False` if `pre_relax=False` too. Also, we have some code duplication here with the `static_job_` call.

This code block could be greatly cleaned up by doing something like the following (note: please sanity check): 

```python
if pre_relax:
    undeformed_result = relax_job_(atoms, relax_cell=True)
    atoms = underformed_result["atoms"]
if run_static:
    underformed_result = static_job_(atoms)
```

src/quacc/recipes/mlp/elastic.py
```diff
@@ -0,0 +1,89 @@
...
+    )  # type: ignore
+
+    if pre_relax:
+        if run_static:
+            undeformed_relax = relax_job_(atoms, relax_cell=True)
+            undeformed_result = static_job_(undeformed_relax["atoms"])
+        else:
+            undeformed_result = relax_job_(atoms, relax_cell=True)
+    else:
+        undeformed_result = static_job_(atoms)
```
Andrew-S-Rosen, line 81: Same deal here about the code block. Hopefully we won't have to duplicate these things between recipes once I allow calculators to be passed, but for now you'll have to make the code change twice.

Andrew-S-Rosen (commented):
@zulissimeta: Thanks! I had a few minutes to give this another look over. We're nearly there.

First, could you see my two unresolved comments below:
- https://github.com/Quantum-Accelerators/quacc/pull/2693/files#r1987581514
- https://github.com/Quantum-Accelerators/quacc/pull/2693/files#r1987582170

For the first one, we are having `bulk_to_deformations_subflow` always take a `static_job` even though running a static calculation is purely optional. This is not ideal. For instance, let's say a set of recipes does not even have a static job made --- now they will have to make one just to have it be ignored by this subflow. We want to avoid such a scenario, which is why it was originally set to `Job | None = None`. The original method seems more suitable in my opinion since one cannot run a static calculation if no static job is provided. But I am open to suggestions.

The other comment is with regards to some `if/else` code blocks that can be cleaned up (see review comments above).

zulissimeta:
There are two stages here:

To calculate an elastic tensor, you should provide the stress of the relaxed configuration. This is especially important if users do not set the convergence settings tight enough (eg fmax=0.01 or less). So, for the undeformed structure, you should either
1. Relax -> static
2. Relax
3. static (if you want to calculate elastic properties of a non-relaxed structure?)

For the deformed structures, you should either
1. Relax -> static
2. Relax

I think this is mostly a documentation problem for what "run_static" should mean - I'm interpreting it here as "run static after relaxation for deformed structures".


> @zulissimeta: Thanks! I had a few minutes to give this another look over. We're nearly there.
> 
> First, could you see my two unresolved comments below:
> 
> * https://github.com/Quantum-Accelerators/quacc/pull/2693/files#r1987581514
> * https://github.com/Quantum-Accelerators/quacc/pull/2693/files#r1987582170
> 
> For the first one, we are having `bulk_to_deformations_subflow` always take a `static_job` even though running a static calculation is purely optional. This is not ideal. For instance, let's say a set of recipes does not even have a static job made --- now they will have to make one just to have it be ignored by this subflow. We want to avoid such a scenario, which is why it was originally set to `Job | None = None`. The original method seems more suitable in my opinion since one cannot run a static calculation if no static job is provided. But I am open to suggestions.
> 
> The other review issue to resolve is with regards to some `if/else` code blocks that can be cleaned up (see review comments below).

src/quacc/recipes/emt/elastic.py
zulissimeta: resolved!

zulissimeta:
> For the first one, we are having `bulk_to_deformations_subflow` always take a `static_job` even though running a static calculation is purely optional. This is not ideal. For instance, let's say a set of recipes does not even have a static job made --- now they will have to make one just to have it be ignored by this subflow. We want to avoid such a scenario, which is why it was originally set to `Job | None = None`. The original method seems more suitable in my opinion since one cannot run a static calculation if no static job is provided. But I am open to suggestions.

I think I got both of these too!

src/quacc/recipes/common/elastic.py
```diff
@@ -4,24 +4,84 @@
...
 def bulk_to_deformations_subflow(
-    atoms: Atoms,
+    undeformed_result: RunSchema,
     relax_job: Job,
     static_job: Job | None = None,
+    run_static: bool = False,
     deform_kwargs: dict[str, Any] | None = None,
-) -> list[dict]:
+) -> ElasticSchema:
     """
```
Andrew-S-Rosen, line 85: The function signature is still not entirely logical here. For instance, a user can pass `run_static=True` without specifying a `static_job`, and then it will crash. Alternatively, a user can pass a `static_job` but this will do nothing when `run_static=False`. We want to avoid these kinds of redundancies, without requiring the user to always be providing a `static_job` that may never be used.

The reason that the code originally had `static_job: Job | None = None` is that when `static_job is None`, it would not do a static job:

```python
if static_job is not None:
    result = static_job(result["atoms"])
```

In other words, when a user supplied a `static_job` keyword argument, it would be used as expected. In the current PR, we have two keyword arguments that have overlapping but conflicting purposes. I still think we should revert to the old format for this `@subflow` where we have `static_job: Job | None = None` and no `run_static` keyword argument for the reasons outlined above. Please let me know if I am misunderstanding something here.

src/quacc/recipes/emt/elastic.py
```diff
@@ -15,17 +15,18 @@
...
-    from quacc.types import OptSchema, RunSchema
+    from quacc.types import ElasticSchema
 
 
 @flow
 def bulk_to_deformations_flow(
     atoms: Atoms,
-    run_static: bool = True,
+    pre_relax: bool = True,
+    run_static: bool = False,
```
Andrew-S-Rosen, line 25: Ideally, one should know what the keyword arguments mean without having to read the docstring. If these are only for the bulk structure before deformation, we should give them suitable names. For instance, `pre_relax_bulk` and `run_static_bulk` or `pre_relax_undeformed` and `run_static_undeformed` or something like that (with the docstring updated accordingly).

src/quacc/recipes/mlp/elastic.py
```diff
@@ -0,0 +1,88 @@
...
+
+    1. Deformed structures generation
+
+    2. Deformed structures relaxations
+        - name: "relax_job"
+        - job: [quacc.recipes.mlp.core.relax_job][]
+
+    3. Deformed structures statics (optional)
+        - name: "static_job"
+        - job: [quacc.recipes.mlp.core.static_job][]
```
Andrew-S-Rosen, line 41: Both here and in `bulk_to_deformations_flow` in src/quacc/recipes/emt/elastic.py, the docstring is no longer complete with your introduced changes. The workflow now consists of up to 5 steps. This should be updated in both sections.

Andrew-S-Rosen (commented):
@zulissimeta: Thanks! A few remaining thoughts below.

Andrew-S-Rosen:
@zulissimeta: Also, it is worth noting that you have modified the various flows/subflows such that their names are no longer logical. For instance, they were originally `bulk_to_deformations_flow` and `bulk_to_deformations_subflow`. They should really be renamed something like `elastic_flow` and `elastic_subflow` if their purpose is not to generate deformed structures (as was originally the case) but rather to generate elastic data.

zulissimeta:
> @zulissimeta: Also, it is worth noting that you have modified the various flows/subflows such that their names are no longer representative. For instance, they were originally `bulk_to_deformations_flow` and `bulk_to_deformations_subflow`. They should really be renamed something like `elastic_flow` and `elastic_subflow` if their purpose is not to generate deformed structures (as was originally the case) but rather to generate elastic data.

Done!

Andrew-S-Rosen:
@zulissimeta: Thanks! Looks good. I will go ahead and merge.

A minor point:

```
    Workflow consisting of:
    1. Deformed structures generation
    2. Deformed structures relaxations
        - name: "relax_job"
        - job: [quacc.recipes.emt.core.relax_job][]
    3. Deformed structures statics (optional)
        - name: "static_job"
        - job: [quacc.recipes.emt.core.static_job][]
    4. Elastic tensor calculation
```

The above docstring is missing the (optional) relaxation of the undeformed structure and (required) static calculation of the undeformed structure, coming before Step 1 in the list above. Part of the reason to enumerate these is that it tells the user what the underlying function is (e.g. `quacc.recipes.emt.core.relax_job`) and the name (e.g. `"relax_job`") needed when one wants to update the defaults via `job_params`. We can take care of that in a future PR. 

Hopefully this will all be a bit easier, with less code duplication, once I can take care of #2701. I will start chipping away at that in earnest next week.

PR #2695, opened 2025-03-07
Freeze arguments to pick_calculator

zulissimeta:
## Summary of Changes

`lru_cache` is used to make sure that each MLP calculator is only instantiated once (great!). However, if you want to pass dictionaries through to the underlying calculator (say, config overrides to the fairchem calculator), you will get errors because dictionaries are mutable and lru_cache is unhappy. We can patch this by converting any dictionary kwargs to a frozendict.

### Requirements


Andrew-S-Rosen:
@zulissimeta: Interesting. Thank you for the PR here.

A question: what is the benefit in using `frozendict` as opposed to `frozenset` without requiring an additional dependency? For instance,

```python
d = {"cow": "moo", "cat": "meow"}
d_frozen = frozenset(d.items()) # frozen now!
d_unfrozen = dict(d_frozen) # same as d again!
```

You would still need a decorator, but I haven't quite figured out yet the value-add for `frozendict`. If we do need `frozendict` though, this should be added to `pyproject.toml` as well.

Andrew-S-Rosen:
Ah, `frozenset` will not work on a nested `dict` like `d={"cow": {"moo": 1}}`.

While not particularly elegant, I am okay with the proposed approach if there's no other option since `frozendict` seems reasonably well maintained. Not a huge fan of having another dependency just for this, but I don't know of a better solution. Happy to merge with a `pyprojoect.toml` update and your approval that it's ready.

zulissimeta:
Thanks, think it should be good to go now.

zulissimeta:
Thanks, think I got the test. Would be good to confirm that the fairchem test did run in the HPC test too.

Andrew-S-Rosen:
Thanks, @zulissimeta! I will go ahead and merge this. As a head's up, I have not implemented the `fairchem` tests on the HPC runner yet. It's going to take more than a few minutes like I had anticipated. Right now, it will run on `main` though, which I will confirm momentarily.

PR #2733, opened 2025-03-27
enforce relax_cell False for correct elastic calculations

lbluque:
## Summary of Changes

`relax_cell=False` should be enforced for relaxations of deformed structures in the `_elastic_tensor_subflow`

### Requirements

- [X] My PR has relevant, comprehensive [unit tests](https://quantum-accelerators.github.io/quacc/dev/contributing.html#unit-tests).

### DIFF
--- a/src/quacc/recipes/common/elastic.py
+++ b/src/quacc/recipes/common/elastic.py
@@ -115,7 +115,7 @@ def _elastic_tensor_subflow(
 
     results = []
     for deformed in deformed_structure_set:
-        result = relax_job(deformed.to_ase_atoms())
+        result = relax_job(deformed.to_ase_atoms(), relax_cell=False)
 
         if static_job is not None:
             result = static_job(result["atoms"])


buildbot-princeton:

Andrew-S-Rosen:
Thanks, @lbluque! I think the default is `relax_cell=False` for all recipes, but I suppose this will catch mistakes by the end-user. Is that the intention? Happy to merge regardless.

lbluque:
That was quick @Andrew-S-Rosen! 

> Thanks, @lbluque! I think the default is `relax_cell=False` for all recipes, but I suppose this will catch mistakes by the end-user. Is that the intention? Happy to merge regardless.
Correct, the default runs as intended. This is meant to protect against user mistakes.

After some additional thought I would say it might also be worth removing the enforced `relax_cell=True` for the first relaxation  here: https://github.com/Quantum-Accelerators/quacc/blob/a507927045363436c9b0bdf49cdfee1c3b28bc31/src/quacc/recipes/common/elastic.py#L61

It is a perfectly valid approach to relax only atomic positions before generating the strained structures if a user wants to. Right if a users wants to use `relax_cell=False` by setting the `relax_job` accordingly for this first relaxation, the flow will quietly enforce  `relax_cell=True` instead.

I think in general, any time a job parameter is explicitly enforced in the code a user should be warned some way if the parameter was changed wrt to the ones in the job given by the user.

Curious to hear your thoughts on this

Andrew-S-Rosen:
@lbluque: Thanks for your thoughtful comment! In short, I completely agree with you. I try not to enforce anything on the user unless it's a guaranteed mistake, but I must have missed this one in refactoring @zulissimeta's workflow. We should remove the `relax_cell=True` call there. However, I am a bit concerned about adding another somewhat vague keyword argument here. I'm not sure offhand the most elegant way to address this, but perhaps I'll have a better idea when it's not midnight 😅 Open to suggestions!! 

> I think in general, any time a job parameter is explicitly enforced in the code a user should be warned some way if the parameter was changed wrt to the ones in the job given by the user.

Absolutely, although in this case it wouldn't matter. If the user did set `relax_cell` to `False`, an error would be raised due to conflicting keyword arguments.

PR #2813, opened 2025-05-28
Add support for FAIR chem v2 calculators (eg UMA).

Andrew-S-Rosen:
Following up on #2812.


Andrew-S-Rosen:
@zulissimeta: Thanks for the suggestion to run in a separate branch. I have done so here and got back a cuda error. Any suggestions on how to best proceed? https://github.com/Quantum-Accelerators/quacc/actions/runs/15310913002/job/43075129580?pr=2813

lbluque:
> @zulissimeta: Thanks for the suggestion to run in a separate branch. I have done so here and got back a cuda error. Any suggestions on how to best proceed? https://github.com/Quantum-Accelerators/quacc/actions/runs/15310913002/job/43075129580?pr=2813

Hi @Andrew-S-Rosen - the issue is we currently set the default device to "cuda", but it's better to set the default based on whats available! I just added that here https://github.com/facebookresearch/fairchem/pull/1202

In case you want this to work before we make a release with the above changes on the fairchem side, you can always set the device to cpu in the CI tests:
```python
predict_unit = pretrained_mlip.get_predict_unit(model_name=..., device="cpu")
```

PR #2977, opened 2025-10-14
Refactor convergence check to include gradient handling

Andrew-S-Rosen:


AitElBadaoui:
Hello @Andrew-S-Rosen,
You need to bump the pydantic version to 2.12.2 (today version), they have some broken wheels in version 2.12.1, that's why all the CI fails

Andrew-S-Rosen:
Thanks @AitElBadaoui. I am trying to fix the bug ASE just introduced. I just bumped pydantic now.

PR #3005, opened 2025-11-04
Torchsim support in Quacc

orionarcher:
## Summary of Changes

This PR introduces support for the 3 high-level runners in TorchSim `optimize`, `integrate`, and `static`. It attempts to provide an easy to use interface for quacc users that relies on minimal TorchSim imports. It also adopts the quacc nomenclature (`relax_job`, `md_job`, and `static_job`) and IO (input atoms, return dicts).

I initially implemented a TorchSim runner but it ended up being just an extra unecessary layer so I decided recipes would be the best way forward.

Happy to take any feedback!

### Requirements

### DIFF
--- a/src/quacc/recipes/torchsim/_base.py
+++ b/src/quacc/recipes/torchsim/_base.py
@@ -212,7 +212,7 @@ def process_trajectory_reporter_dict(
 
 @requires(has_torchsim, "torch_sim is required for this function")
 def pick_model(
-    model_type: TSModelType, model_path: str | Path, **model_kwargs: Any
+    model_type: TSModelType, model: str | Path, **model_kwargs: Any
 ) -> ModelInterface:
     """Pick and instantiate a model based on the model type.
 
@@ -220,8 +220,9 @@ def pick_model(
     ----------
     model_type : TSModelType
         The type of model to instantiate.
-    model_path : str | Path
-        Path to the model file or checkpoint.
+    model : str | Path
+        Path to the model file or checkpoint. For some models, string names may
+        be allowed, such as "uma-s-1" for FairChemModel.
     **model_kwargs : Any
         Additional keyword arguments to pass to the model constructor.
 
@@ -238,39 +239,39 @@ def pick_model(
     if model_type == TSModelType.FAIRCHEMV1:
         from torch_sim.models.fairchem_legacy import FairChemV1Model
 
-        return FairChemV1Model(model=model_path, **model_kwargs)
+        return FairChemV1Model(model=model, **model_kwargs)
     if model_type == TSModelType.FAIRCHEM:
         from torch_sim.models.fairchem import FairChemModel
 
-        return FairChemModel(model=model_path, **model_kwargs)
+        return FairChemModel(model=model, **model_kwargs)
     if model_type == TSModelType.GRAPHPESWRAPPER:
         from torch_sim.models.graphpes import GraphPESWrapper
 
-        return GraphPESWrapper(model=model_path, **model_kwargs)
+        return GraphPESWrapper(model=model, **model_kwargs)
     if model_type == TSModelType.MACE:
         from torch_sim.models.mace import MaceModel
 
-        return MaceModel(model=model_path, **model_kwargs)
+        return MaceModel(model=model, **model_kwargs)
     if model_type == TSModelType.MATTERSIM:
         from torch_sim.models.mattersim import MatterSimModel
 
-        return MatterSimModel(model=model_path, **model_kwargs)
+        return MatterSimModel(model=model, **model_kwargs)
     if model_type == TSModelType.METATOMIC:
         from torch_sim.models.metatomic import MetatomicModel
 
-        return MetatomicModel(model=model_path, **model_kwargs)
+        return MetatomicModel(model=model, **model_kwargs)
     if model_type == TSModelType.NEQUIPFRAMEWORK:
         from torch_sim.models.nequip_framework import NequIPFrameworkModel
 
-        return NequIPFrameworkModel(model=model_path, **model_kwargs)
+        return NequIPFrameworkModel(model=model, **model_kwargs)
     if model_type == TSModelType.ORB:
         from torch_sim.models.orb import OrbModel
 
-        return OrbModel(model=model_path, **model_kwargs)
+        return OrbModel(model=model, **model_kwargs)
     if model_type == TSModelType.SEVENNET:
         from torch_sim.models.sevennet import SevenNetModel
 
-        return SevenNetModel(model=model_path, **model_kwargs)
+        return SevenNetModel(model=model, **model_kwargs)
     if model_type == TSModelType.LENNARD_JONES:
         from torch_sim.models.lennard_jones import LennardJonesModel
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ redun = ["redun>=0.16.2"]
 sella = ["sella>=2.3.5"]
 sevenn = ["sevenn>=0.10.1"]
 tblite = ["tblite>=0.3.0; platform_system=='Linux'"]
-torchsim = ["torch-sim-atomistic>=0.4.1"]
+torchsim = ["torch-sim-atomistic>=0.4.2"]
 dev = ["pytest>=7.4.0", "pytest-cov>=7.0.0", "ruff>=0.0.285", "pytest-asyncio>=0.26.0"]
 docs = [
     "blacken-docs>=1.16.0",


src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,318 @@
...
+    import torch
+    from ase.atoms import Atoms
+    from torch_sim.integrators import Integrator
+    from torch_sim.optimizers import Optimizer
+    from torch_sim.trajectory import TrajectoryReporter
+    from torch_sim.typing import StateLike
+
+    class TorchSimSchema(TypedDict):
+        atoms: list[Atoms]
+        model: str  # string of the model name
```
Andrew-S-Rosen: Can we have this be a `Literal` of all the supported ones here?

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,318 @@
...
+        temperature: float | list[float]
+        timestep: float
+        trajectory_reporter: TrajectoryReporterDetails | None
+        autobatcher: AutobatcherDetails
+        integrator_kwargs: dict[str, Any]
+
+    class TorchSimStaticSchema(TorchSimSchema):
+        all_properties: list[dict[str, torch.Tensor]]
+        trajectory_reporter: TrajectoryReporterDetails | None
+        autobatcher: AutobatcherDetails
```
Andrew-S-Rosen: If you think these type hints might get used in several places, they can go in `quacc.types`.

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,318 @@
+from __future__ import annotations
```
Andrew-S-Rosen: Please add a docstring to the top of the module so it is shown in the docs.

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,318 @@
...
+        max_steps: int
+        steps_between_swaps: int
+        trajectory_reporter: TrajectoryReporterDetails | None
+        autobatcher: AutobatcherDetails
+        init_kwargs: dict[str, Any] | None
+        optimizer_kwargs: dict[str, Any]
+
+    class TorchSimIntegrateSchema(TorchSimSchema):
+        integrator: Integrator
+
```
Andrew-S-Rosen: Extraneous line break.

Andrew-S-Rosen (commented):
Hi @orionarcher! Thank you very much for this PR!

Overall, this is looking to be in relatively good shape. I have left some minor cleanup related comments below.

One larger question: by avoiding the use of runners entirely, there is no directory management. I assume there is some I/O being done? Everything will run in the current working directory at the moment, which would problematic if there is I/O. You will see that the various runners have a `setup` and `cleanup` method that they call from the `BaseRunner` class to take care of file management. Is there I/O to worry about here, or is it all kept in-memory?

Another comment: some of the `@job`-decorated recipes take classes as arguments. If these are classes that are not instantiated, that's fine. But it looks like some of these arguments are instantiated classes (e.g. the `autobatcher`). This is problematic for workflow tools, especially those from the MP side, because it is not possible to (de)serialize an instantiated class. The workaround from MP tooling is, as you know, that such methods need to have an `.as_dict()` and `.from_dict()` class. It's okay if internally called functions have such classes, but the main `@job`-decorated function cannot have them unless they are monty serializable. Any thoughts about how to best address this?

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,318 @@
...
+    )
+
+    model = model if isinstance(model, ModelInterface) else pick_model(*model)
+
+    state = ts.initialize_state(atoms, model.device, model.dtype)
+    if autobatcher:
+        autobatcher = ts.runners._configure_batches_iterator(
+            state, model, autobatcher=autobatcher
+        )  # type: ignore
+        autobatcher_dict = _get_autobatcher_dict(autobatcher)
```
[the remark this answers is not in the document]
orionarcher: Done! Moved this all out to _base.py along with the type hints (some of which needed to be there.)

src/quacc/recipes/torchsim/core.py
orionarcher: If we rewrite the MLIP interface to use TorchSim that might be necessary but otherwise I'd hold off to keep the import localized.

Andrew-S-Rosen:
Thanks, @orionarcher! I will review soon.

> There is no extraneous IO here (i.e. input sets) but we do write out trajectory files as output. Users must provide names for the trajectories. If they use something like "file.traj" it will end up at "./file.traj", of course they could also specify "/exact/path/file.traj". In my mind, this is the correct behavior. Workflow runners will write out un-rooted paths to the current dir and users still have control if they want to hard code it. Does that make sense to you?

Unfortunately, this will not work out because it requires the workflow runner to essentially `cd` into a new directory for every run, which is not what happens across the board. It is what happens with FireWorks but not for most other tools. In the more general case, we will have `file.traj` overwriting itself over and over. Everything related to paths must be handled by quacc and be an absolute path (to avoid multithreading issues where `cd` calls are a nightmare...), which is why the `Runner` classes inherit from the `BaseRunner`, which has a `setup` and `cleanup` method that automatically handles directory setup and cleanup. The added benefit of this approach is that it allows quacc to manage different directories for where calculations should be run vs. where they should be stored after the run.

I will try to provide some guidance about how to address this. I do not think it will be especially challenging. It probably means just creating a `TorchSimRunner` that inherits from `BaseRunner` with basically no extra features. We use the `TorchSimRunner` to setup the directory, pass the runtime directory path to the trajectory name like [here](https://github.com/Quantum-Accelerators/quacc/blob/53f2053c6703eb90af06ada7a922f4de46df851e/src/quacc/runners/ase.py#L234) and don't allow the user to overwrite it. I'm happy to provide more direct suggestions next week!

src/quacc/recipes/torchsim/_base.py
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+
+@requires(has_torchsim, "torch_sim is required for this function")
+def process_trajectory_reporter_dict(
+    trajectory_reporter_dict: TrajectoryReporterDict | None,
+) -> tuple[TrajectoryReporter, TrajectoryReporterDetails]:
+    """Process the input dict into a TrajectoryReporter and details dictionary."""
+    if trajectory_reporter_dict is None:
+        return None, None
```
Andrew-S-Rosen: Incompatible return type with type hint.

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,364 @@
...
+        Keyword arguments to pass to the optimizer step function.
+
+    Returns
+    -------
+    TorchSimOptSchema
+        A dictionary representing the final atoms configuration and metadata geometry
+        optimization schema. See the type-hint for the data structure.
+    """
+    runner = BaseRunner()
+    runner.setup()
```
Andrew-S-Rosen: Here, quacc sets up a unique temporary directory where all file I/O is meant to be generated during runtime. The directory path is assigned to `runner.tmpdir`, and the base path for this unique temporary directory is controlled by the `SCRATCH_DIR` pydantic setting in quacc.

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,364 @@
...
+        **convergence_fn_kwargs or {}
+    )
+
+    state = ts.initialize_state(atoms, model.device, model.dtype)
+
+    # Configure trajectory reporter
+    trajectory_reporter_dict = trajectory_reporter_dict or {}
+    trajectory_reporter_dict["filenames"] = [
+        runner.tmpdir / f"trajectory_{i}.h5md" for i in range(len(atoms))
+    ]
```
Andrew-S-Rosen: Here, we ensure that all files generated at runtime are being written out to the unique temporary directory. It is important that quacc controls all file I/O because that is the only way to prevent concurrency issues. For instance, if the user decides to run 100 jobs in the current working directory without any sort of workflow engine, we want to ensure that quacc prevents those jobs from clashing with one another. That is why quacc handles the file handling. In the case of a workflow engine that handles directory management (e.g. FireWorks, Jobflow-Remote), quacc will let the workflow engine handle it all instead. But not all workflow engines support this, and we need to ensure that someone using no workflow engine will still be able to launch jobs without concern.

src/quacc/recipes/torchsim/core.py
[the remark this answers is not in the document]
orionarcher: I hear you on the modularity point. I've cleaned up the logic to eliminate step 3, simplify step 4, and add the quacc version.

At this point, I feel there is a tradeoff between modularity and readability. Yes, there is some duplicate code in each function, but it's also quite clear what each function is doing. I've spent a while staring at this because I really do believe a modular solution is generally best. At this point, I believe any additional deduplication would require creating a somewhat redundant extra layer of functions, likely with many arguments. From what I can tell, `quacc.recipes.mlp.core` manages to be very readable and low duplication because most of the relevant logic is folded into `runner.run_opt` and `runner.run_dyn`, whereas in the `torchsim` code there isn't an additional execution layer. I could definitely add another layer and reduce the amount of code in the recipes, but I worry I'd just be hiding the logic elsewhere. (I am thinking of jobflow, a very elegant package that I find veeerrry hard to follow). The three functions have different enough signatures that it's not clear to me this would reduce LOC.

As I am imagining this API, I will not need to add additional recipes. TorchSim has three core functions in the high-level API that are all supported here.

I don't think there is a right answer here, we may fall at different points on the modularity vs readability tradeoff. My gut is telling me this is the best middle ground but I am happy to change things up with another layer of logic if that's what you'd prefer. Just lmk!

orionarcher:
Thanks for the feedback, I responded to all the comments. Let me know what you think!

Tests are failing for Prefect for reasons that aren't clear to me. Do you think it's a flake or something in here?

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,336 @@
...
+        TrajectoryReporterDict,
+    )
+
+
+@job
+@requires(has_torchsim, "torch_sim is required for this function")
+def relax_job(
+    atoms: list[Atoms],
+    model_type: TSModelType,
+    model_path: str | Path,
```
Andrew-S-Rosen: @orionarcher Do all models need the `model_path`? For instance, the fairchem models don't require a model path. You can pass `name_or_path="uma-s-1"`, [for instance](https://github.com/facebookresearch/fairchem/blob/a9287b37469a80f9286151d70ceb184f91dfad8e/src/fairchem/core/calculate/ase_calculator.py#L124). This makes me feel that the recipes should support this as well. Do you agree?

src/quacc/recipes/torchsim/core.py
```diff
@@ -0,0 +1,336 @@
...
+    process_binning_autobatcher_dict,
+    process_in_flight_autobatcher_dict,
+    process_trajectory_reporter_dict,
+)
+from quacc.runners._base import BaseRunner
+from quacc.schemas.torchsim import CONVERGENCE_FN_REGISTRY, ConvergenceFn, TSModelType
+
+has_torchsim = bool(find_spec("torch_sim"))
+if has_torchsim:
+    import torch_sim as ts
```
Andrew-S-Rosen: I agree with Aaron's comment on the atomate2 PR that it'd be nice not to call this `ts`. Leaving it as `torch_sim` seems fine.

src/quacc/recipes/torchsim/core.py
Andrew-S-Rosen: @sihoonchoi thanks for testing.

>It took some time to identify the appropriate arguments, but this shouldn't be an issue for users familiar with TorchSim.

We should assume the end user has no knowledge of TorchSim. Is there anything that we can do on the quacc side (e.g. docstrings) where you think it would have made the experience easier? Any things you were particularly unsure about?

Andrew-S-Rosen: Thanks for your input, @sihoonchoi. I must admit that I do not really understand the value proposition for using `StrEnum`s when we can use a `str` with a `Literal["energy", "force"]` type hint.

Andrew-S-Rosen:
Great, thank you @orionarcher for your hard work on this! I will plan to merge this now. If you think of any improvements as you work on the atomate2 PR, please consider adding them here!

orionarcher:
Woo! Thanks @Andrew-S-Rosen!

Yep, will plan to keep both in sync as much as possible!

PR #3101, opened 2026-01-09
Add a workflow engine-safe way of specifying `copy_files`

vineetbansal:
WIP to get CI going on this..

### DIFF
--- a/tests/core/recipes/espresso_recipes/test_phonons.py
+++ b/tests/core/recipes/espresso_recipes/test_phonons.py
@@ -715,7 +715,7 @@ def test_phonon_dvscf_q2r_inplace(tmp_path, monkeypatch):
 
         dvscf_q2r_results = dvscf_q2r_job(
             prev_outdir=c_scf_results["dir_name"],
-            copy_files={c_ph_results["dir_name"]: "*"},
+            copy_files=c_ph_results["dir_name"],
             **dvscf_q2r_params,
         )
 
--- a/tests/core/recipes/espresso_recipes/test_core.py
+++ b/tests/core/recipes/espresso_recipes/test_core.py
@@ -102,7 +102,7 @@ def test_static_job_v2(tmp_path, monkeypatch):
     assert "kpts" not in results["parameters"]
     assert results["parameters"]["kspacing"] == 0.5
 
-    pp_results = post_processing_job(copy_files={results["dir_name"]: "*"})
+    pp_results = post_processing_job(copy_files=results["dir_name"])
     assert Path(pp_results["dir_name"], "pseudo_charge_density.cube.gz").is_file()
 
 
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
   tests-engines:
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
```
Andrew-S-Rosen: Let's turn this back to `true`

tests/core/runners/test_prep.py
Andrew-S-Rosen: Same comment here. We have removed old tests. We should retain the old tests but add in new ones.

tests/core/runners/test_ase.py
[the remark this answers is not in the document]
vineetbansal: I've added a `tests/core/wflow/test_copy_files.py`, i.e. a non-engine equivalent of `tests/<engine>/test_copy_files.py`. This tests the dict case. But I agree about retaining the old tests. I'll fix that in the next commit.

src/quacc/recipes/vasp/mp_legacy.py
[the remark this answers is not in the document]
Andrew-S-Rosen: Jobflow is only looking at the positional arguments and keyword arguments of the `@job`-decorated function. In the example I mentioned, you are passing `copy_files=Copy({prev_dir: ["WAVECAR*"]}) if prev_dir else None` to an internal function `run_and_summarize` that has no decorator. By this point, `prev_dir` would have already been resolved and there is no future-like object anymore.

Unless I've misunderstood, we should only be updating the code when `copy_files` is an argument to a `@job`-decorated function.

vineetbansal:
Thanks @Andrew-S-Rosen - I've addressed these concerns in the latest commit.

After stripping off unnecessary `Copy` invocations, looks like there's only one left - `double_relax_flow` in `vasp/core.py` calling `relax_job` with `copy_files=Copy({summary1["dir_name"]: ["WAVECAR*"]})`.

tests/core/recipes/espresso_recipes/test_phonons.py
```diff
@@ -715,7 +716,7 @@ def test_phonon_dvscf_q2r_inplace(tmp_path, monkeypatch):
 
         dvscf_q2r_results = dvscf_q2r_job(
             prev_outdir=c_scf_results["dir_name"],
-            copy_files=c_ph_results["dir_name"],
+            copy_files=Copy({c_ph_results["dir_name"]: "*"}),
```
Andrew-S-Rosen: Is there a need to change?

vineetbansal:
@Andrew-S-Rosen - I've made the changes.

PR #3139, opened 2026-02-10
Autodiscover paths in the no-engine case

vineetbansal:
## Summary of Changes

Adds support for inspecting the current running context using:

- `get_directory_context()` <- The top-level directory specified by the `@flow`.
- `get_context_path()` <- The position of the running code relative to the top-level `@flow` as `/`-separated string elements.

Together, these allow the code to create a folder using - `settings.RESULTS_DIR / Path(get_directory_context()) / get_context_path()`.

`calc_setup` uses this to create timestamped folders for output.

Still a WIP. I need to add support for a few things we discussed, and add tests for these:

- `@job`s might be running without an encompassing `@flow` and need to be able to set their own top-level directory.
- A `@flow` might have several `@subflow`s with the same name. Thus the subflow level folders need to have a timestamp appended to them, just like the folders at the `@job` level.
- Add tests in `test_autodiscover_paths` to cover the case of `QUACC_AUTODISCOVER_DIR` to `True`/`False` to demonstrate the directory structure in both modes. Add tree to their docstrings so it's clear what to expect in each case.

### DIFF
--- a/tests/requirements-mace.txt
+++ b/tests/requirements-mace.txt
@@ -1,2 +1,2 @@
-mace-torch==0.3.14
+mace-torch==0.3.15
 mace-models==0.1.6
--- a/tests/requirements-parsl.txt
+++ b/tests/requirements-parsl.txt
@@ -1,1 +1,1 @@
-parsl[monitoring]==2026.2.9
+parsl[monitoring]==2026.2.16
--- a/tests/requirements-torchsim.txt
+++ b/tests/requirements-torchsim.txt
@@ -1,2 +1,2 @@
 torch-sim-atomistic==0.5.0
-mace-torch==0.3.14
+mace-torch==0.3.15
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -5,7 +5,7 @@ frozendict==2.4.7
 monty==2025.3.3
 psutil==7.2.2
 pydantic==2.12.5
-pydantic-settings==2.13.0
+pydantic-settings==2.13.1
 pymatgen==2025.10.7
 ruamel.yaml==0.19.1
-typer==0.17.4
+typer==0.24.1


Andrew-S-Rosen:
Thanks, @vineetbansal! One other ToDo here: can you also update the docs with info about your new feature? https://quantum-accelerators.github.io/quacc/user/settings/file_management.html

vineetbansal:
@Andrew-S-Rosen - yes this is ready to be reviewed now.

Two things that we can either do as part of this PR or later:
- One of the issues raised during the meeting was the ability to assign the "top-level" folder name within `RESULTS_DIR` instead of it being auto-generated from the flow name. This should be possible with a few lines of code.
- It is not too much more code (nowhere compared to PR #3137) to support all this for `jobflow` as well. Let me know if this is a good time to put that in (along with tests).

docs/user/settings/file_management.md
```diff
@@ -57,6 +136,8 @@ Here, let's assume the user has specified the `SCRATCH_DIR` setting to be a cust
 
 At job runtime, the file structure looks like:
 
+If `AUTODISCOVER_DIR` is `false`:
```
Andrew-S-Rosen: Can we make this `False` instead?

docs/user/settings/file_management.md
```diff
@@ -70,12 +151,52 @@ SCRATCH_DIR
     └── OUTPUT
 ```
 
-Here, the `symlink-tmp-quacc-2023-12-08-67890` is a temporary symbolic link that points to `SCRATCH_DIR/tmp-quacc-2023-12-08-67890` so you can easily monitor the progress of the calculation. On Windows, no symbolink link is created, but the `tmp-quacc-2023-12-08-67890` directory is still created in `SCRATCH_DIR`.
+If `AUTODISCOVER_DIR` is `true`:
```
Andrew-S-Rosen: Can we make this `True` instead?

docs/user/settings/file_management.md
```diff
@@ -88,9 +209,25 @@ SCRATCH_DIR
...
+SCRATCH_DIR
+│
+```
+
 ### Job Failure
 
-If the job fails or does not complete, then the file structure looks like:
+If the job fails or does not complete, then the file structure looks as follows.
+
+If `AUTODISCOVER_DIR` is `false`:
```
Andrew-S-Rosen: `False`

docs/user/settings/file_management.md
```diff
@@ -104,3 +241,43 @@ SCRATCH_DIR
 │   ├── INPUT
     └── OUTPUT
 ```
+
+If `AUTODISCOVER_DIR` is `true`:
```
Andrew-S-Rosen: `True`

docs/user/settings/file_management.md
```diff
@@ -104,3 +241,43 @@ SCRATCH_DIR
...
+        │         ├── opt.log.gz
+        │         └── opt.traj.gz
+        ├── relax_job-2026-02-16-42525
+        │         ├── opt.json.gz
+        │         ├── opt.log.gz
+        │         └── opt.traj.gz
+        ...
+```
+
+The `symlink-` points to the corresponding `failed-..` job folder in `SCRATCH_DIR`.
```
Andrew-S-Rosen: `failed-...`

src/quacc/wflow_tools/context.py
```diff
@@ -0,0 +1,194 @@
...
+    # delegate to the original function directly.
+    if not settings.AUTODISCOVER_DIR:
+        return func(*args, **kwargs)
+
+    # Create a unique name we can use at this level.
+    name = make_unique_name(prefix=f"{func.__name__}-")
+
+    if is_top_level():
+        # This is the outermost tracked call: create a unique root
+        # directory (e.g. ``quacc-abc123/``) and initialize both the
```
Andrew-S-Rosen: I think this might need updating.

tests/core/wflow/test_autodiscover_paths.py
Andrew-S-Rosen: Is there any way we can test the symlinking?

Andrew-S-Rosen (commented):
Hi @vineetbansal! I have some comments worth addressing below. As for your "Two things that we can either do as part of this PR or later", please see my comment above.

## Directory in Output

When I run the following example, the output schema (JSON) has a `dir_name` field that includes the directory. When I run with `AUTODISCOVER_DIR` set to `False`, the full path is shown (this is the desired behavior). When I run with `AUTODISCOVER_DIR` set to `True`, the relative path is shown. Can you ensure that when it gets stored in the output, it is resolved so that the user can find the files afterwards? 

Note that this is also a relevant concern with the logging, where it does things like

```
INFO:quacc.runners.prep:Moving C:\Users\asros\Desktop\tmp-bulk_to_slabs_flow-2026-02-22-01-32-35-651667-90541\bulk_to_slabs_subflow-2026-02-22-01-32-35-652280-77098\relax_job-2026-02-22-01-32-35-825258-48557 contents to bulk_to_slabs_flow-2026-02-22-01-32-35-651667-90541\bulk_to_slabs_subflow-2026-02-22-01-32-35-652280-77098\relax_job-2026-02-22-01-32-35-825258-48557
```

```python
from quacc.recipes.emt.core import relax_job
from ase.build import bulk

atoms = bulk("Cu")
output = relax_job(atoms)
print(output["dir_name"])
```

## Tmp Path Naming on Failed Jobs

There is now a consideration that we did not need to think about before. Namely, if someone runs a flow that launches 10 independent and concurrent jobs, they might (or might not) be fine with 9/10 finishing depending on whether the data for the 9/10 might still be useful. 

As it stands right now, the parent flow will stay in the scratch directory with the `tmp- name if 1/10 jobs fails. Admittedly, I am not sure really what to do about this and am open to suggestions. I will need to think on this some more too.

```python
from ase.build import bulk

atoms = bulk("Xe")
bulk_to_slabs_flow(atoms)
```

PR #3203, opened 2026-04-17
Use fairchem-data-omc in the fairchem recipes

Andrew-S-Rosen:


Andrew-S-Rosen:
Requires https://github.com/facebookresearch/fairchem/pull/1970

PR #3238, opened 2026-05-14
publish a wheel

dimbleby:
publish a wheel (in addition to a source distribution)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,8 +22,8 @@ jobs:
           pip install uv
           uv pip install --system "quacc @ ."
 
-      - name: Build sdist
-        run: python -m build --sdist
+      - name: Build distributions
+        run: python -m build
 
       - name: Publish package
         uses: pypa/gh-action-pypi-publish@release/v1


buildbot-princeton:

Andrew-S-Rosen:
What is the benefit of doing so?

dimbleby:
In python, all package installation is from wheel.  If you don't publish a wheel, every user must build the wheel at install time.

That's slower, and can go wrong.  Eg if setuptools release a breaking change - either intentionally or not - then without a published wheel your package suddenly becomes uninstallable.

Aside: setuptools _is_ giving deprecation warnings and promising to break your build in February 2027.  You will want to address that: and it shows a good reason to have building the wheel be part of your CI pipeline.

Andrew-S-Rosen:
> Aside: setuptools is giving deprecation warnings and promising to break your build in February 2027. You will want to address that: and it shows a good reason to have building the wheel be part of your CI pipeline.

Any info on what the error is and how to address?

dimbleby:
> Any info on what the error is and how to address?

Run `python -m build` and read the output!

PR #3302, opened 2026-06-24
Update Python version to 3.13 in workflow

Andrew-S-Rosen:


Andrew-S-Rosen:
@zulissimeta --- just a head's up that your Ray serve tests fail on Python 3.13.

Andrew-S-Rosen:
Never mind. It is also failing in `main` for some reason. Seems like a fragile test at the moment. I think it is related to `protobuf` 7.35.1.

PR #3405, opened 2026-08-10
Validate Jobflow workflow outputs

Andrew-S-Rosen:
## Summary

- Make every `tests/jobflow` execution inspect a concrete returned result.
- Resolve upstream quacc jobs automatically inside Jobflow job arguments and ordinary containers.
- Remove explicit `.output` composition from quacc-decorated Jobflow workflows.
- Preserve native Jobflow syntax in tests that intentionally demonstrate `@jf.job`.
- Use the plain file-transfer specification API introduced by #3409 for deferred `copy_files` directory references.
- Remove the obsolete Jobflow-specific `JobflowCopy` compatibility wrapper after merging current `main`.

## Why

`ensure_success=True` only confirms that Jobflow did not mark execution as failed. It does not prove that the expected value was returned or that downstream jobs received usable outputs. Quacc workflows also need to remain workflow-engine agnostic rather than requiring Jobflow-specific `.output` access.

PR #3409 replaced custom copy wrappers with ordinary list-of-dictionary transfer specifications. Keeping the old Jobflow wrapper after merging that PR caused an import from the deleted `job_argument.py` module, so this branch now relies entirely on Jobflow's normal argument traversal.

## Validation

- Jobflow test suite: 27 passed
- Ruff lint and formatting checks
- Final audit: all `jf.run_locally` calls in `tests/jobflow` capture and inspect responses
- Final audit: no explicit `.output` calls in `src/quacc/recipes`


src/quacc/wflow_tools/decorators.py
Andrew-S-Rosen, line 1: @vineetbansal ChatGPT had to make several changes to the source code. I imagine this may not be necessary and that there is some simpler change we can be making upstream in jobflow to avoid this.

PR #3449, opened 2026-08-27
MD recipe added for MLIPs with ensemble presets inspired by matcalc

### LINKED ISSUE #3424: Add an MD recipe for MLIPs

We have a toy MD recipe for the EMT calculator [here](https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/emt/md.py). We should add an analogous MD recipe for MLIPs, which is lacking.

There are also probably ways to improve the existing MD user experience. We may want to take inspiration from matcalc https://github.com/materialsvirtuallab/matcalc/blob/main/src/matcalc/_md.py.

RussellKwon:
## Summary of Changes

- I added MD recipe for MLIPs mirroring the EMT recipe.
- MD with MLIP accepts an ensemble name ("nve", "nvt", "nvt_langevin",...)
- Bug Fix: In previous code, temperature_K or pressure_bar of ) was discarded as falsy in emt/md.py.

## Things to discuss
- Ensemble stuff could be moved to Runner.run_md (or whatever you prefer) so it can be used for different calculator settings.
- The four MD recipes currently use three parameter conventions 
(`emt`/`mlip`: `steps`, `timestep_fs`, `temperature_K`, `pressure_bar`; `vasp`: `nsteps`, `timestep`, `temperature`, `pressure`; `torchsim`: `n_steps`, `timestep`, `temperature`)

Closes #3424.

### DIFF
--- a/src/quacc/recipes/emt/md.py
+++ b/src/quacc/recipes/emt/md.py
@@ -8,7 +8,6 @@ from typing import TYPE_CHECKING
 
 from ase.calculators.emt import EMT
 from ase.md.npt import NPT
-from ase.md.verlet import VelocityVerlet
 from ase.units import bar, fs
 
 from quacc import Remove, job
@@ -30,7 +29,7 @@ if TYPE_CHECKING:
 @job
 def md_job(
     atoms: Atoms,
-    dynamics: MolecularDynamics | MDEnsemble = VelocityVerlet,
+    dynamics: MolecularDynamics | MDEnsemble = "nve",
     steps: int = 10000,
     timestep_fs: float = 0.5,
     temperature_K: float | None = None,
@@ -48,23 +47,33 @@ def md_job(
     atoms
         Atoms object
     dynamics
-        Either an ASE `MolecularDynamics` class (from `ase.md.md.MolecularDynamics`)
-        or an ensemble name from [quacc.utils.md.MDEnsemble][], resolved to an ASE
-        dynamics class with sensible defaults by
-        [quacc.utils.md.resolve_md_ensemble][] (all of which can be overridden
-        via `md_params["dynamics_kwargs"]`). For the NPT-family ensembles,
-        `pressure_bar` is routed to the appropriate keyword argument of the
-        dynamics class (`externalstress` or `pressure_au`) and defaults to 0 bar
-        when not specified. If the `NPT` class is used, the cell is transformed
-        to the upper-triangular form it requires.
+        Either an ensemble name from [quacc.utils.md.MDEnsemble][] (the default
+        is `"nve"`), resolved to an ASE dynamics class with sensible defaults
+        by [quacc.utils.md.resolve_md_ensemble][] (all of which can be
+        overridden via `md_params["dynamics_kwargs"]`), or an ASE
+        `MolecularDynamics` class (from `ase.md.md.MolecularDynamics`). For the
+        NPT-family ensembles, `pressure_bar` is routed to the appropriate
+        keyword argument of the dynamics class (`externalstress` or
+        `pressure_au`) and defaults to 0 bar when not specified. If the `NPT`
+        class is used, the cell is transformed to the upper-triangular form it
+        requires.
     steps
         Number of MD steps to run.
     timestep_fs
         Time step in fs.
     temperature_K
-        Temperature in K, if applicable for the given ensemble.
+        Temperature in K, if applicable for the given ensemble. Unless the
+        input atoms already have nonzero momenta, a Maxwell-Boltzmann
+        distribution at this temperature is also applied to initialize the
+        velocities; set `md_params={"maxwell_boltzmann_kwargs": None}` to
+        disable this. When `dynamics` is an ASE class, `temperature_K` is also
+        passed to the class and is only compatible with classes that accept
+        that keyword argument.
     pressure_bar
-        Pressure in bar, if applicable for the given ensemble.
+        Pressure in bar, if applicable for the given ensemble. When `dynamics`
+        is an ASE class, this is passed as `pressure_au` and is only compatible
+        with classes that accept that keyword argument (e.g. the Berendsen
+        family); prefer an ensemble name for NPT runs.
     md_params
         Dictionary of custom kwargs for the MD run. For a list of available
         keys, refer to [quacc.runners.ase.Runner.run_md][].
@@ -98,6 +107,8 @@ def md_job(
         atoms = upper_triangular_cell(atoms)
 
     md_defaults = {"steps": steps, "dynamics_kwargs": dynamics_defaults}
+    if temperature_K is not None and not atoms.get_momenta().any():
+        md_defaults["maxwell_boltzmann_kwargs"] = {"temperature_K": temperature_K}
     md_params = recursive_dict_merge(md_defaults, md_params)
 
     calc = EMT(**calc_kwargs)
--- a/src/quacc/recipes/emt/md.py
+++ b/src/quacc/recipes/emt/md.py
@@ -23,7 +23,8 @@ if TYPE_CHECKING:
     from ase.atoms import Atoms
     from ase.md.md import MolecularDynamics
 
-    from quacc.types import CopyFiles, DynSchema, MDEnsemble, MDParams
+    from quacc.types import CopyFiles, DynSchema, MDParams
+    from quacc.utils.md import MDEnsemble
 
 
 @job
@@ -48,7 +49,7 @@ def md_job(
         Atoms object
     dynamics
         Either an ASE `MolecularDynamics` class (from `ase.md.md.MolecularDynamics`)
-        or an ensemble name from [quacc.types.MDEnsemble][], resolved to an ASE
+        or an ensemble name from [quacc.utils.md.MDEnsemble][], resolved to an ASE
         dynamics class with sensible defaults by
         [quacc.utils.md.resolve_md_ensemble][] (all of which can be overridden
         via `md_params["dynamics_kwargs"]`). For the NPT-family ensembles,
--- a/src/quacc/recipes/emt/md.py
+++ b/src/quacc/recipes/emt/md.py
@@ -94,7 +94,10 @@ def md_job(
     """
     if isinstance(dynamics, str):
         dynamics, dynamics_defaults = resolve_md_ensemble(
-            dynamics, timestep_fs * fs, temperature_K, pressure_bar
+            dynamics,
+            timestep_fs=timestep_fs,
+            temperature_K=temperature_K,
+            pressure_bar=pressure_bar,
         )
     else:
         dynamics_defaults = {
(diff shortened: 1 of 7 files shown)


buildbot-princeton:

tests/core/recipes/emt_recipes/test_emt_recipes.py
```diff
@@ -177,6 +178,21 @@ def test_md_job3():
...
+    atoms = molecule("H2O")
+    output = md_job(
+        atoms,
+        dynamics=Langevin,
+        temperature_K=0,
+        steps=10,
+        md_params={"dynamics_kwargs": {"friction": 0.01}},
+    )
+    assert output["parameters_md"]["temperature_K"] == 0
+    assert len(output["trajectory"]) == 11
```
Andrew-S-Rosen: This test is a bit silly because a 0 K MD simulation is nonsensical. We can keep the change to the source code but probably don't need to add this test.

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+    from ase.md.md import MolecularDynamics
+
+    from quacc.types import DynSchema, MDParams
+
+
+@job
+def md_job(
+    atoms: Atoms,
+    library: Literal["fairchem", "matcalc", "rootstock"],
+    dynamics: MolecularDynamics | str = VelocityVerlet,
```
Andrew-S-Rosen: Here, a `str` type hint is used, but then it is not clear to the user what options they can pick from. A better type hint would be a `Literal` type hint like `Literal["npt", "nvt", ...]` to enumerate the possible options.

(Changes we make to the MLIP class we should also make the same changes to the toy EMT calculator just for good measure.)

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+    steps: int = 1000,
+    timestep_fs: float = 1.0,
```
Andrew-S-Rosen: Do you think we should keep the default at 1.0 fs or 0.5 fs? The latter is typically used when you have H atoms. I am genuinely not sure which we should keep as the default.

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+
+    from quacc.types import DynSchema, MDParams
+
+
+@job
+def md_job(
+    atoms: Atoms,
+    library: Literal["fairchem", "matcalc", "rootstock"],
+    dynamics: MolecularDynamics | str = VelocityVerlet,
+    steps: int = 1000,
```
Andrew-S-Rosen: Can we change the default to 10000? This is what Yuliang added to the VASP AIMD so I'm going to try to keep it consistent now to these "new" MD classes.

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+    calc = pick_calculator(library, **calc_kwargs)
+    dyn = Runner(atoms, calc).run_md(dynamics, **md_params)
+
+    return Summarize(
+        additional_fields={"name": f"{library} MD"} | (additional_fields or {})
+    ).md(dyn)
+
+
+def _resolve_ensemble(
+    ensemble: str,
```
Andrew-S-Rosen: Should use a string `Literal` here for same reason as above. If you'd like to prevent code duplication, this string `Literal` could be defined as a global type hint in the module and reused. This will prevent us from having to update the type hint in multiple places if we add or remove an ensemble.

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+        "npt": (
+            NPT,
+            temperature
+            | {"externalstress": pressure, "ttime": 25 * fs, "pfactor": 75.0**2 * fs},
+        ),
+        "npt_nose_hoover": (
+            NPT,
+            temperature
+            | {"externalstress": pressure, "ttime": 25 * fs, "pfactor": 75.0**2 * fs},
+        ),
```
Andrew-S-Rosen: These two look identical. We can probably just keep "npt"

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+        raise ValueError(msg)
+
+    return presets[ensemble.lower()]
+
+
+def _upper_triangular_cell(atoms: Atoms) -> Atoms:
+    """
+    Transform the cell to the upper-triangular form required by the ASE `NPT`
+    class, if it is not already in that form. Adapted from matcalc's `MDCalc`.
```
Andrew-S-Rosen: ASE's `NPT` supports both upper and lower triangular matrices as of https://gitlab.com/ase/ase/-/merge_requests/3277. ASE has a built-in function to convert a cell to the lower (or upper) triangular form here https://docs.ase-lib.org/ase/cell.html#ase.cell.Cell.standard_form. Seems like we should probably use the built-in function instead of coding up our own.

Andrew-S-Rosen (commented):
Thanks, @RussellKwon!! Really appreciate it! I have reviewed the PR and have some minor comments to address before this will be merged.

tests/core/recipes/mlip_recipes/test_md_recipes.py
```diff
@@ -0,0 +1,259 @@
...
+
+    output = md_job(
+        atoms,
+        library=library,
+        steps=20,
+        md_params={
+            "maxwell_boltzmann_kwargs": {
+                "temperature_K": 300,
+                "rng": np.random.default_rng(seed=42),
+            },
```
Andrew-S-Rosen, line 52: This is a note more generally about the `md_job` rather than this specific test: I find it rather awkward that, by default, a Maxwell-Boltzmann distribution is not made when `md_job()` is called. Naively, one might expect that the Maxwell-Boltzmann distribution should be initialized to the user-defined temperature unless explicitly requested otherwise. We should probably adjust the recipes to address this.

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,246 @@
...
+
+    calc = pick_calculator(library, **calc_kwargs)
+    dyn = Runner(atoms, calc).run_md(dynamics, **md_params)
+
+    return Summarize(
+        additional_fields={"name": f"{library} MD"} | (additional_fields or {})
+    ).md(dyn)
+
+
+def _resolve_ensemble(
```
Andrew-S-Rosen: This function and the triangular cell helper below are not specific to MLIPs. So, since these features might be used in other codes with MD, we should move them somewhere else. Perhaps it makes sense to move them to `quacc.utils.md`

src/quacc/recipes/mlip/md.py
RussellKwon: I changed to 0.5 fs. I think for the sake of default we should keep it at 0.5 fs, since 1.0 fs would not work for certain cases like H atoms as you pointed out.

src/quacc/recipes/mlip/md.py
RussellKwon: I was referring to matcalc when I wrote this code. I removed npt_nose_hoover (and nvt_nose_hoover as well).

src/quacc/recipes/emt/md.py
```diff
@@ -22,14 +23,15 @@
...
 
 @job
 def md_job(
     atoms: Atoms,
-    dynamics: MolecularDynamics = VelocityVerlet,
-    steps: int = 1000,
-    timestep_fs: float = 1.0,
+    dynamics: MolecularDynamics | MDEnsemble = "nve",
+    steps: int = 10000,
+    timestep_fs: float = 0.5,
```
Andrew-S-Rosen: Now that we adjusted the timestep, can we actually change `steps` to `20000` in both recipes to align with https://github.com/Quantum-Accelerators/quacc/blob/main/src/quacc/recipes/vasp/md.py? It was 10 ps total. Apologies about that.

src/quacc/recipes/emt/md.py
```diff
@@ -72,14 +92,23 @@ def md_job(
...
-        "dynamics_kwargs": {
-            "timestep": timestep_fs * fs,
-            "temperature_K": temperature_K if temperature_K else Remove,
-            "pressure_au": pressure_bar * bar if pressure_bar else Remove,
-        },
-    }
+    if isinstance(dynamics, str):
+        dynamics, dynamics_defaults = resolve_md_ensemble(
+            dynamics, timestep_fs * fs, temperature_K, pressure_bar
+        )
```
Andrew-S-Rosen, line 101: Since the pressure and pressure are keyword arguments, they should be specified accordingly (e.g. `temperature_K=temperature_K`) rather than positional.

src/quacc/recipes/mlip/md.py
```diff
@@ -0,0 +1,136 @@
...
+    Returns
+    -------
+    DynSchema
+        Dictionary of results, specified in [quacc.schemas.ase.Summarize.md][].
+        See the type-hint for the data structure.
+    """
+    if isinstance(dynamics, str):
+        dynamics, dynamics_defaults = resolve_md_ensemble(
+            dynamics, timestep_fs * fs, temperature_K, pressure_bar
+        )
```
Andrew-S-Rosen, line 118: Since the pressure and pressure are keyword arguments, they should be specified accordingly (e.g. `temperature_K=temperature_K`) rather than positional.

src/quacc/utils/md.py
```diff
@@ -0,0 +1,144 @@
...
+    "npt_inhomogeneous",
+    "npt_mtk",
+    "npt_isotropic_mtk",
+]
+"""Ensemble names accepted by [quacc.utils.md.resolve_md_ensemble][]."""
+
+
+def resolve_md_ensemble(
+    ensemble: MDEnsemble,
+    timestep: float,
```
Andrew-S-Rosen: The `timestep` argument here seems dangerous because elsewhere we use `timestep_fs`. We should probably change this to `timestep_fs` and do the unit conversion inside this function rather than in all the recipes that call it.

tests/core/recipes/mlip_recipes/test_md_recipes.py
```diff
@@ -0,0 +1,161 @@
...
+        dynamics=Langevin,
+        steps=20,
+        timestep_fs=0.5,
+        temperature_K=300,
+        md_params={
+            "dynamics_kwargs": {"friction": 0.01},
+            "maxwell_boltzmann_kwargs": {
+                "temperature_K": 300,
+                "rng": np.random.default_rng(seed=42),
+            },
```
Andrew-S-Rosen: The duplicate setting of temperature in the keyword argument and maxwell_boltzmann_kwargs is now redundant. We can remove the latter here and in related tests where both are specified as the same value.

Andrew-S-Rosen:
Thanks, @RussellKwon! I left a few remaining comments. We are almost there. Please feel free to "resolve" comments you have already addressed as well. 👍

Andrew-S-Rosen:
Merging. Thank you for your contribution, @RussellKwon!

PR #3456, opened 2026-08-31
Handle graceful MD wall-time interruptions

### LINKED ISSUE #2408: [Proposal]: Better handling of MD-type jobs where a termination may not be a failure

This proposal aim to extend the work of the terminate() function by making it call a new `failed_schema()` function. Ideally this function should attempt to fetch the current available results. For example in the case of a timed-out MD, the traj and log should be read and put in a dictionary in the same way that this is already done.

The problem, as mentioned by @Andrew-S-Rosen is that:

> To do what you're suggesting, we would need to try to prepare a schema, write it to disk, and then terminate. Output files will also not always be able to be parsed (e.g. if the calculator crashes instantly), and this would cause the schema generation to crash.

Indeed such function would need to be full or try/exception as no assumption is made on the current state of the calculation. My interpretation is that such function should not attempt to summarise results (no call to pymatgen etc...) but to barely read what is available: the calculation is not done. In the case of logfile and trajfile that's easy, the files are known. In the case of software specific files, it would be nice to come up with a solution to attempt to read them, for example by using #2407.

From the discussions in #2399

Andrew-S-Rosen:
## Summary of Changes

Closes #2408. Unfortunately though, it does not address MD runs where the MD is done in the simulation package itself rather than quacc. For instance, if one does an AIMD run with VASP, this PR will not resolve that problem... :(

### Requirements

### DIFF
--- a/src/quacc/runners/ase.py
+++ b/src/quacc/runners/ase.py
@@ -9,6 +9,7 @@ from importlib.util import find_spec
 from logging import getLogger
 from pathlib import Path
 from shutil import copy, copytree
+from time import monotonic
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


sihoonchoi:
https://github.com/Quantum-Accelerators/quacc/pull/3465#issuecomment-5516684659

@Andrew-S-Rosen I checked it on Della, and both defining `stop_condition` and `max_runtime` work as expected.

PR #3465, opened 2026-09-01
Adding SIGTERM handling for graceful MD exit

sihoonchoi:
## Summary of Changes

Following up on #3456. Added SIGTERM handling to guarantee a graceful MD exit before the walltime limit, providing a fallback when `max_runtime` is not declared.

### Requirements


buildbot-princeton:

sihoonchoi:
@Andrew-S-Rosen What I’ve been trying to do here is catch a SIGTERM signal from the scheduler so that the MD run can be terminated safely. However, it seems that if the signal is sent while a single `dyn.irun` is running, the defined trigger function is not called, so the MD run continues until the walltime is reached. This happens even with a very simple MD setup, where a single `dyn.irun` takes only a fraction of a second. Therefore, I think the previous setup (#3456) handles this issue better.

The other option would be to get the walltime from the batch submission file, but it looks like accessing that information through `os.environ.get` is restricted (at least on Princeton Research Computing). Therefore, explicitly defining `max_runtime` seems to be the best way to ensure that the MD run exits gracefully before reaching the walltime.

Andrew-S-Rosen:
Thanks, @sihoonchoi. Can you check PR #3456 and make sure that it works as expected?

Andrew-S-Rosen:
I have interpreted the comment here as closing this PR and merging #3456. I will proceed accordingly.

PR #3466, opened 2026-09-01
Raises the `quacc[mp]` extra's floor from `atomate2>=0.0.14` to `atomate2>=0.1.5`

hanaol:
## Summary
While testing some features, `omc_static_job` (importing `atomate2.vasp.sets.core`) failed with:

    ImportError: cannot import name 'LobsterSet' from 'pymatgen.io.vasp.sets'

`atomate2==0.0.23` still imports `LobsterSet` from `pymatgen.io.vasp.sets`, but recent `pymatgen` releases no longer expose it there. Since `pymatgen-core>=2026.7.16` is already quacc's declared floor, the old `atomate2>=0.0.14` constraint lets pip resolve to a version of atomate2 that's incompatible with the pymatgen quacc already requires. Bumping to `atomate2>=0.1.5` fixes the import.

## Test plan
- Verified every atomate2 symbol quacc imports (`StaticSetGenerator`, `MatPesGGAStaticMaker`, `MP24PreRelaxMaker`/`MP24RelaxMaker`/`MP24StaticMaker`, `MPGGARelaxMaker`/`MPGGAStaticMaker`, `BaseVaspMaker`, `VaspInputGenerator`) still resolves under `atomate2==0.1.5`.
- Ran the full mocked VASP recipe suite and VASP calculator/params suite (102 tests total, covering every module that touches atomate2) against `atomate2==0.1.5` all of which passed.


buildbot-princeton:

PR #3481, opened 2026-09-10
Aqcat25 static recipe

RussellKwon:
- I added in aqcat25_static_job (RPBE, ENCUT 500, ISMEAR 0 / SIGMA 0.1, EDIFF 1e-4, NELM 250, NSW 0).
- Calculates k points using the same technique implemented inthe fairchem repository for slabs and adsorbate-slab systems. 
- Uses magnetic moments used by AQCat25 as a default, which can be turned off.
-Added in a feature to overide default elemental_magmoms if the user wants.

### DIFF
--- a/src/quacc/recipes/vasp/aqcat.py
+++ b/src/quacc/recipes/vasp/aqcat.py
@@ -1,0 +1,117 @@
+"""AQCat25-compatible slab recipes"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+from quacc import job
+from quacc.recipes.vasp._base import run_and_summarize
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from ase.atoms import Atoms
+
+    from quacc.types import CopyFiles, VaspSchema
+
+
+@job
+def aqcat25_static_job(
+    atoms: Atoms,
+    copy_files: CopyFiles | None = None,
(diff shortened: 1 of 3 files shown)


buildbot-princeton:

Andrew-S-Rosen:
Thanks, @RussellKwon!!! Can you also add an entry to it in the docs [here](https://github.com/Quantum-Accelerators/quacc/blob/main/docs/user/recipes/mlip_recipes.md)?

src/quacc/recipes/vasp/aqcat.py
```diff
@@ -0,0 +1,144 @@
...
+    if spin_polarized:
+        calc_defaults["magmom"] = [magmoms.get(atom.symbol, 0.0) for atom in atoms]
+    if "kpts" not in calc_kwargs and "kspacing" not in calc_kwargs:
+        # Match fairchem's calculate_surface_k_points used by AQCat25.
+        cell = atoms.get_cell()
+        calc_defaults["kpts"] = (
+            max(1, round(40 / np.linalg.norm(cell[0], ord=np.inf))),
+            max(1, round(40 / np.linalg.norm(cell[1], ord=np.inf))),
+            1,
+        )
```
Andrew-S-Rosen: instead of checking if kpts or kspacing is in calc kwargs, just put this in `calc_defaults`. The `calc_defaults` do not override any user settings.

